PR #1, opened 2020-09-01
Added some asyncio support

shibshib:
** Work in progress **

This adds some asyncio support to the dendrite module.

*TODO*: 
- Implement more in depth async tasks at the grpc level.
- Implement some sort of performance test for bittensor with vs without asyncio support.


shibshib:
Removing as it's an unneeded optimization.

PR #3, opened 2020-09-10
Changed up MNIST architecture and cleaned up code a little

shibshib:
This PR brings MNIST to state of the art, with 98% accuracy by the third epoch.

image


unconst (approved):
SOTA baby

PR #5, opened 2020-09-18
Added dockerization

shibshib:
This adds the ability to run dockerized containers of all the bittensor examples.

Example: `./start_bittensor.sh` where the default neuron is the MNIST example.

**NOTE**: For now, this only does LOCAL dockerization, no digital ocean stuff yet.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,59 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get install -y wget sudo curl vim
+
+
+RUN apt-get update \
+  && apt-get install -y python3-pip python3-dev \
+  && cd /usr/local/bin \
+  && ln -s /usr/bin/python3 python \
+  && pip3 install --upgrade pip
+
+  RUN apt-get update && apt-get install -y --no-install-recommends \
+        git \
+        build-essential \
+        curl \
+        python \
+        python3-dev \
+        python3-setuptools \
+        python3-pip \
+        libffi-dev \
+        libssl-dev \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install --upgrade pip && \
+  pip3 install argparse && \
+  pip3 install grpcio && \
+  pip3 install grpcio-tools && \
+  pip3 install libeospy && \
+  pip3 install loguru && \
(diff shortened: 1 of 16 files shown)


unconst:
Looks like you should remove eospy from the dockerfile
also, suggest not using unconst/bittensor docker image as the base

shibshib:
Thanks! Let me fix this

Dockerfile.base
```diff
@@ -1,55 +1,23 @@
...
-  RUN apt-get update && apt-get install -y --no-install-recommends \
-        git \
+ RUN apt-get update && apt-get install -y --no-install-recommends \
         build-essential \
-        curl \
-        python \
-        python3-dev \
         python3-setuptools \
-        python3-pip \
         libffi-dev \
```
unconst, line 6: Where do we use libffi?

unconst: resolved

bittensor/metagraph.py
```diff
@@ -68,7 +68,7 @@ def get_peers(self, n: int) -> List[str]:
         Returns:
             List[str]: List of peers.
         """
-        peer_list = list(self._peers.values())
+        peer_list = list(self._peers)
```
unconst, line 71: Good find

bittensor/neuron.py
```diff
@@ -91,6 +91,7 @@ def subscribe(self, synapse: bittensor.Synapse):
         logger.info("subscribe synapse: {}", synapse_proto)
         self._metagraph.subscribe(synapse_proto)
         self._axon.subscribe(synapse_proto, synapse)
+        synapse.setup_optimizer()
```
unconst, line 94: This looks good although we will need to document this well. 
Another way to do this would be making the optimizer a property of the synapse. I think that might be cleaner.

bittensor/synapse.py
```diff
@@ -20,8 +22,14 @@ def outdef(self) -> bittensor_pb2.TensorDef:
         return NotImplementedError  
     
     def synapse_key(self) -> str:
-        return self._synapse_key      
+        return self._synapse_key
     
+    def setup_optimizer(self):
```
unconst, line 27: Take a look at python @property types. Might be better used here.

unconst (approved):

PR #8, opened 2020-09-22
Distill

unconst:
Adds distillation into the mnist main synapse. Leaves all other files unchanged.
Distillation minorly reduces the performance in this commit for MNIST which is to be suspected. 
Also, adds back the local model params with the optimizer in main. This creates a double gradient update over the synapse model. 
Not optimal, but will be changed in later commits.

### DIFF
--- a/examples/mnist/main.py
+++ b/examples/mnist/main.py
@@ -36,8 +36,7 @@ class Mnist(bittensor.Synapse):
         self.dist_average2 = nn.AvgPool2d(2, stride=2)
         self.dist_conv3 = nn.Conv2d(16, 120, kernel_size=4, stride=1)
         self.dist_fc1 = nn.Linear(120, 82)
-        self.dist_fc2 = nn.Linear(82, 10)
-        
+        self.dist_fc2 = nn.Linear(82, 10)    
         
     # TODO(const): hide protos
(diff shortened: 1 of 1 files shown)


examples/mnist/main.py
```diff
@@ -48,23 +57,33 @@ def outdef(self):
...
+        x = x.view(-1, 1, 28, 28)
+        x = torch.tanh(self.dist_conv1(x))
+        x = self.dist_average1(x)
+        x = torch.tanh(self.dist_conv2(x))
+        x = self.dist_average2(x)
+        x = torch.tanh(self.dist_conv3(x))
+        x = x.view(-1, x.shape[1])
+        x = F.relu(self.dist_fc1(x))
+        x = F.relu(self.dist_fc2(x))
+        return x
```
shibshib, line 70: Is the distill always going to look like the fwd pass?

shibshib (approved):

examples/mnist/main.py
unconst: Not always, but in this case the output of the network is [batch_size, 10] to fit mnist. In all likelihood, outputs would be 1024 + embeddings in which case, the distillation layer would learn x -> 1024

PR #13, opened 2020-09-23
Updated CIFAR with a DPN model

shibshib:
This PR updates the CIFAR example with a DPN model, specifically the one described in: https://arxiv.org/pdf/1707.01629.pdf where a dual path network (DPN) is proposed that utilizes elements of RESNETs and DenseNets to achieve the highest image recognition accuracy possible under this architecture.

The general architecture of the DPN92 network used here is as follows:

Conv 1  | 7 x 7, 64, stride 1
Conv 2 | batch norm 2D (64)
             |   1 x 1, 96
             |   3 x 3, 96
             |   1 x 1, 256
Conv 3 |   1 x 1, 192
             |   3 x 3, 384
             |   1 x 1, 1024

Conv 4 |   1 x 1, 768
             |   3 x 3, 768
             |   1 x 1,  2048

global average pool  |
Softmax layer |

This PR also removes the junk .data files that were added erroneously in previous commits. 

Additionally, it also fixes the cuda issue in MNIST and rehoboam. 

Finally, this PR also renames the rehoboam example to its correct name "transformer". This is more accurate as rehoboam will eventually be a miner, not just a transformer example.


examples/cifar/main.py
```diff
@@ -14,45 +15,92 @@
...
-        self.dist_fc3 = nn.Linear(84, 10)
+        self.dist_conv1 = nn.Conv2d(3, 64, kernel_size=3, stride=1, padding=1, bias=False)
+        self.dist_bn1 = nn.BatchNorm2d(64)
+        self.dist_layer1 = copy.deepcopy(self.layer1) 
+        self.dist_layer2 = copy.deepcopy(self.layer2) 
+        self.dist_layer3 = copy.deepcopy(self.layer3)
+        self.dist_layer4 = copy.deepcopy(self.layer4)
+
+    def set_batch_size(self, batch_size):
+        self.batch_size = batch_size
```
shibshib: Had to perform this mini-hack in checking the batch size because we flatten the input before we give it to the network.. this loses the shape information of the tensor, so while it runs well out of the network we still cannot reshape out input back on time to get the right batch_size back, so this is an extra module added  to help us set the batch size for the network to shape its output properly

shibshib: There's probably a better way of doing this, but for now this will do.

examples/cifar/main.py
```diff
@@ -63,10 +111,41 @@ def input_shape(self):
...
+            self.conv2 = nn.Conv2d(in_planes, in_planes, kernel_size=3, stride=stride, padding=1, groups=32, bias=False)
+            self.bn2 = nn.BatchNorm2d(in_planes)
+            self.conv3 = nn.Conv2d(in_planes, out_planes+dense_depth, kernel_size=1, bias=False)
+            self.bn3 = nn.BatchNorm2d(out_planes + dense_depth)
+
+            self.shortcut = nn.Sequential()
+            if first_layer:
+                self.shortcut = nn.Sequential(
+                    nn.Conv2d(last_planes, out_planes + dense_depth, kernel_size=1, stride=stride, bias=False),
+                    nn.BatchNorm2d(out_planes + dense_depth)
```
shibshib, line 122: The DPN network is built by stacking multiple modualized mirco-blocks, the structure of each micro-block is designed with a bottleneck style, which starts
with a 1 × 1 convolutional layer followed by a 3 × 3 convolutional layer, and ends with a 1 × 1 convolutional layer. The output of the last 1 × 1 convolutional layer is split into two parts: the first
part is element-wisely added to the residual path, and the second part is concatenated with the densely
connected path.

unconst (approved):

examples/cifar/main.py
```diff
@@ -14,45 +15,92 @@
...
+        x = self.layer2(x)
+        x = self.layer3(x)
+        x = self.layer4(x)
+        x = F.avg_pool2d(x, 4)
+        x = x.view(-1, x.size(1))
+        x = self.linear(x)
+
+        if self.batch_size:
+            x = x.view(self.batch_size, -1)
+
```
shibshib: Change to (-1, 10) to match output.

PR #38, opened 2020-10-07
Loadsave

unconst:
Move the bittensor metagraph/ dendrite and axon behind the api. Move the router params into the model object. Makes code cleaner and joins the router and model params under the same object to allow proper checkpointing.


bittensor/metagraph.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, config: bittensor.Config):
         self._server_thread = None
         self._running = False
         
-    def get_synapses(self, n: int) -> List[bittensor_pb2.Synapse]:
+    def synapses(self, n: int = 1000) -> List[bittensor_pb2.Synapse]:
```
shibshib, line 46: Should do something similar for `get_peers`. i.e. `n = 10`

examples/bert/main.py
```diff
@@ -168,7 +177,7 @@ def forward(    self,
             loss = loss + local_target_loss
             
         # Compute NSP loss for network outputs. Only run this if we have passed network inputs.
-        if network is not None and next_sentence_labels is not None:
+        if query and next_sentence_labels is not None:
```
shibshib: Little nitpicky: can just keep indented and check for next sentence labels inside the `if query`

shibshib (approved):
See comments, but LGTM otherwise.

bittensor/__init__.py
```diff
@@ -29,4 +29,59 @@
...
+    # Build the dendrite and router. 
+    # The dendrite is a torch nn.Module object which makes calls to synapses across the network
+    # The router is responsible for learning which synapses to call.
+    global dendrite
+    dendrite = bittensor.Dendrite( config )
+
+
+def serve (synapse: Synapse):
+    # Subscribe the synapse object to the network.
+    global metagraph
```
shibshib: take these out

shibshib (approved):

PR #56, opened 2020-10-16
Loss checkpointing

unconst:
-- adds a swarm test for mnist accuracy.

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -31,7 +31,7 @@ import bittensor.utils.batch_transforms
 
 
 __version__ = '0.0.0'
-__network_dim__ = 256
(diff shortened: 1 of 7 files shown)


bittensor/__init__.py
```diff
@@ -31,12 +31,24 @@
 
 
 __version__ = '0.0.0'
-__network_dim__ = 256
+__network_dim__ = 512
```
shibshib, line 34: should we make this network dim a param that we can adjust?

shibshib (approved):

bittensor/__init__.py
unconst: it needs to be fixed or else the peers won't agree on the encoding

PR #70, opened 2020-10-19
Added DPN for CIFAR-10 dataset

shibshib:
This PR brings back the DPN developed earlier for the CIFAR-10 dataset, but this time under the new structure for defining custom models on bittensor.

This PR also fixes several bugs in all synapses in which they erroneously leave some tensors on the CPU while trying to calculate them on the GPU, breaking execution for any gpu-enabled devices.

### DIFF
--- a/bittensor/synapses/bert/model.py
+++ b/bittensor/synapses/bert/model.py
@@ -4,9 +4,7 @@ import torch
 from torch import nn
 import torch.nn.functional as F
 import transformers
-from transformers import BertModel, BertTokenizer, BertConfig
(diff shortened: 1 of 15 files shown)


unconst:
Had to make a few changes to the files because there were come naming issues from the copy from mnist. 
There are few remaining issues, 
1) The forward pass doesn't use the student or external context inputs tor produce the hidden units, you need to join the units together.
2) There were a number of problems here which made it unlikely that you were ever passing messages to the wire. I suggest writing an integration test which makes sure that the synapse is working as intended.

unconst (approved):

PR #71, opened 2020-10-21
Logbandwidth

unconst:
-- adds a universal bittensor tblogger 
-- moves the mnist synapse to an ffnn synapse for convention
-- takes out the model toolbox where it is not needed yet
-- removes a faulty test from test_crypto (test an fail randomly by design)
--  change naming convention for ffnn synapse to the same as cifar
-- adds a config file for the ffnn synapse
-- allows models to be deep copied using a config


bittensor/synapse.py
```diff
@@ -25,15 +25,25 @@ def synapse_key(self) -> str:
...
         SynapseClass = self.__class__
-        synapse_copy = SynapseClass()
+        
+        model_config = None
+        if config == None:
+            # If no passed config, try model config.
+            if self.config == None:
+                raise ValueError('Deep copy requires a passed model config object or a member model.config')
+            model_config = self.config
+        synapse_copy = SynapseClass(model_config)
```
shibshib: this looks weird. `config` vs `self.config` can be confusing. Maybe rename config to `synapse_config`?

bittensor/synapses/dpn/config.py
```diff
@@ -28,13 +30,13 @@ class DPNConfig:
...
 
-        >>> from bittensor.synapses.dpn.dpn_configuration import DPNConfig
+        >>> from bittensor.synapses.dpn.config import DPNConfig
 
         >>> # Initializing a DPN configuration
         >>> configuration = DPNConfig()
 
-        >>> # Accessing the model configuration
-        >>> configuration = model ( configuration )
+        >>> # Initializing a DNP Synapse
```
shibshib: `DNP` -> `DPN`

bittensor/synapses/dpn/model.py
```diff
@@ -17,22 +17,22 @@ class DPNSynapse(bittensor.Synapse):
...
-                    config: DPNConfig,
+                    config: DPNConfig = None,
                     dendrite: bittensor.Dendrite = None,
                     metagraph: bittensor.Metagraph = None
                 ):
         r""" Init a new DPN synapse module.
 
             Args:
-                config (:obj: `bittensor.dpn.dpn_configuration.DPNConfig`, `required`)
+                config (:obj: `bittensor.dpn.dpn_configuration.DPNConfig`, `required`, defaults to DPNConfig())
```
shibshib: `bittensor.dpn.dpn_configuration.DPNConfig` should be `bittensor.dpn.config.DPNConfig`

bittensor/synapses/dpn/model.py
```diff
@@ -17,22 +17,22 @@ class DPNSynapse(bittensor.Synapse):
     """
 
     def __init__(   self, 
-                    config: DPNConfig,
+                    config: DPNConfig = None,
```
shibshib, line 20: should we allow this to be None? it'd break the model init

bittensor/utils/dispatcher.py
```diff
@@ -113,6 +113,9 @@ def combine(self, expert_out, gates, multiply_by_gates=True):
         combined = zeros.index_add(0, batch_index, flat_stitched.float())
 
         # reshape as original
-        combined = combined.view(expert_out[0].shape)
+        try:
+          combined = combined.view(expert_out[0].shape)
+        except:
+          pass
```
shibshib: this is dangerous. Should at least give a logger error if we don't want to raise here.

bittensor/synapses/dpn/model.py
unconst: Shouldn't break because we have a check in the init

bittensor/utils/dispatcher.py
unconst: This shouldn't be here.  Thanks

shibshib (approved):

PR #77, opened 2020-10-23
Re-add removed test and add integration tests

shibshib:
This PR is pretty big. List of changes:

- Restructures `tests` directory into `integration_tests` and `unit_tests`. Each subdirectory will mirror the `bittensor/` development directory for testing. 
- `integration_tests` is split into two subdirectories, `cpu_tests` that run on circleCI and `gpu_tests` that run on GradientCI. 
- `gpu_tests` run MNIST and test and CIFAR swarm style nodes on GPUs. a TODO here is to also add GPT2 and BERT to be sure. 
- Adds `.ps_project/config.yaml` for gradientCI
- Adds a new `RemoteIPException` that triggers if we mess up setting up the remoteIP somehow.


.circleci/config.yml
```diff
@@ -58,11 +72,14 @@ workflows:
...
-      - "Push Production candidate to DockerHub":
+      - "Run Integration Tests":
               requires:
                 - "Build Bittensor"
-                - "Run Unit Tests"
-              filters:
-                branches:
-                  only:
-                    - master
+      #- "Push Production candidate to DockerHub":
```
unconst: Should we be leaving comments in these commit? @parall4x @shibshib

bittensor/config.py
```diff
@@ -5,6 +5,7 @@
 
 import bittensor
 from bittensor.crypto import Crypto
+from bittensor.exceptions.ResponseExceptions import RemoteIPException
```
unconst: @shibshib @parall4x  since we only have a few exceptions and they extend beyond "Response Exceptions" should we just rename ResponseExceptions to Exceptions?

bittensor/config.py
```diff
@@ -62,7 +63,7 @@ class Config(object):
         Crypto.generate_private_ed25519().public_key())
     try:
         __remote_ip_default__ = requests.get('https://api.ipify.org').text
-    except:
+    except RemoteIPException as e:
```
unconst, line 70: This occurs when the synapse is not internet connected. Allows for local training. Perhaps we should have a log statement here to inform the user? @parall4x @shibshib

bittensor/config.py
```diff
@@ -128,7 +129,7 @@ def __str__(self):
             self.chain_endpoint, self.neuron_key, self.axon_port, self.metagraph_port,
             self.metagraph_size, self.bootstrap, self.remote_ip, self.datapath, self.logdir)
 
-    def from_hparams(hparams):
+    def from_hparams(self, hparams):
```
unconst: This shouldn't be a non-static function because we call it on bittensor.Config.from_hparams(hparams) to return a new config object. Pretty sure this will cause an error in the mnist tests?

bittensor/exceptions/ResponseExceptions.py
```diff
@@ -21,3 +21,7 @@ class SerializationException(Exception):
 class NonExistentSynapseException(Exception):
     """ Raised when the called synapse is not in the local synapse set """
     pass
+
+class RemoteIPException(Exception):
```
unconst, line 25: Same point from above, we should move this into its own exception type file or rename ResponseExcpetions.

bittensor/synapses/bert.py
```diff
@@ -6,6 +6,32 @@
 import transformers
 from transformers import BertModel, BertConfig
 
+def mlm_batch(data, batch_size, tokenizer, collator):
```
unconst, line 9: This is a good idea, moving it out of the specific example, however, we should create a new dir for dataloaders. This is probably to big for this commit but we will want them to be abstracted and easy to import. 

tldr: mlm_batch is not bert specfic

examples/bert/main.py
```diff
@@ -7,7 +7,7 @@
 
 """
 import bittensor
-from bittensor.synapses.bert import BertSynapseConfig, BertMLMSynapse
+from bittensor.synapses.bert import BertSynapseConfig, BertMLMSynapse, mlm_batch
```
unconst, line 10: Same comment from above.

tests/integration_tests/cpu_tests/bittensor/test_swarm.py
```diff
@@ -11,6 +11,7 @@
 import torchvision.transforms as transforms
 import copy
 from typing import List, Tuple, Dict, Optional
+import unittest
```
unconst, line 14: This needs to be added to the requirements file

bittensor/config.py
dehekkendekrekker, line 70: Not sure this is going to work, unless you've made requests throw custom exceptions. 
Furthermore, if you handle a specific type of exception,
make sure there is another one that catches the rest.
Finally, log all exception throw before they are handles. This will pay itself back when trying to debug.

Suggestion:
except RemoteIPException as e:
   log(DEBUG, e.msg, e.trace)
        __remote_ip_default__ = 'localhost'
   __
except Exception as e:
 log(DEBUG, e.msg)
 // Do error handling

bittensor/config.py
shibshib: Good point. I changed it because i didn't see an `@staticmethod` in there. Added it in, and removed the `self`

bittensor/config.py
shibshib: Added log, and raised exception if requests.get fails.

bittensor/config.py
shibshib: Done.

.circleci/config.yml
shibshib: Leftover comment, fixed.

bittensor/exceptions/ResponseExceptions.py
shibshib: Done.

bittensor/synapses/bert.py
shibshib: Good point! I'll make the new dir next commit.

tests/integration_tests/cpu_tests/bittensor/test_swarm.py
shibshib: Done

bittensor/config.py
shibshib: Good point. I added a `raise` to the initial if statement and added a log.

tests/integration_tests/cpu_tests/bittensor/test_swarm.py
shibshib: `unittest` is a standard module, we don't need to add it to our requirements file.

unconst (approved):

PR #79, opened 2020-10-29
Refactor config

dehekkendekrekker:
This PR comprises of a refactoring of the bittensor config.
The most important change is the availability to include a config.ini to configure the node.
The interface to the config object and the command line parameters have been left intact, so everything _should_ continue to work as it does.

In addition, all config fields are validated, so BT won't crap out with some weird exception if the user screws up the config.

The loading of the config is abstracted away from the user. This means the user does not have to explicitly load the config for BT.

There was a weird call to a network service upon source code parsing, which has been removed.

Unit tests have been adapted where necessary
Integration tests have been fixed

Summary
- Parsing of config file
- Validation of config
- Further abstraction of config loading in client main script (is now done in bittensor.init())
- Implemented optimization that make BT load faster
-Fixed tests


shibshib:
I think you need to add `validators` to the requirements.txt, tests are failing because of that

shibshib:
Gradient failed, but this looks good to me otherwise.

shibshib (approved):

PR #83, opened 2020-11-10
Added serialization unit tests

shibshib:
This PR adds some reinforcement to Bittensor for error checking, error recovery, and increases unit test coverage.

More specifically, this PR adds unit tests to the Dendrite, Serialization, and Axon. While adding unit tests I found a few weak spots that didn't throw proper exceptions or needed some reinforcements as well. Now we are catching exceptions but not throwing them, since that'd interrupt the bittensor flow. Instead we are returning null tensors, which are handled on the other side accordingly.


dehekkendekrekker (approved):
Good stuff. Perhaps you could add some comments that explain what is tested and why?

PR #85, opened 2020-11-17
Tensorboard logging

shibshib:
Adding more sophisticated metrics to tensorboard. 

image

image


unconst (approved):

bittensor/dendrite.py
```diff
@@ -234,7 +236,13 @@ def forward(ctx, caller: RemoteNeuron, dummy: torch.Tensor,
...
 
             # Forward tensor.
-            response = ctx.caller.stub.Forward(request, timeout=1.0)
+            pre_response_time = time.time() # in seconds
+            response = ctx.caller.stub.Forward(request)
+            # Time (in seconds) response took
+            elapsed_time = time.time() - pre_response_time
+            bittensor.session.tbwriter.write_dendrite_network_data('Remote Module Forward Call Response Message Size (MB)', response.ByteSize() / 1024)
+            bittensor.session.tbwriter.write_dendrite_network_data('Remote Module Forward Call Turnaround latency (seconds)', round(elapsed_time, 2))
+
```
dehekkendekrekker, line 245: You may want to put this in a private wrapper method, as to not have one large blurb of code in one method, but split off in to several smaller ones that do one thing.

Alternatively, you could think of using a decorator pattern with a method that takes a function pointer and parameters and does the logging.

def tested_function():
  do_heavy_stuff()

Class PerformanceDecorator:
  logger = logger.package.logger
  timer = timer.package.timer

def measure(fn_ptr, params):
  start_time = time
  result= fnptr(params)
  elapsed_time = ...
  do_stuff_with_result

bittensor/dendrite.py
```diff
@@ -282,7 +290,11 @@ def backward(ctx, grads: torch.Tensor) -> Optional[torch.Tensor]:
                 tensors=[serialized_inputs, serialized_grads])
 
             # Attain backward response
-            response = ctx.caller.stub.Backward(request, timeout=1.0)
+            pre_response_time = time.time()
+            response = ctx.caller.stub.Backward(request)
+            elapsed_time = time.time() - pre_response_time
+            bittensor.session.tbwriter.write_dendrite_network_data('Remote Module Backward Call Response Message Size (MB)', response.ByteSize() / 1024)
+            bittensor.session.tbwriter.write_dendrite_network_data('Remote Module Backward Call Turnaround latency (seconds)', round(elapsed_time, 2))
```
dehekkendekrekker, line 297: see above

dehekkendekrekker (commented):
I'm sure the code works, but needs organizing. If time is pressing, I'd suggest to make a PT story addressing this

PR #95, opened 2020-11-25
fix test

unconst:
--fixes the mnist tests to use the asyncio backend


bittensor/__init__.py
```diff
@@ -11,6 +11,7 @@
 from bittensor.session import BTSession
 
 __version__ = '0.0.0'
+__message_size__ = 65535
```
shibshib: Any reasoning for this message size? Is this a limit?

bittensor/axon.py
```diff
@@ -161,6 +161,14 @@ def Forward(self, request: bittensor_pb2.TensorMessage,
...
                             len(x.shape)))
 
+
+            # filter on weight filter on size filter
+            queue.append(y)
+
+            ...
+
+            x = queue.pop()
+
```
shibshib: You didn't define queue here, perhaps you should pass it to the axon?

tests/integration_tests/cpu_tests/bittensor/test_mnist_node.py
```diff
@@ -53,31 +55,37 @@ def setUp(self):
...
+                best_loss = test_loss
+                logger.info("Best test loss: {}".format(best_loss))
+                self.session.serve( self.model.deepcopy() )
+
+        assert best_loss <= 0.1
+        assert best_accuracy >= 75 
+        # Neurons should be 3, but we are setting to >= 2 due to the
+        # unpredictability of circleci booting up neurons, some neurons might
+        # start and finish before others even boot up.
+        # TODO: bring back this assert once chain issues are resolved
```
shibshib: We should add an assert for the number of peers here, we would've caught the tests incorrectly passing if we had done that earlier, no?

shibshib (approved):

tests/integration_tests/cpu_tests/bittensor/test_mnist_node.py
unconst: However, because we are using a shared chain this number is going to be all over the place.

PR #99, opened 2020-11-25
Logscores

unconst:
-- logs scores to in the mnist tests
-- uses numpy stacking to produce a table
-- colorizes text

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,9 @@ jobs:
           command: pip install -e .
           name: setup
       - run:
+          name: extra
+          command: pip3 install termcolor
+      - run:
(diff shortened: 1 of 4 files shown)


shibshib (approved):

.circleci/config.yml
```diff
@@ -48,6 +48,9 @@ jobs:
       - run:
           command: pip install -e .
           name: setup
+      - run:
+          name: extra
+          command: pip3 install termcolor
```
shibshib, line 53: Can we rename it to something better than "extra"? This won't be so informative. Maybe something like `install termcolor` or something?

shibshib (approved):

PR #106, opened 2020-12-08
Examples

unconst:
-- this PR moves the neurons to examples.


dehekkendekrekker (approved):
I like how the examples have been cleaned up. Looking great, this is something presentable.
A suggestion though, there's a bit of boilerplate code in the __main__ function that could be abstracted away.

You could for instance tuck the argument parsing and config loading/checking away in Config class. 

So you could end up with just:
config = Config.load(add_args)

PR #107, opened 2020-12-08
176034966 extend bittensor with error tracking system

dehekkendekrekker:
Hey guys, we've got an error tracking system. Asyncio does not make this an easy job and it feels pretty hacky, but it should report any exceptions thrown (both in an async task as without).

This is merged with jakes 'examples' PR, so once that gets merged into master, this should merge as well.


shibshib (approved):

shibshib:
You might have some testing issues, but otherwise lgtm.

PR #108, opened 2020-12-09
Metagraphsync

unconst:
-- changes the torch state object in the metagraph, now a torchchainstate and chainstate cache which improves consistency.
-- users now access uids like metagraph.uids rather than metagraph.state.uids
-- the sync chain all emits weights and returns the current weight set on chain.
-- poll pub key is cleaner


bittensor/session.py
```diff
@@ -74,9 +74,12 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
 
         Returns:
             Session: present instance of session.
-        """
+        """        
+        logger.info('session exit')
+        loop = asyncio.get_event_loop()
+        loop.set_debug(enabled=True)
+        loop.run_until_complete(self.stop())
```
shibshib, line 81: Should this get moved until after the `if exc_value` statement is run?

shibshib (approved):

PR #114, opened 2020-12-13
Stats

unconst:


bittensor/metagraph.py
```diff
@@ -214,14 +224,79 @@ def uids(self) -> torch.LongTensor:
...
         r""" Returns the stake held by each known neuron.
         Returns
-            stake: (:obj:`torch.LongTensor` of shape :obj:`(n)`):
+            stake: (:obj:`torch.FloatTensor` of shape :obj:`(n)`):
+                stake of each known neuron.
+        """
+        return self.state.stake
+
+    @property
+    def S(self) -> torch.FloatTensor:
```
shibshib, line 236: nit: Do we really want to have singular letter function names? It's understood that that is the stake, but we need better naming.

bittensor/metagraph.py
```diff
@@ -214,14 +224,79 @@ def uids(self) -> torch.LongTensor:
...
+        Returns
+            incentive: (:obj:`torch.FLoatTensor` of shape :obj:`(n)`):
+                inflation incentive of each each known neuron.
+        """
+        I =  (self.tau * self.ranks) / torch.sum(self.ranks)
+        I = torch.where(torch.isnan(I), torch.zeros_like(I), I)
+        return I
+
+    @property
+    def I(self) -> torch.FloatTensor:
```
shibshib, line 265: Same comment as above.

bittensor/metagraph.py
```diff
@@ -291,6 +366,20 @@ def uids_to_neurons(self, uids: torch.Tensor) -> List[bittensor_pb2.Neuron]:
...
+                neuron info ordered by passed uids.
+        Returns:
+            uids: (torch.LongTensor)
+                uids associated with neurons.
+        """
+        uids = []
+        for n in neurons:
+            uids.append(self.state.uid_for_pubkey[n.public_key])
+        return torch.tensor(uids)
+
```
shibshib: Can probably make this more pythonic: `return [self.state.uid_for_pubkey[n.public_key] for n in neurons]`

bittensor/dendrite.py
```diff
@@ -250,50 +299,111 @@ def __init__(self, neuron: bittensor_pb2.Neuron, config, keypair):
...
+        self.stats.codes[code.item()] += 1
+        if code.item() == bittensor_pb2.ReturnCode.Success:
+            # We only need to record elapsed times / bytes on success.
+            self.stats.total_success_time += elapsed_time
+            self.stats.avg_success_time = self.stats.total_success_time / self.stats.codes[bittensor_pb2.ReturnCode.Success]
+            self.stats.success_bytes_out += sys.getsizeof(inputs)
+            self.stats.success_bytes_in += sys.getsizeof(outputs)
+
+        # ---- On Success: set zero backoffand halve next_backoff ---- 
+        if code.item() == bittensor_pb2.ReturnCode.Success:
```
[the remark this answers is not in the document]
unconst: cause we want to know the return time for success specifically

bittensor/metagraph.py
unconst: reason for this is the relationship between this and the paper

bittensor/metagraph.py
unconst: same as above

PR #121, opened 2020-12-16
Standardized the flow of MNIST and CIFAR neurons, and CUDA fixes

shibshib:
This PR places some standardization on MNIST and CIFAR neurons. They are now essentially identical except for the dataset being used and the model. 

This also makes some CUDA fixes to the flow of both neurons.


unconst (approved):
Nice, very clean

PR #123, opened 2020-12-17
Subscribefix

unconst:
--- Includes a fix for the subscription error, the peer now attempts scubscription multipl times to avoid the priority issues. Also includes a minor fix to the async_chain_weights function

-- Stops the session from borking when an emission throws a priority error, now silently retries until timeout. The priority error is caused by a nonce match issue between two transactions with no fees and the same nonce. This will force a single one of them to back off and retry at a later time.

### DIFF
--- a/bittensor/metagraph.py
+++ b/bittensor/metagraph.py
@@ -28,27 +28,27 @@ def ip_to_int(str_val):
 
 
 class EmitSuccess (Exception):
-    """ Raised when try_async_emit emits weights successfully with known result """
+    """ Raised when try_async_emit emits successfully with known result. """
     pass
 
 class EmitNoOp(Exception):
-    """ Raised when calling emit does not change weights on chain. """
+    """ Raised when try_async_emit has no weights to change on chain. """
     pass
 
 class EmitUnknownError (Exception):
-    """ UnknownError during emit. """
+    """ UnknownError error thrown during try_async_emit. """
     pass
 
 class EmitValueError (Exception):
-    """ Raised during emission when passed weights are not properly set."""
+    """ Raised during try_async_emit when passed weights are not properly set."""
     pass
 
 class EmitTimeoutError (Exception):
-    """ Raised during emission during a timeout """
+    """ Raised during try_async_emit after a wait_for_inclusion timeout. """
     pass
 
 class EmitResultUnknown(Exception):
-    """ Called when an emit step end without a known result, for instance, if the user has wait_for_inclusion = False """
+    """ Raised when try_async_emit returns without known result, notably when the call has wait_for_inclusion = False """
     pass
 
 class ChainState():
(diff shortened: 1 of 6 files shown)


bittensor/metagraph.py
```diff
@@ -18,12 +18,49 @@
...
+    """ Raised during try_async_emit when passed weights are not properly set."""
+    pass
+
+class EmitTimeoutError (Exception):
+    """ Raised during try_async_emit after a wait_for_inclusion timeout. """
+    pass
+
+class EmitResultUnknown(Exception):
+    """ Raised when try_async_emit returns without known result, notably when the call has wait_for_inclusion = False """
+    pass
```
shibshib, line 63: Are these exceptions? Some of them appear to be non-erroneous, like `SubscribeSuccess`..

Should move this to the `Exceptions.py` class we have.

bittensor/metagraph.py
```diff
@@ -567,112 +553,365 @@ async def _poll_pubkey(self, pubkey):
...
+                info = await self.subtensor_client.neurons(self.__keypair.public_key)
+
+            except Exception as e:
+                # ---- Catch errors in request ----
+                raise SubscribeUnknownError("Subscription threw unknown exception {}".format(e))
+
+            if info != None:
+                # ---- Subscription was a success ----
+                raise SubscribeSuccess("Subscription success")
+
```
shibshib, line 640: Not sure why we're raising an exception if subscription was a success...?

bittensor/metagraph.py
```diff
@@ -567,112 +553,365 @@ async def _poll_pubkey(self, pubkey):
...
+            # ---- Timeout while waiting for inclusion ----
+            logger.warning("Emission timeout after {} seconds with error {}", timeout, e)
+        
+        except EmitResultUnknown as e:
+            # ---- Did not wait, result unknown ----
+            logger.trace("Emit results unknown.")
+
+        except EmitNoOp as e:
+            # ---- Emit is a NoOp ----
+            logger.info("When trying to set weights on chain. Weights are unchanged, nothing to emit.")
```
shibshib, line 710: Same comment as above, should be in `Exceptions.py`

bittensor/metagraph.py
```diff
@@ -18,12 +18,49 @@
 
 from bittensor.exceptions.handlers import rollbar
 
+MAX_INT_WEIGHT = 4294967295 # Max weight value on chain.
+
```
shibshib, line 22: If someone changes this, would it break things on the chain?

bittensor/metagraph.py
unconst: Keeping it local for now while in active dev. They are acting more like Dispatch Results than actual errors.

bittensor/metagraph.py
unconst: basically they act as dispatch results not as true errors

bittensor/metagraph.py
unconst: It would destroy the universe.

shibshib (approved):
Approved with condition that we move those exceptions to a `dispatch_results.py` file and change the `SubscribeSuccess` to an enum as discussed

PR #128, opened 2020-12-22
upnpc

unconst:
-- adds 4 new remote ip methods
-- adds optional use_upnpc flag, if this is set, the axon tries to use upnpc to punch a hole in the router

### DIFF
--- a/bittensor/dendrite.py
+++ b/bittensor/dendrite.py
@@ -279,9 +279,9 @@ class RemoteNeuron(nn.Module):
             }
         ) 
         # Loop back if the neuron is local.
-        if neuron.address == config.axon.remote_ip:
+        if neuron.address == config.axon.external_ip:
             ip = "localhost:"
-            if config.axon.remote_ip == "host.docker.internal":
+            if config.axon.external_ip == "host.docker.internal":
                 ip = "host.docker.internal:"
             self.endpoint = ip + str(neuron.port)
         else:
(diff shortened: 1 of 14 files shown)


bittensor/dendrite.py
```diff
@@ -279,9 +279,9 @@ def __init__(self, neuron: bittensor_pb2.Neuron, config, keypair):
             }
         ) 
         # Loop back if the neuron is local.
-        if neuron.address == config.axon.remote_ip:
+        if neuron.address == config.axon.external_ip:
             ip = "localhost:"
-            if config.axon.remote_ip == "host.docker.internal":
+            if config.axon.external_ip == "host.docker.internal":
                 ip = "host.docker.internal:"
             self.endpoint = ip + str(neuron.port)
```
shibshib, line 286: do we really care about docker stuff anymore?

bittensor/utils/networking.py
```diff
@@ -0,0 +1,125 @@
+import miniupnpc
+import os
+import requests
+import urllib
+
+from loguru import logger
+
+class ExternalIPNotFound(Exception):
+    """ Raised if we cannot attain your external ip from CURL/URLLIB/IPIFY/AWS """
```
shibshib, line 9: we are starting to scatter exceptions everywhere, can we put them all into exceptions.py or something?

shibshib (approved):

bittensor/dendrite.py
unconst: no we don't, perhaps jsut get rid of this now

PR #129, opened 2020-12-23
[#176175273]

dehekkendekrekker:
- ~/.bittensor dir is created when it does not exist
- Its permissions are set to user access only
- Hotkey is now auto generated when it does not already exist in ~/.bittensor/hotkey
- Permissions are set to user r/w only
- A public cold key should be put in ~/.bittensor/coldkey, but can be specified with the --session.coldkeyfile argument
- The cold key should take the format 0x[a-z0-9]{64}

### DIFF
--- a/bittensor/config.py
+++ b/bittensor/config.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import validators
 import yaml
+import stat
 from munch import Munch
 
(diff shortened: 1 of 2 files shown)


unconst (approved):
Will we need to pip install re or stat?

PR #133, opened 2020-12-28
Minorfixes

unconst:
-- removal of some unused files
-- fix of the metagraph descriptions


bin/bittensor
```diff
@@ -1,24 +1,23 @@
-#!/bin/python3
+#!/usr/bin/env python
+
+import argparse
+import asyncio
+import json
+import os
+import sys
```
shibshib: Can we rename the name of this script? perhaps `bittensor_cli`?

shibshib (approved):

PR #144, opened 2021-01-05
Added __to_tensorboard__ methods

shibshib:
This PR adds new tensorboard metrics such as dendrite, axon, and metagraph stats.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -84,8 +84,9 @@ class Axon(bittensor_grpc.BittensorServicer):
     def __to_tensorboard__(self, tensorboard, global_step):
         total_in_bytes = (self.stats.total_in_bytes.value * 8)/1000
         total_out_bytes = (self.stats.total_out_bytes.value * 8)/1000
-        tensorboard.add_scalar("axon/total_in_bytes", total_in_bytes, global_step)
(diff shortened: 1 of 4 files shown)


bittensor/axon.py
```diff
@@ -80,6 +80,12 @@ def __str__(self):
         total_out_bytes_str = colored('\u290A {:.1f}'.format((self.stats.total_in_bytes.value * 8)/1000), 'green')
         qps_str = colored("{:.3f}".format(float(self.stats.qps.value)), 'blue')
         return "(" + qps_str + "q/s|" + total_out_bytes_str + "/" + total_in_bytes_str + "kB/s" + ")"
+    
+    def __to_tensorboard__(self, tensorboard, global_step):
+        total_in_bytes = (self.stats.total_in_bytes.value * 8)/1000
+        total_out_bytes = (self.stats.total_out_bytes.value * 8)/1000
```
dehekkendekrekker, line 86: What's this conversion about?

dehekkendekrekker (approved):

bittensor/axon.py
shibshib: This is to convert from bits to kilobytes

PR #164, opened 2021-01-18
Added potential fix to pkm layer

shibshib:
This fixes a potential issue in pkm layer as weights are being sent to device in one line, resulting in erroneous emissions.

### DIFF
--- a/bittensor/config.py
+++ b/bittensor/config.py
@@ -42,7 +42,7 @@ class Config:
         params = parser.parse_known_args()[0]
         config_file = None
         config = Munch()
-        if 'neuron.config_file' in vars(params).keys():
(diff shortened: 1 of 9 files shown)


bittensor/config.py
```diff
@@ -42,7 +42,7 @@ def to_config(parser: argparse.ArgumentParser) -> Munch:
         params = parser.parse_known_args()[0]
         config_file = None
         config = Munch()
-        if 'neuron.config_file' in vars(params).keys():
+        if 'session.config_file' in vars(params).keys():
```
unconst, line 45: we could make this a non indented object like just config_file

unconst (approved):

PR #179, opened 2021-01-26
fix cli regen wallet

unconst:
-- fixes the regen_wallet command on the bittensor-cli


shibshib (approved):
Curious. What issues did this code you took out make?

PR #183, opened 2021-01-27
Fix Token Issue

shibshib:
This fixes the issue with `token indices sequence length is longer than the specified maximum sequence length ` that crops up and breaks things later.

### DIFF
--- a/examples/TEXT/gpt2_genesis.py
+++ b/examples/TEXT/gpt2_genesis.py
@@ -62,7 +62,7 @@ class AdamCorpus():
             random_line = self.lines[ random.randint(0, len(self.lines)-1)]
             batch_text.append( random_line )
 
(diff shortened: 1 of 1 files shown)


examples/TEXT/gpt2_genesis.py
```diff
@@ -62,7 +62,7 @@ def nextbatch(self, batch_size, tokenizer):
             random_line = self.lines[ random.randint(0, len(self.lines)-1)]
             batch_text.append( random_line )
 
-        batch_inputs = tokenizer(batch_text, return_tensors='pt', padding=True)['input_ids']
+        batch_inputs = tokenizer(batch_text, return_tensors='pt', padding=True, truncation=True)['input_ids']
```
shibshib, line 65: This is the fix needed. We created another `next_batch` inside gpt2_genesis and it did not have this fix, whereas gpt2.py has a `next_batch` with this fix in it.

unconst (approved):

PR #186, opened 2021-01-31
initial commit

ghost:
Implemented a bert_grunt node following the example from FFNN_grunt

### DIFF
--- a/examples/TEXT/bert_grunt.py
+++ b/examples/TEXT/bert_grunt.py
@@ -1,10 +1,10 @@
-"""FFNN Grunt node
+"""BERT Grunt Miner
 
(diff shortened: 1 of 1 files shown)


unconst (commented):
NIT: please write a test for this PR. You can find other tests in the tests/examples/ dir

NIT 2: Please run a save for the model in the init should be something like model toolbox save model.

FEATURE 3: This grunt should probably print out its incentive to the screen (logging)

Enhancement: Batch your gradients and apply them all at once
Enhancement: You should probably check the magnitude of the passed gradients before applying (maybe they should be scaled based on the score people set to you ?)

well done :)

PR #188, opened 2021-02-02
Add dockerfile for  bittensor 

shibshib:
This PR adds the ability to dockerize bittensor, with a CUDA-compatible container so as to take full advantage of any GPUs on the host machine.


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get install -y -V libarrow-dev 
+RUN apt-get install -y -V libarrow-glib-dev 
+RUN apt-get install -y -V libarrow-dataset-dev 
+RUN apt-get install -y -V libarrow-flight-dev
+
+WORKDIR /bittensor
+RUN pip install --upgrade numpy pandas setuptools "tqdm>=4.27,<4.50.0"
+RUN pip install -r requirements.txt
+RUN pip install -e . 
+RUN ./scripts/create_wallet.sh
```
unconst: Not sure we should create a wallet for the user, they wont have access to this.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+EXPOSE 8091
+VOLUME ["/bittensor"]
```
unconst: Isn't the volume ~/.bittensor?

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  dev:
+    container_name: node-bittensor
+    image: "bittensor/bittensor:latest"
+    ports:
+      - "8091:8091"
+    
+    command: /bin/bash -c "
+      cd /bittensor 
+      && python3.7 examples/IMAGE/mnist.py --metagraph.chain_endpoint 172.17.0.1:9944"
```
unconst: Does this point to feynman?

requirements.txt
```diff
@@ -40,4 +39,7 @@ torchvision
 transformers
 validators
 xxhash==1.3.0
-codecov
\ No newline at end of file
+pandas
```
unconst, line 42: Did you say this worked because we install cython first?
I thought you mentioned that this doesnt work?

unconst (changes requested):

Dockerfile
shibshib: It doesn't work without a wallet though, so it needs some sort of wallet. Unless we assume users have a wallet in their ~/.bittensor?

Dockerfile
shibshib: No, this is the bittensor code volume, which will be mounted in the root dir of the image

docker-compose.yml
shibshib: No, this is the docker bridge network between the two containers running on the same computer.

requirements.txt
shibshib: We need to install pandas, it wasn't added anywhere, so not sure what you mean :)

shibshib: Ah, just caught what you meant. Yes I moved it to the bottom before thinking that the install order mattered, but it apparently doesn't.

Dockerfile
shibshib: We can always ensure users created their wallet before running this, so that it copies it over to the container

unconst: really think this should be taken out
if a user runs they will be subscribing with these keys.
Not good.

Dockerfile
unconst: right, i think we can mount the root dir so we can access the wallet created by the user

unconst (changes requested):

Dockerfile
shibshib: Done!

Dockerfile
shibshib: Good idea, done!

unconst (approved):

krishadhithya (commented):
LGTM @unconst

PR #194, opened 2021-02-08
Adhi/pkm fixes

krishadhithya:
@shibshib added device selection to 3 models which didn't have it in there.

For the rest I modified what was already written.

**Note**:
1. Some of these models have this: `torch.cuda.empty_cache()` explicitly in their `train()` method to clear memory. Not sure if this is compatible with CPUs.
2. I have written it such that the device config is specified in the depth/level 0 of the yaml. eg in line 16 below:
image
If this is against convention, lmk and I'll modify

### DIFF
--- a/bittensor/routers/pkm.py
+++ b/bittensor/routers/pkm.py
@@ -41,7 +41,10 @@ class PKMRouter():
         if config == None:
             config = PKMRouter.build_config()
         self.config = config
-        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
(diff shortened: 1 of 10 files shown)


unconst (commented):
@krishadhithya unfortunately I dont think this will work well, the issue is that the session is a super class of the synapse and you can't be sure that it will always have the device argument. 

I think it makes more sense for you to define the device argument right into the synapse. The session can then optionally set this argument from the main script, notably

self.model.device = self.config.session.device

This way the user has control.

PR #201, opened 2021-02-10
version fix

shibshib:


scripts/install.sh
```diff
@@ -0,0 +1,187 @@
...
+    wait_for_user
+    mac_install_brew
+    mac_install_cmake
+    mac_install_python
+    install_bittensor
+    ohai "Installation successful!"
+    
+
+else
+  abort "Bittensor is only supported on macOS and Linux."
```
dehekkendekrekker: Change Linux to Ubuntu (version ... and up). The aptitude package manager is not available in other distros.
Arch linux for instance has pacman, Centos has yum, etc ...

I think it's worthwhile pointing this out in the documentation as well, until will support other linux distros.

dehekkendekrekker (dismissed):
See line comment.

scripts/install.sh
unconst: Very good point

unconst: Stackoverflow say's this might work

if [  -n "$(uname -a | grep Ubuntu)" ]; then
    sudo apt-get update && sudo apt-get upgrade 
else
    yum update
fi  

perhaps I should try catch for apt?
which -s apt 
for instance?

unconst:
@parall4x  check the linux apt check, might work, let me know

scripts/install.sh
```diff
@@ -0,0 +1,194 @@
...
+    fi
+    ohai "Installing bittensor deps"
+    python3.8 -m pip install --upgrade pip
+}
+
+
+linux_install_bittensor() {
+    ohai "Installing bittensor"
+    python3.8 -m pip install bittensor
+}
```
shibshib: we use 3.8 here and 3.7 elsewhere, we need to be consistent and always use 3.7

shibshib: We should also just run `pip3.7 install -e .` since pip version is out of date and maintaining that is a hassle.

unconst (approved):

PR #213, opened 2021-02-18
Subtensor update PR

unconst:
-- changes the wallet: no longer needs a default, the user is prompted if there is no hotkey coldkey,
-- --wallet.name can be used to run your node
-- --subtensor.network is used to specify the network
-- removes the asyncio code from the metagraph
-- moves the subtensor client into the main file
-- adds an executor class which follows our config structure
-- adds extrinsic watching to the substrate client
-- returns errors from extrinsic calls.
-- other minor fixes and changes


unconst:
PR will pass once Akira is up again, unfortunately not today

PR #219, opened 2021-03-01
Separate Script for Fedora / RPM based Linux Distro's

il-dar:
Modified script to work with Linux Fedora distribution, so far as a separate script for people to test. I kept the mac parts in it so its easier to merge later once people test it and we can have something that checks the linux distro. 
I am not 100% on line 87:
_python -m pip install python-devel_ it might be _dev_ or _dev-tools_ but everything needed installed on my machine since it pulled it from dnf in the script as well on line 74 from what I understand.

### DIFF
--- a/scripts/installfedora.sh
+++ b/scripts/installfedora.sh
@@ -1,0 +1,239 @@
+#!/bin/bash
+set -u
+
+abort() {
+  printf "%s\n" "$1"
+  exit 1
+}
+
+getc() {
+  local save_state
+  save_state=$(/bin/stty -g)
+  /bin/stty raw -echo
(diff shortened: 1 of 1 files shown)


unconst:
I think we could probably move this into the other install script and just check to see what distro people are running on. 

-- couple nits: probably want to remove the mac stuff here

shibshib (changes requested):
As @unconst said, we can move this functionality to just check what distro people are on and then calling the right methods. One way to check distro would be something like `cat /etc/*-release` that would give you info on your linux distro, you can grep that to find the actual distro name.

PR #226, opened 2021-03-02
adds metagraph overview text file logging

unconst:
--- adds an analysis tool to the bittensor-cli save the metagraph state to a file stored in the wallet.

### DIFF
--- a/bittensor/metagraph.py
+++ b/bittensor/metagraph.py
@@ -19,6 +19,7 @@ import argparse
 import asyncio
 import copy
 import pandas as pd
+import json
 import math
 import numpy
 import random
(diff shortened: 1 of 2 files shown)


bittensor/metagraph.py
```diff
@@ -135,6 +136,22 @@ def __init__(self):
...
+            'tau': self.tau.tolist(),
+            'n': self.n,
+            'uids': self.uids.tolist(),
+            'indices': self.indices.tolist(),
+            'stake': self.stake.tolist(),
+            'lastemit': self.lastemit.tolist(),
+            'W': self.W.tolist(),
+            'neurons': [ {'uid': n.uid, 'ip': n.address, 'port': n.port, 'ip_type': n.ip_type, 'modality': n.modality, 'hotkey': n.public_key} for n in self.neurons]
+        }
+        with open( filepath, 'w') as fp:
```
shibshib, line 151: Should we overwrite the file each time or augment?

shibshib (approved):

bittensor/metagraph.py
unconst: the file name is appended with the block height

PR #228, opened 2021-03-02
add jupyter notebook

unconst:
--- adds a jupyter notebook


unconst:
-- removed notebook

bittensor/metagraph.py
```diff
@@ -581,13 +580,18 @@ async def _async_sync_cache(self):
         last_emit = dict( await self.subtensor.async_get_last_emit() )
         self_uid = await self.subtensor.async_get_uid_for_pubkey( self.wallet.hotkey.public_key )
         if self_uid != None:
-            calls.append ( self._poll_uid ( self.wallet.hotkey.public_key, self_uid ) )        
+            try:
+                calls.append ( self._poll_uid ( self.wallet.hotkey.public_key, self_uid ) )     
+            except KeyFileError:
+                pass
```
shibshib: Is there a reason we're passing here? Shouldn't we at least print a log?

unconst:
--- adds has_coldkey and other key file functions to the wallet
--- adds checks to the metagraph to ensure it does not fail without a wallet
--- kills the process in the executor gracefully using self.wallet.assert_coldkey ...

bittensor/metagraph.py
```diff
@@ -240,10 +241,9 @@ def add_args(parser: argparse.ArgumentParser):
...
         bittensor.subtensor.Subtensor.add_args( parser )
         try:
-            parser.add_argument('--metagraph.stale_emit_filter', default=10000, type=int, 
-                                help='''The metagraph filters neurons with last emit beyond this many blocks.
-                                        Note, this is used to trim the graph size,
-                                        but may change your incentive mechanism view.''')
+            parser.add_argument('--metagraph.stale_emit_filter', default=-1, type=int, 
+                                help='''Filter neurons who have not emitted in this number of blocks.
+                                        -1 for no filter.''')
         except:
```
shibshib, line 247: Can we also fix this except: pass while we're at it?

shibshib (approved):

bittensor/metagraph.py
unconst: this except pass just ensures that the args are not added twice. but yes in another commit

PR #253, opened 2021-04-06
Modified old GPT2 module

shibshib:
This PR contains a few things: 

1. Puts aside the old gpt2 module that uses the huggingface API (renamed to `huggingface_gpt2.py`)
2. Sets up a new `gpt2.py` partially based on the work by Karpathy et al. on their MinGPT. 
3. Sets up a new training workflow where training logs are printed in place, no more long long messages. 
4. First step in benchmarking Bittensor. 
5. New GPT2 module has its own optimizer based on AdamW that decays the learning rate using linear warmup with cosine annealing. 
6. GPT2 now trained on CoLA dataset.
7. Each GPT miner now prints how many params it has before proceeding. Useful for back-of-envelop calculations.

When the current GPT is run using all defaults (i.e. using simply `python ~/.bittensor/bittensor/miners/TEXT/gpt2_genesis/gpt2_genesis.py --subtensor.network $NETWORK --wallet.name $WALLET --wallet.hotkey $HOTKEY`) it will run a medium sized GPT2 version, with **32 heads, 16 layers, and block size 20 (you can think of block size as sequence length, we should look into extending this to 128)** This makes a model with **90,214,400** parameters. 

NOTE: This GPT performs very well when trained locally, but seems to hit a plateau in learning rate when training on Bittensor network, this is likely due to the other Bittensor models using a different GPT that is less optimized, or perhaps due to the dataset being used, or even perhaps due to the PKM layer. This investigation remains to be done. Pivotal ticket: https://www.pivotaltracker.com/story/show/177647854

**EDIT: After running it for a long time (approx. 5 hours) on a GPU it seems confirmed that the loss does go down, albeit slowly. This is likely because it's getting inputs from other GPTs on Bittensor and they're muddling its training because they all have high loss. After a while the local loss went down to approx 2-3 and the remote loss stayed high, likely because the model decided the external nodes aren't so useful to it.**

`GS: 27064 | LS: 10 | Epoch: 55 | LTL: 2.98795 | RTL: 6.18005 | DL: 0.29483  | lr 2.165200e-04. Axon: (107.884q/s|⤊ 69.6/⤋ 69.6kB/s) Dendrite: (1.708q/s|⤊ 50.5/⤋ 41.2kB/s):   2%|██▎                                                                                                        | 11/501 [00:06<04:50,  1.69it/s]`


shibshib:
lol. test_gpt2_wiki failed because it's trying to download the entire wiki dataset. Since both miners are identical i'm going to take out the gpt2_wiki test and rename it to just test_gpt2

unconst (approved):

PR #255, opened 2021-04-06
Yuma

unconst:
--- small fixes from singletons PR that can be merged easily into master


bittensor/wallet.py
```diff
@@ -244,7 +244,7 @@ def _load_coldkeypub(self) -> str:
         print(colored("Loaded coldkey.pub: {}".format( coldkeypub ), 'green'))
         return coldkeypub
 
-    def _load_hotkey(self) -> bittensor.substrate.Keypair:
+    def _load_hotkey(self) -> 'bittensor.substrate.Keypair':
```
0xshreyash, line 247: why do you need a forward reference here?

0xshreyash (approved):

0xshreyash:
One minor nit, approving to unblock since it doesn't really make a difference.

PR #257, opened 2021-04-13
Padding added to received texts

### LINKED ISSUE #274: Too many open files bug
**Describe the bug**
The current huggingface tokenizer has an inherent bug in which it tries to open all the files all the time, this leads to OS errors because there are limitations on how many files can be open at any given time. 

** Fix**:
We need to either reduce the number of times we use a tokenizer in an epoch OR see if huggingface fixed the issue. 

UPDATE: Huggingface fixed the issue and pushed to master recently, we simply need to update our version: https://github.com/huggingface/tokenizers/issues/178

shibshib:
Added padding to received texts. This avoids the error `Forward request returned tensor with incorrect shape [2, 5, 512], expected [2, 20]`


robertalanm (approved):
looks good to me

unconst (approved):

PR #276, opened 2021-04-15
decouple_wallet_from_subtensor

unconst:
--- removes the wallet from the subtensor client, now wallet is passed to relevant functions.


bittensor/executor.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         self.wallet = wallet
 
         if subtensor == None:
-            subtensor = bittensor.subtensor.Subtensor( config = self.config, wallet = self.wallet )
+            subtensor = bittensor.subtensor.Subtensor( config = self.config )
```
dehekkendekrekker, line 64: Is there a reason to assume subtensor == None?

bittensor/metagraph.py
```diff
@@ -634,7 +634,7 @@ def set_weights(self, weights: torch.FloatTensor, wait_for_inclusion = False, ti
         code, message = self._try_emit(weights, wait_for_inclusion, timeout)
         if code == Metagraph.EmitSuccess:
             # ---- Emit was a success. ----
-            logger.info("Emission was successful and entered the block.")
+            logger.success("Emission was successful and entered the block.")
```
dehekkendekrekker, line 637: Dope

bittensor/subtensor.py
```diff
@@ -230,7 +230,7 @@ def connection_error_message():
             # ---- Get next endpoint ----
             ws_chain_endpoint = self.endpoint_for_network( blacklist = attempted_endpoints )
```
dehekkendekrekker, line 231: This logic can be abstracted out

dehekkendekrekker (approved):
Good PR. It's small enough to not get lost in complexity, and clearly shows the intention.
However, the proposed changes, even though this could work, are a bit hacky.

This client class was intended to be a low-level class, basically only interfacing with the subtensor server. Some higher level objects have been introduced (the config object and the wallet object) being container objects for data that is used throughout the codebase. 

This PR removes the injection of the wallet object from the constructor, and injects them into the methods. This is not necessary. Injection into the constructor is a valid approach, as the contents of the objects do not change. It's just that another layer is needed that is able to funnel the data  specific to the client from the config and wallet objects into client's constructor.

Especially in combination with the logic that looks for an appropriate endpoint to use for the client class, a factory class could be a good solution to this problem.

Given a service class that determines the endpoint to use, the factory could simply derive a client object using the wallet and the config as a parameter.

This effectively decouples the code, making it easier to test, without having to resort to cumbersome mocking strategies. (For the client and factory class that is, the service class is a different case).

I'll approve this PR, as to not stall the process, but I do recommend implementing the proposed structure.

bittensor/executor.py
unconst: You can instantiate the executor with a subtensor client, otherwise it just uses the config file.

PR #285, opened 2021-04-26
Bittensor dataloader

shibshib:
**Context**
In order for the incentive mechanism to standardize across miners, we need a lowest-common-denominator dataset for miners to use. This dataset will expand over time as we add more training data to it. Since miners will be pulling data often (on each start, at least) it needs to be hosted on a proper server that is resilient to many miners talking to it. It was determined after discussing with @unconst and @parall4x that IPFS is likely the best place to host these files, as it can act like a file server that miners can talk to by using the [infura IPFS API](https://infura.io/docs/ipfs/post/block-get). This way, each time we add a file to the IPFS dataset, we need to only update the hash corresponding to the genesis dataset.

**Changes**
This PR creates the `BittensorDataLoader` superclass. This superclass defines the interface for every miner to use when pulling data down from IPFS, the idea being to create a subclass for a given modality or dataset and overriding the `__len__` and `__getitem__` methods and then calling the `BittensorDataLoader.dataloader()` method to return a `torch` dataloader instance, which can be iterated over to retrieve batches of data. The `BittensorDataLoader` also contains methods to retrieve the directory and file hashes of a dataset on IPFS.


unconst (approved):

dehekkendekrekker (approved):
dope

PR #286, opened 2021-04-26
Implementation of integration test for subtensor v1.1.0 [Hold review, needs more work]

dehekkendekrekker:
This PR fixes the subtensor v1.0.1 subtensor integration tests, as well as implements the integration tests for subtensor v1.1.0

I did some magic with the Dockerfile in order to make this work. The major difference is that it pulls from an image that has glibc 2.29 so that subtensor can be ran. This image however, does not support python 3.7 natively, so this is installed separately. Right now, this PR uses the bittensor/bittensor_subtensor tests image. When approved, let me know, so I can update bittensor/bittensor:latest with this setup


unconst (approved):
beauty

shibshib (approved):

PR #309, opened 2021-05-09
Fixing path in command

ghost:
Previous path yielded the error "no such file or directory miners/TEXT/gpt2_genesis/gpt2_genesis.py" 
I have updated the path to the actual location of gpt2_genesis.py


unconst (approved):

unconst:
Thanks for the fix

PR #311, opened 2021-05-11
Replaced hard-coded ips to external-access nodes with hostnames

dehekkendekrekker:
Replaces the hard-coded ips with hostnames


unconst (approved):

unconst:
Not sure about this error in the integration tests but I think it has something to do with the new master branch on subtensor

unconst:
Make sure to pull from the latest develop also, they were passing on that branch just now

shibshib (approved):

PR #329, opened 2021-05-18
Factory pattern (experimental)

unconst:


bittensor/axons/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+        config.axon.max_workers = max_workers if max_workers != None else config.axon.max_workers
+        config.axon.forward_processing_timeout = forward_processing_timeout if forward_processing_timeout != None else config.axon.forward_processing_timeout
+        config.axon.backward_processing_timeout = backward_processing_timeout if backward_processing_timeout != None else config.axon.backward_processing_timeout
+        axon.check_config( config )
+
+        # Wallet: Holds you hotkey keypair and coldkey pub, which can be used to sign messages 
+        # and subscribe to the chain.
+        if wallet == None:
+            wallet = bittensor.wallet.Wallet( config = config )
+        wallet = wallet
```
shibshib, line 53: This overall looks good to me, just a heads up though, when using __new__ in python and you instantiate the axon, it's going to call __new__ and then call __init__. Just a heads up!

shibshib (approved):

shibshib:
Awesome PR

bittensor/axons/__init__.py
unconst: Right, but because the __new__ call on the sub Axon is not implemented, this is a no-op right?

PR #410, opened 2021-07-11
Migrating to GitHub Actions

ivanezeigbo:
Config file for migrating from CircleCI to GitHub Actions
- Lint syntax checker added for CI 

Would deprecate CircleCI config once Actions is confirmed to be working properly

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -1,0 +1,52 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Bittensor Python Application
+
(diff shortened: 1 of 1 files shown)


.github/workflows/config.yml
```diff
@@ -0,0 +1,52 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Bittensor Python Application
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
```
shibshib, line 11: Is it possible to also set it up on pull requests? I recall we briefly chatted about this, but I think I mistakenly said let's take out pull requests :) I think we should have them all run on each pull request except for `Push image to DockerHub`

PR #411, opened 2021-07-12
Fix Router Load Bug

Eugene-hu:
Router now updates metagraph before loading and does not require the state dict to contain all of the parameters.


shibshib (approved):

unconst (approved):
Well done, super simple fix

PR #412, opened 2021-07-13
Adding GitHub actions & Deprecating Circle CI

ivanezeigbo:
PR to add Github Actions for CI tests and build and actions for pushing image to docker on merge. Also involves commit to deprecate Circle CI configs.


.github/workflows/config.yml
```diff
@@ -0,0 +1,49 @@
...
+    #     flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+    #     # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+    #     flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Build Bittensor
+      run: |
+        python -m pip install -e .
+        ./scripts/create_wallet.sh
+    - name: Run Unit Tests
+      run: |
+        ./scripts/build_protos.sh
```
unconst: maybe put this in Build bittensor

.github/workflows/config.yml
```diff
@@ -0,0 +1,49 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
shibshib: Should rename this file to something like "pull_request_tests" or something? config is not so descriptive.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,53 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
```
shibshib: Same comment for renaming, maybe rename this to docker_image_push

unconst (approved):

PR #414, opened 2021-07-13
Subtensor fix

Eugene-hu:
Bug Fix for subtensor class. This includes a new test for overrides of the subtensor network by both config and network arg. Also updated the default chain entry point from akira to kusanagi


unconst (approved):

unconst:
Go a head an merge

PR #417, opened 2021-07-14
Fix docker image push CI tests

ivanezeigbo:


.github/workflows/docker_image_push.yml
```diff
@@ -44,10 +44,14 @@ jobs:
...
+      uses: docker/login-action@v1 
+      with:
+        username: ${{secrets.DOCKER_USER}}
+        password: ${{secrets.DOCKER_PASS}}
+    - name: Build and Push to Docker
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        push: true
+        tags: bittensor/bittensor:subtensor_tests
```
unconst: why do we have this tag here? bittensor:subtensor_tests? shouldn't it just be bittensor:bittensor.

ivanezeigbo: @unconst do you mean `bittensor/bittensor:latest` ? So following the circle ci config, the push image script seems to target `bittensor/bittensor:subtensor_tests` 

`./scripts/build_protos.sh`
`docker build -t bittensor/bittensor:subtensor_tests -f Dockerfile .`
`docker push bittensor/bittensor:subtensor_tests`

Should I use `bittensor/bittensor:latest` instead?

unconst: Yes I think you should yes. @shibshib  can you comment on this? Or was Parralax the prev owner here

shibshib: Definitely some remnant of parallaxs code, should use bittensor/bittensor:latest

shibshib: I deleted that subtensor thing

ivanezeigbo: made the necessary changes. @unconst @shibshib please review

shibshib (approved):

PR #444, opened 2021-07-30
dataloader update

Eugene-hu:
separate PR for dataloader update: added the wikitext dataset and changed the gateway from infura to pinata. Also added the wikitext test and validation datasets for future testing.

### DIFF
--- a/bittensor/_dataloader/__init__.py
+++ b/bittensor/_dataloader/__init__.py
@@ -33,6 +33,7 @@ class dataloader:
             batch_size: int = None,
             max_corpus_size:int = None,
             num_workers: int = None,
+            dataset: str=None
         ):
         if config == None: config = dataloader.config()
         config = copy.deepcopy( config )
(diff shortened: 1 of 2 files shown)


unconst (approved):

bittensor/_dataloader/__init__.py
```diff
@@ -72,4 +76,5 @@ def check_config( cls, config: 'bittensor.Config' ):
         assert config.dataloader.block_size > 0, 'Block size must be larger than 0'
         assert config.dataloader.max_corpus_size > 0, 'max_corpus_size must be larger than 0'
         assert config.dataloader.num_workers >= 0, 'num_workers must be equal to or larger than 0'
+        assert config.dataloader.dataset in ['genesis','wikitext','test','validation'], 'dataset must be one of the following choices: genesis, wikitext, test, or validation'
```
shibshib, line 75: Perhaps i misunderstood here. Not sure if test and validation should be in the same place/hierarchy as genesis and wikitext. For example, We can divide wiki into test and validation.

But what are test and validation? Are they also datasets or subsets of datasets?

bittensor/_dataloader/dataloader_impl.py
```diff
@@ -36,19 +36,22 @@ def __init__(self):
...
+        self.file_cat = 'https://gateway.pinata.cloud/api/v0/cat'
 
         # Used when current corpus has been exhausted
         self.refresh_corpus = False
 
     @staticmethod
     def requests_retry_session(
-            retries=3,
-            backoff_factor=0.3,
+            retries=5,
```
shibshib, line 53: Any reason for 5 instead of 3?

shibshib (approved):
Looks good overall! Would appreciate some feedback on my comments :)

bittensor/_dataloader/__init__.py
Eugene-hu: test and validation are the test and validation subsets of the wikitext dataset. I left it separate so that when we do run a test/validation run, we can be sure that it was never sampled during training. I could rename it or combine it with wikitext if you want

bittensor/_dataloader/dataloader_impl.py
Eugene-hu: Slightly longer load times to account for possible traffic in IPFS, had a few issues with IPFS being too slow and timing out.

unconst (approved):

PR #451, opened 2021-08-09
Bit 57 local training

Eugene-hu:
Networking update: subscription works with both ipv4 and ipv6 addresses

Miner update: updated template learning parameters, introduced a momentum hyperparameter, and introduced a learning scheduler.  Removed some old debugging print statements


miners/text/sample_config.txt
```diff
@@ -14,14 +14,14 @@ dendrite.max_worker_threads: 150
 dendrite.requires_grad: true
 dendrite.timeout: 5
 
-logging.debug: false
+logging.debug: true
```
shibshib, line 17: Should logging debug be true by default?

miners/text/sample_config.txt
```diff
@@ -14,14 +14,14 @@ dendrite.max_worker_threads: 150
...
 logging.logging_dir: ~/.bittensor/miners/
 logging.record_log: false
 logging.trace: false
 
 miner.device: cpu
-miner.epoch_length: 500
-miner.learning_rate: 0.03
+miner.epoch_length: 100
+miner.learning_rate: 1
```
shibshib, line 24: learning rate 1? isn't that epically high? :D

miners/text/template_miner.py
```diff
@@ -82,9 +82,9 @@ def add_args( parser: argparse.ArgumentParser ):
     def init_weights(self):
         initrange = 0.1
         self.remote_decoder.weight.data.uniform_(-initrange, initrange)
-        self.remote_decoder.bias.data.zero_()
+        #self.remote_decoder.bias.data.zero_()
```
shibshib: If we're not using this, let's just remove it altogether.

shibshib (approved):
Generally looks good, just a couple nits and comments

miners/text/template_miner.py
```diff
@@ -262,8 +264,9 @@ def config() -> 'bittensor.Config':
         # ---- Add miner args.
         parser = argparse.ArgumentParser()
         parser.add_argument('--miner.config', type=str, help='If set, defaults are overridden by passed file.')
-        parser.add_argument('--miner.learning_rate', type=float, help='Training initial learning rate.', default=3e-2)
+        parser.add_argument('--miner.learning_rate', type=float, help='Training initial learning rate.', default=1)
```
ivanezeigbo, line 268: just fyi if we decide that the learning rate of 1 is high

ivanezeigbo (approved):
Just left some comments

isabella618033 (approved):

miners/text/sample_config.txt
Eugene-hu: logging should be turned off by default, but I left it on due to the current bug where it doesn't display any of the subscription logs

miners/text/sample_config.txt
Eugene-hu: yeap, it is extremely high. I actually tuned it down from 5 in the original PyTorch example

miners/text/template_miner.py
Eugene-hu: Shows good, will remove

miners/text/sample_config.txt
```diff
@@ -14,14 +14,14 @@ dendrite.max_worker_threads: 150
...
 miner.name: gpt2_exodus
```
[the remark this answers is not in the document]
Eugene-hu: removed!

unconst (approved):

PR #452, opened 2021-08-11
Bit 65 subscription log

isabella618033:
When it is not in debug mode, logging all messages with level != DEBUG.


unconst (approved):
Nice catch!

PR #457, opened 2021-08-19
Wallet functions

unconst:
Adds the following properties to the wallet:

wallet.uid
-- returns the hotkey uid
wallet.stake
-- returns the hotkey stake
wallet.balance
--returns the coldkey balance

Adds the following functions to the wallet:

wallet.add_stake()
-- adds stake to the hotkey.
wallet.remove_stake()
--removes stake from the hotkey.


bittensor/_subtensor/subtensor_impl.py
```diff
@@ -285,56 +227,63 @@ def subscribe(
...
+            if wait_for_inclusion or wait_for_finalization:
+                response.process_events()
+                if response.is_success:
+                    bittensor.logging.success( 'Staked', str(amount.tao) )
+                    return True
+                else:
+                    bittensor.logging.warning( 'Failed to stake', str(response.error_message) )
+                    return False
+            else:
+                return True
```
shibshib, line 286: Don't think we need an else here, can just return True. Same for the else on line 282.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -377,13 +323,22 @@ def transfer(
...
+            if wait_for_inclusion or wait_for_finalization:
+                response.process_events()
+                if response.is_success:
+                    bittensor.logging.success( 'Transfered', str(amount) + '->' + str(dest) )
+                    return True
+                else:
+                    bittensor.logging.warning( 'Failed to transfer:', str(response.error_message) )
+                    return False
+            else:
+                return True
```
shibshib, line 336: Same comment as above.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -394,29 +349,39 @@ def unstake(
...
+            if wait_for_inclusion or wait_for_finalization:
+                response.process_events()
+                if response.is_success:
+                    bittensor.logging.success( 'Unstaked', str(amount.tao) )
+                    return True
+                else:
+                    bittensor.logging.warning( 'Failed to unstake:', str(response.error_message) )
+                    return False
+            else:
+                return True
```
shibshib, line 384: Same thing.

shibshib (approved):

bittensor/_subtensor/subtensor_impl.py
unconst: mega nit

PR #466, opened 2021-08-31
Bit 37 linterization

isabella618033:


shibshib:
Is there some way we can run linterization tests on our PRs from here on out?

shibshib (approved):

PR #468, opened 2021-09-08
Bit 83 dataloader download from ipfs

isabella618033:
1. dataloader error handling
2. default corpus size to 50MB
3. logger printing non debug messages when debug = False


shibshib (approved):

shibshib:
Great PR @isabella618033 !

PR #485, opened 2021-09-17
Bit 37 lint tokenizer

isabella618033:


bittensor/_tokenizer/__init__.py
```diff
@@ -14,42 +16,44 @@
...
-        tokenizer.add_prefix_space = False
-        tokenizer.add_special_tokens({'bos_token': "[BOS]"}) # A special token representing the beginning of a sentence.
-        tokenizer.add_special_tokens({'eos_token': "[EOS]"}) # A special token representing the end of a sentence.
-        tokenizer.add_special_tokens({'unk_token': "[UNK]"}) # A special token representing an out-of-vocabulary token.
-        tokenizer.add_special_tokens({'sep_token': "[SEP]"}) # A special token separating two different sentences in the same input (used by BERT for instance)
-        tokenizer.add_special_tokens({'pad_token': "[PAD]"}) # A special token used to make arrays of tokens the same size for batching purpose. Will then be ignored by attention mechanisms or loss computation.
-        tokenizer.add_special_tokens({'cls_token': "[CLS]"}) # A special token representing the class of the input (used by BERT for instance).
-        tokenizer.add_special_tokens({'mask_token': "[MASK]"}) # A special token representing a masked token (used by masked-language modeling pretraining objectives, like BERT).
+        """ Return the GPT2 tokenizer with bittersor's special tokens
```
Eugene-hu, line 45: small nit, technically, its special tokens from all other major models (Bert,XLM, etc).

Eugene-hu (approved):
Looks good!

bittensor/_tokenizer/__init__.py
isabella618033: oh it is not removed, just that the name was changed to _tokenizer because "tokenizer" this name is exactly the same as its class name

PR #494, opened 2021-09-22
Fix meta

unconst:
-- fixes the metagraph for bonds matrix


bittensor/utils/weight_utils.py
```diff
@@ -20,6 +20,8 @@
 from typing import Tuple, List
 import torch
 
+U32_MAX = 4294967295
```
shibshib, line 23: Can this not be a sys call?

shibshib (approved):

PR #497, opened 2021-09-25
BIT-123 - quit() is not defined

unconst:


shibshib (approved):

bittensor/_metagraph/metagraph_impl.py
```diff
@@ -344,11 +327,11 @@ def sync ( self, block: int = None ) -> 'Metagraph':
...
-            trust[n.uid] =n.trust / float(1000000000)
-            consensus[n.uid] = n.consensus / float(1000000000)
-            incentive[n.uid] = n.incentive / float(1000000000)
-            inflation[n.uid] =n.inflation / float(1000000000)
-            dividends[n.uid] =n.dividends
+            trust[n.uid] = n.trust / float(1000000000)
+            consensus[n.uid] = n.consensus / float(18446744073709551615)
+            incentive[n.uid] = n.incentive / float(18446744073709551615)
+            inflation[n.uid] = n.inflation / float(1000000000)
+            dividends[n.uid] = n.dividends / float(1000000000)
```
shibshib: I think we should start using constants here. These numbers are confusing because there's no description as to what they are.

PR #498, opened 2021-09-25
BIT-125 -- wallet hotkey function use_password, wallet regen from string not list

unconst:
--- adds regen_coldkey, regen_hotkey, create_coldkey, create_hotkey
--- adds default wallet use_passord
--- allows us to regen from uri
--- allows passing string to mnemonic


Untitled1.ipynb
```diff
@@ -0,0 +1,306 @@
+{
```
shibshib: Why is this notebook here? Should remove, looks like it's for testing purposes.

bittensor/__init__.py
```diff
@@ -21,7 +21,7 @@
 
 # Bittensor code and protocol version.
 __version__ = '1.5.0'
-__version_as_int__ = (100 * 1) + (10 * 4) + (1 * 0)  # Integer representation
+__version_as_int__ = (100 * 1) + (10 * 5) + (1 * 0)  # Integer representation
```
shibshib: Is there a better way to convert to int? I can definitely see the next person updating the version forgetting to update the version on the second line.

Maybe let's do something like 

```
version_split = __version__.split(".")
__version_as_int__ = (100 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
```

Also, I don't see this being used anywhere, so maybe better to just take this line out

shibshib (approved):
Please merge after addressing feedback.

PR #501, opened 2021-09-25
BIT-134 -- forward call to metagraph uses previous incentive function.

unconst:
-- updates the incentive function in the metagraph.


miners/text/template_miner.py
```diff
@@ -303,7 +303,7 @@ def config() -> 'bittensor.Config':
         """
         # ---- Add miner args.
         parser = argparse.ArgumentParser()
-        parser.add_argument('--miner.config', type=str, help='If set, defaults are overridden by passed file.')
+        parser.add_argument('--config', type=str, help='If set, defaults are overridden by passed file.')
```
shibshib: Why the change here? because config is universal and not just to miner?

shibshib (approved):

PR #509, opened 2021-09-29
BIT-185 BIT-182 BIT-183

unconst:
-- adds a registration function.
-- changes all wallet.public_keys to wallet.ss58_address


bittensor/_subtensor/subtensor_impl.py
```diff
@@ -487,12 +483,13 @@ def neurons(self, block: int = None) -> List[SimpleNamespace]:
                 result.append( n )
             return result
 
-    def neuron_for_uid( self, uid: int, block: int = None ) -> Union[ dict, None ]: 
+    def neuron_for_uid( self, uid: int, ss58_hotkey: str, block: int = None ) -> Union[ dict, None ]:
```
Eugene-hu, line 486: So you need both uid and the hokey to get the neuron from the subtensor?

Eugene-hu (approved):
Looks good to me, left you a question

PR #510, opened 2021-09-29
BIT-181- Fixes for the master

Eugene-hu:
Bug fixes for things that passed our tests

### DIFF
--- a/bittensor/_metagraph/metagraph_impl.py
+++ b/bittensor/_metagraph/metagraph_impl.py
@@ -225,7 +225,7 @@ class Metagraph( torch.nn.Module ):
         """
         if self.n.item() == 0:
             return []
-        return [ neuron.hotkey if neuron != None else '' for neuron in self.endpoint_objs ]
+        return [ neuron.hotkey if neuron.hotkey != None else '' for neuron in self.endpoint_objs ]
 
     @property
     def coldkeys( self ) -> List[str]:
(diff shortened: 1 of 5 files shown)


bittensor/_metagraph/metagraph_impl.py
```diff
@@ -247,7 +247,7 @@ def modalities( self ) -> List[str]:
         """
         if self.n.item() == 0:
             return []
-        return [ neuron.modality if neuron != None else '' for neuron in self.endpoint_objs ]
+        return [ neuron.modality if neuron.modality != None else '' for neuron in self.endpoint_objs ]
```
unconst: What is your thinking here? this will fail if the endpoint is Null?

miners/text/template_miner.py
```diff
@@ -371,16 +371,14 @@ def run( self ):
...
                     end_block = start_block + self.config.miner.epoch_length
                     block_steps = [ start_block + block_delta for block_delta in range(start_block, end_block)]
-                    print(block_steps)
                     progress_bar = qqdm( block_steps, total=len(block_steps), desc=format_str('white', f'Epoch:'))
                     for block in progress_bar:
-
                         # --- Iterate over batches until the end of the block.
                         current_block = self.neuron.subtensor.get_current_block()
-                        while block <= current_block:
+                        while block >= current_block:
```
unconst, line 381: This will cause a run away issue where the blocks just progress quickly while the current block is slow.

bittensor/_metagraph/metagraph_impl.py
Eugene-hu: The custom endpoint.__eq__ breaks the comparison in neuron != None, therefore breaking metagraph.hotkeys/coldkeys/modalities

unconst (approved):

PR #512, opened 2021-09-30
BIT 191 fisher info

isabella618033:


bittensor/__init__.py
```diff
@@ -49,7 +49,7 @@
 ]
 
 __registration_servers__ = [
-    '127.0.0.1:5000'
+    '206.189.228.91:5000'
```
shibshib, line 52: We need a DNS here, having an exposed IP address is a bit dangerous.

PR #513, opened 2021-10-01
BIT-193

unconst:
-- small fix for subtensor subscribe caused by calling public key on wallet.


Untitled1.ipynb
```diff
@@ -0,0 +1,306 @@
+{
```
shibshib: Why is this file here? Might want to take it out.

shibshib (approved):
@unconst take out that untitled.ipynb file before you merge.

PR #518, opened 2021-10-05
Remove Rollbar Dependency

ndubuisx:


requirements.txt
```diff
@@ -29,7 +29,6 @@ pytest-xdist==2.2.0
 pytorch_transformers
 pyyaml
 rich
```
shibshib, line 31: Are you sure we're still using this? I'm fairly certain we don't have rich terminal anymore.

shibshib: Same for prettytable, also not being used anymore.

ndubuisx: this PR only removes rollbar - i was going to open another PR with the updated requirements file @shibshib

Eugene-hu (approved):

PR #541, opened 2021-10-15
Bit 224 Bit 225 Bit 228 blacklist update

Eugene-hu:
- Fixes for the config parser
- Gradients are now being normalized so that each sample's gradients sum to one
- Add additional features to the axon security. Users can now define the time between requests. The information containing if the request is forward or a backwards request is now included in metadata and passed to the blacklist callback.
- Removed tests from the test coverage


miners/text/advanced_server.py
```diff
@@ -178,26 +195,26 @@ def blacklist(pubkey:str) -> bool:
...
 
+        # --- creating our chain weights
+        chain_weights =torch.zeros(metagraph.n)
+        uid = metagraph.hotkeys.index( wallet.hotkey.ss58_address )
+        chain_weights[uid] = 1 
+
         while True:
             # --- Run 
-            dataset = iter(dataset.dataloader(epoch_length=config.server.blocks_per_epoch))
+            dataloader = iter(dataset.dataloader(epoch_length=config.server.blocks_per_epoch))
```
shibshib: I had thought @isabella618033 renamed dataloader to dataset, no?

shibshib (approved):

bittensor/_config/__init__.py
```diff
@@ -63,30 +63,21 @@ def __new__( cls, parser: ArgumentParser = None):
...
-                    if not hasattr(section, split_keys[1]):
-                        head = _config
-                        for key in split_keys[:-1]:
-                            if key not in _config:
-                                head[key] = config_impl.Config()
-                            head = head[key] 
-                        head[split_keys[-1]] = arg_val
+            head = _config
+            keys = split_keys
+            while len(keys) > 1:
```
unconst, line 71: An awesome little bit of code.  :)

miners/text/advanced_server.py
```diff
@@ -111,44 +116,62 @@ def backward_text (pubkey:str, inputs_x, grads_dy ):
...
-                    with torch.autograd.set_detect_anomaly(True):
-                        torch.autograd.backward (
-                            tensors = [ gp_server.outputs_cache ],
-                            grad_tensors = [ gp_server.gradients_cache ],
-                            retain_graph=True
-                        )
-                    gp_server.outputs_cache = None
-                    gp_server.gradients_cache = None  
-                    logger.info('Backwards axon gradient applied')
+                with torch.autograd.set_detect_anomaly(True):
```
unconst, line 118: Does this slow things down?

miners/text/advanced_server.py
```diff
@@ -111,44 +116,62 @@ def backward_text (pubkey:str, inputs_x, grads_dy ):
...
+                    torch.autograd.backward (
+                        tensors = [ outputs_y ],
+                        grad_tensors = [ grad ],
+                        retain_graph=True
+                    )
+                logger.info('Backwards axon gradient applied')
                     
         uid = metagraph.hotkeys.index(pubkey)
         priority = metagraph.S[uid].item()/ sys.getsizeof(inputs_x)
```
unconst, line 127: Did we ever agree on a priority mechanism which remembers if the peer is querying us too often? 
something like stake/size * 1/seconds_since_last_call   ?

miners/text/advanced_server.py
```diff
@@ -111,44 +116,62 @@ def backward_text (pubkey:str, inputs_x, grads_dy ):
...
         try:
             future = threadpool.submit(call, input=inputs_x.to( gp_server.device ), grad=grads_dy.to( gp_server.device ),mutex=mutex, priority=priority)
+            
         except concurrent.futures.TimeoutError :
             raise TimeoutError('TimeOutError')
         except Exception as e:
             logger.error('Error found: {}, with message {}'.format(repr(e), e))
 
-    def blacklist(pubkey:str) -> bool:
+    def blacklist(pubkey:str, meta:tuple) -> bool:
```
unconst: It might be cleaner to pass and object or a namespace rather than a tuple, then lines like meta[1].value can be explicit.

i.e. 

meta.public_key

miners/text/advanced_server.py
unconst: She did.

unconst (approved):

miners/text/advanced_server.py
Eugene-hu: Not that I have noticed, but it helps to propagate any errors on the graph back to the main thread.

miners/text/advanced_server.py
Eugene-hu: Not that I could remember, but there is a black list functionality for too many requests by a user.

miners/text/advanced_server.py
Eugene-hu: yes, so here we are calling the dataloader function of the bittensor dataset.

miners/text/advanced_server.py
Eugene-hu: sounds good, I can change it to be more explicit

PR #553, opened 2021-10-22
Bit 229 axon priority

Eugene-hu:
- Axon update with priority queue: Users can now add a priority function to the axon which will set up an priority queue automatically. If not set, axon will simply call the forward and backward function as the requests come.


unconst (approved):

Eugene-hu:
Merged into nobunaga

PR #588, opened 2021-11-04
Bit 253 metagraph size mismatch

isabella618033:


shibshib (approved):
I think @Eugene-hu 's fix has the same self-device update, so might run into a conflict before we merge this if we merge Eugene's first.

PR #603, opened 2021-11-10
fix cli reprompt

ndubuisx:


unconst (commented):
So the previous purpose was to check if the user has set these items. 
If they are set on the command line then we would not prompt them. Unless they are set to defaults of course, because the would trigger the if statement regardless. 

This works better since you check to see if the user has loaded the defaults in by asserting not None, however you are going to lose the defaults mechanism in place. Users should be able to load values through the ENV vars for subtensor and wallet, 
For instance:
BT_SUBTENSOR_NETWORK='nobunaga' will set these arguments cross API, however as implemented here these ENV vars will have no effect. 

I think we should reject this change until we can come up with a better way of specifying this functionality. Perhaps there is a way for us to tell if the arg is set without looking at the default value?

unconst:
Check comments.

PR #620, opened 2021-11-25
Bit 188 dendrite backward request

isabella618033:
**Objective**
Cleaner code for the receptor. Optimization for the backward function.

**1. New Request class**

The request class holds all the inputs/intermediate/output states of a forward/backward request. It works closely with the receptor class only.

`
  class Request():
          # ---- Inputs ----
          self.inputs = inputs
          self.grads_dy = grads_dy
          self.zeros = nill_response_for(inputs)
          
          # ---- Setups ----
          self.modality = modality
          self.backward = backward
          self.start_time = clock.time()
  
          # ---- Intermediate states ---- 
          self.serialized_inputs = None
          self.grpc_request = None
          self.future = None
  
          # ---- Outputs ----
          self.code = None
          self.message = None
          self.outputs = None
`

**2. Refactor the longgg forward / backward call in receptor**

Broke the long function into the following smaller functions

1.     def prerequisite_check(self, request):
2.     def serialization(self, request):
3.     def build_grpc_request(self, request):
4.     def collect_future(self, request):
5.     def check_response(self, request):
6.     def deserialize_forward_response(self, request): /     def deserialize_backward_response(self, request):

**3. Corrected the timing from response**

with text forwarding to 500 peers
image


shibshib (approved):

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -112,20 +108,17 @@ def forward(
         requests = []
         for arg in call_args:
             receptor, inputs, modality = arg
-            requests.append(receptor.forward_prep( inputs = inputs, modality = modality ))
+            requests.append(receptor.preprocess_resquest( inputs = inputs, modality = modality ))
```
shibshib: `preprocess_resquest` seems misspelled. Is this on purpose? Shouldn't it be `preprocess_request`?

shibshib (changes requested):

shibshib (approved):

PR #626, opened 2021-12-06
dont override chain_endpoint if loaded in config

camfairchild:
I was just debugging this
/bittensor/_subtensor/init.py:92
the command line argument gets stored into `config`, but the function argument `chain_endpoint == None` so we override `config.subtensor.chain_endpoint` on line 92 and set it to one of the endpoints on a network (local, nakamoto, etc.)

This makes it 127.0.0.1 on localhost, effectively ignoring the set chain_endpoint
`if (config.subtensor.chain_endpoint == None):` above line 92 would fix


unconst (approved):

unconst:
Thanks for the PR cam

PR #639, opened 2021-12-14
Metagraph sync optimization

shibshib:
- Adds a new IPFS class under `Bittensor` that handles all IPFS calls
- Adds an update to meta graph `sync` function where `cached = True` is default. If `cached == True` then it will pull from IPFS cached neurons. Otherwise, it will manually synchronize like it does today. 
- **NOTE** This caching mechanism *only works for Nakamoto* all other networks will synchronize manually.


Eugene-hu (approved):
LGTM!

PR #655, opened 2022-01-07
Small Bug fixes for the servers

Eugene-hu:
Added parameters --neuron.metagraph_sync and --neuron.set_weights to allow users to determine how often they should set weights and sync the metagraph

### DIFF
--- a/bittensor/_neuron/text/advanced_server/run.py
+++ b/bittensor/_neuron/text/advanced_server/run.py
@@ -192,6 +192,10 @@ def serve( config, gp_server):
 
     nn = subtensor.neuron_for_pubkey(wallet.hotkey.ss58_address)
 
+    # --- last sync block 
+    last_sync_block = subtensor.get_current_block()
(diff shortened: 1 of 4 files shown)


bittensor/_neuron/text/advanced_server/run.py
```diff
@@ -192,6 +192,10 @@ def time_check():
 
     nn = subtensor.neuron_for_pubkey(wallet.hotkey.ss58_address)
 
+    # --- last sync block 
+    last_sync_block = subtensor.get_current_block()
+    last_set_block = subtensor.get_current_block()
```
shibshib: Do we need to call subtensor twice here? why not just do `last_set_block = last_sync_block`?

shibshib (approved):

bittensor/_neuron/text/advanced_server/run.py
Eugene-hu: sounds good, I will change that.

PR #656, opened 2022-01-10
Akatsuki Updates

Eugene-hu:
Major Changes:

1. Improved metagraph sync time

- Downloads cache metagraph from ipfs which improves on the overall metagraph sync by several folds
- Includes the creation of the bittensor.ipfs class

2. Template Server Revamp

- Template server now runs on inference mode. This means that template server requires significantly less memory to run large models
- This comes at a cost of training. Template servers do not train themselves and cannot improve themselves over time. 
- For GPU users, it is also possible to turn on --server.autocast which casts your server to float16, further reducing the memory cost.

3. Validator Fixes

- Fixes the segmenation fault error, and Nonetype issues. Will require consistent monitoring for future stability.

4. Increases pytorch version to 1.10 

- All transformers now contain the batch first argument which matches our data from ipfs.

5 Added metagraph_sync and set_weights parameters for the server


bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -217,6 +217,9 @@ def config ():
         parser.add_argument('--neuron.no_restart', action='store_true', help='if the model should restart', default=False)
         parser.add_argument('--neuron.blacklist.stake.forward', type=float, help='Amount of stake (tao) in order not to get blacklisted for forward requests', default=10)
         parser.add_argument('--neuron.blacklist.stake.backward', type=float, help='Amount of stake (tao) in order not to get blacklisted for backward requests', default=100)
+        parser.add_argument('--neuron.metagraph_sync', type=float, help='how often to sync the metagraph', default=100000)
+        parser.add_argument('--neuron.set_weights', type=float, help='how often to sync set weights', default=10)
```
unconst: What does this flag mean I dont think it is self explanatory. Also setting weights every 10 blocks is way to fast.

unconst (approved):

bittensor/_neuron/text/advanced_server/nucleus_impl.py
Eugene-hu: Sounds good; let me change it to 100 blocks and update the flags

PR #671, opened 2022-01-24
Bit 297 data parallelism

isabella618033:
- Added the ddp functionality, successful to be ran on the roc machine with 3 processes.
- With axon disabled, currenyl only support validator's functionality.
- The local training is essentially faster then the master one, but still skeptical to errors, a fix with testing is upcoming. 
- A downgrade in performance for distillation loss/remote target loss is still under investigation. 
- By fixing the clip_grad_norm with a hook, the downgrad in performance is already a lot less.

**Before clip_grad_norm fix**
image

**After clip_grad_norm fix**
image

**After tuning the clip_norm max value to 0.5**
image


bittensor/_dataset/__init__.py
```diff
@@ -36,8 +36,10 @@ def __new__(
             max_corpus_size:int = None,
             num_workers: int = None,
             dataset_name: list = [],
-            save_dataset: bool=None,
-            no_tokenizer: bool=None
+            save_dataset: bool = None,
+            no_tokenizer: bool = None,
+            world_size: int = 0,
```
Eugene-hu: What is world_size used for? I don't see it being used anywhere

bittensor/_dendrite/__init__.py
```diff
@@ -190,7 +191,7 @@ def manager_serve(cls, config, wallet, receptor_pool = None):
                 max_worker_threads = config.dendrite.max_worker_threads,
                 max_active_receptors = config.dendrite.max_active_receptors
             )
-        ManagerServer.register('get_receptorpool', callable=lambda:receptor_pool,exposed=['forward','backward','get_receptors_state'])
-        manager = ManagerServer(address=('', 50000), authkey=b'abracadabra')
+        ManagerServer.register('get_receptorpool', callable=lambda:receptor_pool,exposed=['forward','backward','get_receptors_state', 'get_total_requests'])
+        manager = ManagerServer(address=('', 4098), authkey=b'abracadabr')
```
Eugene-hu: For security reasons, can we change the authkey to the wallet hotkey?

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -214,7 +214,7 @@ def config ():
         parser.add_argument('--neuron.inter_degree', type=str, help='Interpolate algorithm (nearest | linear | bilinear | bicubic | trilinear | area)', default='nearest')
         parser.add_argument('--neuron.name', type=str, help='Trials for this miner go in miner.root / (wallet_cold - wallet_hot) / miner.name ', default='advanced_server')
         parser.add_argument('--neuron.checking', action='store_false', help='To check if server settings are correct',default=True)
-        parser.add_argument('--neuron.no_restart', action='store_true', help='if the model should restart', default=False)
+        parser.add_argument('--neuron.no_restart', type=bool, help='If True, train the neuron from the beginning', default=True)
```
Eugene-hu: type bool doesn't work well, lets change it to store_false

bittensor/_neuron/text/multitron_validator/neuron_impl.py
```diff
@@ -0,0 +1,484 @@
...
+    
+    def init_process(self, rank):
+        r""" For each process, anchor them to the process group 
+        so that they know how to communication with each other.
+
+        Args:
+            rank (int):
+                rank (id) of the process.
+        """
+        os.environ['MASTER_ADDR'] = self.config.neuron.address
```
Eugene-hu, line 95: Can we make it more specific? Maybe BIT_MASTER_ADDR?

Eugene-hu (approved):
Looks good to me! Left you a few comments

bittensor/_neuron/text/multitron_validator/neuron_impl.py
isabella618033: This one cannot do, cause MASTER_ADDR is the env variable that DDP listens to

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: but then when user type no_restart, they do not want a restart, if we store false, then no no_restart = restart haha

how about we change it back to restart and just remind Jake not to put this flag everywhere XD

bittensor/_dendrite/__init__.py
isabella618033: Goooddd idea! Changed accordingly:)

bittensor/_dataset/__init__.py
isabella618033: Hahaha omg nice catch thank you!

bittensor/_neuron/text/advanced_server/run.py
```diff
@@ -187,7 +186,7 @@ def time_check():
     dataset = bittensor.dataset(config=config)
 
     # load our old model
-    if config.neuron.no_restart != True:
+    if not config.neuron.no_restart :
```
shibshib: Does using `not` instead of `!=` really make a huge difference? I'm just curious.

bittensor/_neuron/text/template_miner/neuron_impl.py
```diff
@@ -122,7 +122,7 @@ def run( self ):
             self.epoch = 0   
 
             # ---- reloads previous run if not restart ----
-            if self.config.neuron.no_restart:
+            if not self.config.neuron.no_restart:
```
shibshib: This is a very important change haha, we had been reloading previous runs before if no_restart = false but now we are reloading previous runs if no_restart = true?

shibshib (approved):

bittensor/_neuron/text/template_miner/neuron_impl.py
isabella618033: TD;LR, i think --no_restart is kind of messy, can we just use --restart?

because no_restart = False means that we should restart, no? 
I have also changed the logic on the default end that no_restart by default would be True (meaning it should not restart),
and users should configure it to restart with --no_restart False (meaning that it should restart) 
but eugene raised a problem where, bool type dosent work well with argparse, so that we should actually store_false for the --no_restart, 
but then when users type --no_restart, wouldnt they expect there is no restart?

bittensor/_neuron/text/advanced_server/run.py
isabella618033: nooo, it was just for aligning with the rest of the code

isabella618033: and seems a little bit easier to read

PR #673, opened 2022-01-28
Bit 336 ddp server

isabella618033:
(1) Added server miner that runs with DDP, there are 2 major class 
class Server 
- get axon requests and pipe them into a forward_queue
- hold the DDPPipe object that process the forward_queue

class DDPPipe
- spawn processes, each of the processes process requests from the forward_queue

(2) Validation
- Dosent seem to have a boost on usage on a  8 GB Memory / 4 AMD vCPUs / 160 GB Disk DO machine for gpt2, possibly because the usage was already being capped / was already enough
- (TBC) Try to host a bigger model on the roc machine to compare the usage


.coveragerc
```diff
@@ -6,3 +6,4 @@ omit =
   ./tests/*
   ./bittensor/_neuron/*
   ./bittensor/_cli/__init__.py
+  ./bittensor/_cli/cli_impl.py
```
shibshib: Did we add a JIRA ticket to put this back in later?

bittensor/_neuron/text/multitron_server/ddp_run.py
```diff
@@ -0,0 +1,375 @@
...
+from torch.nn.utils import clip_grad_norm_
+from datetime import datetime,timedelta
+from threading import Lock
+from torch.nn.parallel import DistributedDataParallel as DDP
+import torch.distributed as dist
+import torch.multiprocessing as mp
+import time
+from multiprocessing import Process, Manager, Event 
+
+os.environ['TOKENIZERS_PARALLELISM'] = 'false'
```
shibshib, line 44: Maybe add a comment here that explains what TOKENIZERS_PARALLELISM means?

shibshib (approved):

PR #679, opened 2022-02-01
Dataset close logic

isabella618033:
Reviewed all the closing logic
- Make sure dataset thread queue / dendrite manager would be closed upon garbage collection or upon keyboard interrupt for miners and validators.

### DIFF
--- a/bittensor/_dendrite/__init__.py
+++ b/bittensor/_dendrite/__init__.py
@@ -41,7 +41,7 @@ class dendrite:
             max_worker_threads: int = None,
             max_active_receptors: int = None,
             receptor_pool: 'bittensor.ReceptorPool' = None,
-            multiprocess: bool = False,
+            multiprocess: bool = None,
             compression: str = None,
         ) -> 'bittensor.Dendrite':
         r""" Creates a new Dendrite object from passed arguments.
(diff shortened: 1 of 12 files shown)


Eugene-hu (approved):

bittensor/_dendrite/__init__.py
```diff
@@ -41,7 +41,7 @@ def __new__(
             max_worker_threads: int = None,
             max_active_receptors: int = None,
             receptor_pool: 'bittensor.ReceptorPool' = None,
-            multiprocess: bool = False,
+            multiprocess: bool = None,
```
shibshib, line 44: Should Bool have a default value of `None`?

shibshib (approved):

shibshib:
Need description here @isabella618033

unconst (approved):

bittensor/_dendrite/__init__.py
isabella618033: hmmm this is for the logic checking below, if multiprocess is default True/False, then we do not know when should we  overwrite the param with the config.

config.dendrite.multiprocessing = multiprocess if multiprocess != None else config.dendrite.multiprocessing

PR #685, opened 2022-02-15
A new core validator, removing others.

unconst:
This PR contains the newly created Core Validator which:
--  uses Shapely values to compute scoring.
-- uses a max/min scaling function to reduce weight skew.

Major Changes:
- Adds a new core validator that replaces the sgmoe and template validator from before. 
- Includes a new scoring/normalization methodology that removes the power law scaling in shapley scores
- Includes a mock setting for subtensor/wallet/dendrite
- The new validator now tries to match some of @opentaco's research with proper positional embedding, and masking.


bittensor/_dendrite/__init__.py
```diff
@@ -86,7 +90,12 @@ def __new__(
                 max_active_receptors = config.dendrite.max_active_receptors,
                 compression = config.dendrite.compression,
             )
-        if config.dendrite.multiprocessing:
+        if config.dendrite._mock:
+            return dendrite_impl.DendriteMock ( 
+                config = config,
+                wallet = wallet
+            )
+        elif config.dendrite.multiprocessing:
```
shibshib: We shouldn't have testing code inside of production code. creating mocks should be left to testing portions of the code, can we move this out?

bittensor/_endpoint/__init__.py
```diff
@@ -139,7 +139,7 @@ def from_tensor( tensor: torch.LongTensor) -> 'bittensor.Endpoint':
 
     @staticmethod
     def dummy():
-        return endpoint_impl.Endpoint(uid=-1,version=0, hotkey = "", ip_type = 4, ip = '0.0.0.0', port = 0, modality= 0, coldkey = "")
+        return endpoint_impl.Endpoint(uid=0, version=0, hotkey = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", ip_type = 4, ip = '0.0.0.0', port = 0, modality= 0, coldkey = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
```
shibshib, line 142: Same concern here, if dummy() is only being used in testing, let's put it in a test file.

bittensor/_metagraph/__init__.py
```diff
@@ -50,10 +51,15 @@ def __new__(
...
                 chain_endpoint (default=None, type=str)
                     The subtensor endpoint flag. If set, overrides the network argument.
+                _mock (:obj:`bool`, `optional`):
+                    For testing, if true the metagraph returns mocked outputs.
         """      
         if config == None: 
             config = metagraph.config()
         config = copy.deepcopy(config)
+        config.metagraph._mock = _mock if _mock != None else config.metagraph._mock
+        if config.metagraph._mock:
```
shibshib, line 62: Same concern again. If we are mocking, let's not put it in a config, this is for testing only. Mock object construction should happen in test code.

bittensor/_metagraph/metagraph_impl.py
```diff
@@ -599,3 +599,348 @@ def __str__(self):
...
+        r""" Initializes a new Metagraph torch chain interface object.
+        """
+        super(MockMetagraph, self).__init__()
+        tn = torch.tensor( 2000, dtype=torch.int64 )
+        tblock = torch.tensor( 0, dtype=torch.int64 )
+        tuids = torch.tensor( list( range( 2000 )), dtype=torch.int64 )
+        tactive = torch.tensor( [ 1 for _ in range (2000)], dtype=torch.int64 )
+        tstake = torch.tensor( [ 1.0 for _ in range (2000) ], dtype=torch.float32 )
+        tranks = torch.tensor(  [1.0/2000 for _ in range (2000) ], dtype=torch.float32 )
+        ttrust = torch.tensor( [ 1.0 for _ in range (2000) ], dtype=torch.float32 )
```
shibshib: Definitely should be its own class inside testing, not in here.

bittensor/_receptor/receptor_impl.py
```diff
@@ -267,7 +267,7 @@ def prerequisite_check(self, request):
             return False, request
         
         # ---- Check endpoint----
-        if self.endpoint.uid == -1:
+        if self.endpoint.hotkey  == 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX':
```
shibshib, line 270: This is a bit weird too, looks like it's checking if we are in test mode. We shouldn't be "checking" if we are in test mode, should just act upon it instead.

bittensor/_subtensor/__init__.py
```diff
@@ -99,7 +99,9 @@ def __new__(
         config = copy.deepcopy( config )
 
         # Returns a mocked connection with a background chain connection.
-        if _mock == True or network == 'mock' or config.subtensor.network == 'mock':
+        config.subtensor._mock = _mock if _mock != None else config.subtensor._mock
+        if config.subtensor._mock == True or network == 'mock' or config.subtensor.network == 'mock':
```
shibshib, line 103: Again, checking if we are in mock mode in production. That's cheating. A mock should fool production code, not the other way around

bittensor/utils/weight_utils.py
```diff
@@ -22,9 +22,26 @@
...
+    shift = 1 / ( multiple - 1 )
+    x = x - x.min()
+    x = x / x.sum()
+    y = (torch.tanh(x * len(x)) + shift)/(torch.tanh( x * len(x) ) + shift).sum()
+    return y
+
 def convert_weight_uids_and_vals_to_tensor( n: int, uids: List[int], weights: List[int] ):
     r""" Converts weights and uids from chain representation into a torch tensor (inverse operation from convert_weights_and_uids_for_emit)
-        Returns:
+        Args:
```
shibshib, line 45: nitpick: should add a 'returns' here.

bittensor/utils/weight_utils.py
```diff
@@ -39,7 +56,7 @@ def convert_weight_uids_and_vals_to_tensor( n: int, uids: List[int], weights: Li
 
 def convert_bond_uids_and_vals_to_tensor( n: int, uids: List[int], bonds: List[int] ):
     r""" Converts bond and uids from chain representation into a torch tensor.
-        Returns:
+        Args:
```
shibshib, line 63: Same nitpick as above, should add a 'returns' here.

shibshib (changes requested):

bittensor/_endpoint/__init__.py
unconst: Not only in testing, we also fill these null objects when there is a parsing error from the chain.

bittensor/_dendrite/__init__.py
unconst: Lets redo the mocking but lets do it at a later date.

bittensor/_receptor/receptor_impl.py
unconst: Its not so much of a testing endpoint but a 'null' endpoint. We put the check here because it keeps all the other code the same, rather than say filtering at a higher level, you actually query the null endpoint as usual.

shibshib:
This branch is out of date with master and coverage is down by 5.8%, let's please fix that first before merging. And we need a ticket to fix these mocks in the critical path.

unconst:
Running with new coverage

shibshib (approved):

PR #688, opened 2022-02-18
Fixes Hotkey Ordering on Overview

camfairchild:
fixes hotkey ordering issue

### Description of the Change

Fixes the hotkey ordering issue in the overview
### Verification Process

Ran locally and verified using `btcli list`

### Release Notes

- Fixed overview hotkey ordering

### DIFF
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -391,7 +391,7 @@ class CLI:
         total_dividends = 0.0
         total_emission = 0     
         for nn, hotwallet in tqdm(list(zip(neurons,all_hotkeys))):
-            if nn.is_null: # skip unregistered neurons
+            if nn.is_null: # skip unregistered hotkeys
                 continue
             uid = nn.uid
             active = nn.active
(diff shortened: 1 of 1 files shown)


bittensor/_cli/cli_impl.py
```diff
@@ -379,8 +379,7 @@ def overview(self):
         with console.status(":satellite: Syncing with chain: [white]{}[/white] ...".format(self.config.subtensor.network)):
             for wallet in tqdm(all_hotkeys):
                 nn = subtensor.neuron_for_pubkey( wallet.hotkey.ss58_address )
-                if not nn.is_null:
-                    neurons.append( nn )
+                neurons.append( nn )
```
unconst, line 382: This line is needed to keep to remove de-registered neurons from the sync? What is the justification for removing it?

bittensor/_cli/cli_impl.py
```diff
@@ -431,7 +432,7 @@ def overview(self):
         table.title = (
             "[white]Wallet - {}:{}".format(self.config.wallet.name, wallet.coldkeypub.ss58_address)
         )
-        table.add_column("[overline white]HOTKEY",  str(total_neurons), footer_style = "overline white", style='bold white')
+        table.add_column("[overline white]HOTKEY NAME",  str(total_neurons), footer_style = "overline white", style='bold white')
```
unconst, line 435: More specific, ok

unconst (approved):

bittensor/_cli/cli_impl.py
camfairchild: the line removes the neurons after connecting with the chain. 
the problem is this avoids removing the neuron, but not the hotkey
I moved the check to inside the table creation to avoid mismatching the hotkeys
Instead, I could remove the hotkey from the all_hotkeys but before the table, but complexity would be the same I think

PR #690, opened 2022-02-18
update peer weight when hotey get updated 

isabella618033:
- update the peer_weight with the mean when a hotkey got changed during metagraph update

### DIFF
--- a/bittensor/_neuron/text/neuron_utilities.py
+++ b/bittensor/_neuron/text/neuron_utilities.py
@@ -9,19 +9,11 @@ def update_metagraph_peerweight(metagraph, nucleus, device):
     metagraph.sync()
     new_hotkeys = metagraph.hotkeys
     peer_weight_mean = torch.mean(nucleus.peer_weights)
-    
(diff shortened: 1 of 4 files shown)


bittensor/_neuron/text/neuron_utilities.py
```diff
@@ -1,8 +1,33 @@
...
+        Args:
+            metagraph (:obj:`bittensor.metagraph`, `required`):
+                The metagraph to sync.
+            nucleus (:obj:`bittensor.neuron.text.nucleus`, `required`):
+                The nn.Module class that needs the peerweight to be updated.
+            device (:type:`torch.device`)
+                The device where peer_weight should be stored. 
+    """ 
+    old_hotkeys = metagraph.hotkeys
+    metagraph.sync()
```
Eugene-hu, line 20: This will sync the metagraph in the neuron as well, correct?

Eugene-hu (approved):
LGTM

bittensor/_neuron/text/neuron_utilities.py
isabella618033: yea, cause they are the same

PR #695, opened 2022-02-28
WIP: Strict validators

unconst:
This extension to the core validator ensures that the user sets the required number of weights on the chain, and also ensures that the max/min ratio is allowed. 

Changes:
-- new core validator compatible with new chain code
-- new template miner compatible with new chian code
-- minor text fixes in the cli
-- adds the query command into the cli


bittensor/_neuron/text/template_miner/__init__.py
```diff
@@ -1,161 +1,702 @@
...
+            Arguments reflect an RPC backward request from another miner in the network. No response
+            needed for tokenized text inputs (uint64s have no gradient).
+
+            Args:
+                inputs_x ( :obj:`torch.Tensor`, `required`):
+                    torch inputs from previous forward call.
+                grads_dy ( :obj:`torch.Tensor`, `required`):
+                    torch grads of forward output.                    
+        """
+        if self.config.neuron.accumulate_remote_gradients:
```
Eugene-hu: accumulate_remote_gradients is not part of the config

Eugene-hu: I am also fairly certain that this old backward function does not work. It might be better to remove it and have the template miner only respond to forward requests

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -324,10 +330,11 @@ def forward(self, x: torch.tensor) -> torch.tensor:
 class nucleus( torch.nn.Module ):
     """ Nucleus class which holds the validator model.
     """
-    def __init__( self, config, device ):
+    def __init__( self, config, device, subtensor ):
         super(nucleus, self).__init__()
         self.config = config
         self.device = device
+        self.max_n = subtensor.max_n
```
isabella618033, line 341: have we tested on nobunaga with a changing max_n value to make sure a changing size in the validator would not break anything?

bittensor/_neuron/text/template_miner/__init__.py
```diff
@@ -1,161 +1,702 @@
...
+                '\n\t until_reset:', self.epoch % epochs_until_reset, '\n\t current_block:', current_block, '\n')
+        if self.config.using_wandb:
+            wandb.log( {    'batch_size': batch_size, 'sequence_length': sequence_length, 'n_topk_peer_weights': n_topk_peer_weights, 
+                            'max_allowed_ratio': max_allowed_ratio, 'blocks_per_epoch': blocks_per_epoch, 'epochs_until_reset': epochs_until_reset, 
+                }, step = current_block )
+
+        # === Reset Epochs with new params. ===
+        # Pulls new default validator training parameters and resets 
+        # the model and dataset for the following epoch.
+        if self.epoch % epochs_until_reset == 1:
```
Eugene-hu: Should this be == 1? I think it is 0 on the core validator.

Eugene-hu (approved):
Looks fairly good to me; left you a few suggestions!

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: also, do we have to change the n in metagraph as well? or it dosent matter

bittensor/_neuron/text/template_miner/__init__.py
unconst: Can we make this fix and allow gradients rather than remove
should simply require the config item

bittensor/_neuron/text/template_miner/__init__.py
unconst: yes

PR #701, opened 2022-03-11
Validator updates

unconst:
-- adds proper wandb log sorting
-- increases the dendrite number of active receptors 
-- decreases default dataset loading


bittensor/_dataset/__init__.py
```diff
@@ -158,6 +158,7 @@ def add_defaults(cls, defaults):
         defaults.dataset.data_dir = os.getenv('BT_DATASET_DATADIR') if os.getenv('BT_DATASET_DATADIR') != None else '~/.bittensor/data/'
         defaults.dataset.save_dataset = os.getenv('BT_DATASET_SAVE_DATASET') if os.getenv('BT_DATASET_SAVE_DATASET') != None else False
         defaults.dataset.max_datasets = os.getenv('BT_DATASET_max_datasets') if os.getenv('BT_DATASET_max_datasets') != None else 3
+        defaults.dataset.num_batches = os.getenv('BT_DATASET_num_batches') if os.getenv('BT_DATASET_num_batches') != None else 500
```
Eugene-hu, line 161: We dropped the default value down to 150. Do we want to push it back up to 500?

Eugene-hu (approved):
LGTM! Left you a question

isabella618033 (approved):

bittensor/_dataset/__init__.py
isabella618033: just FYI, if a batch take 12 sec, then 150 batches would be every 30 mins, and 500 would be every ~1.5 hr

PR #703, opened 2022-03-14
Speed update core

unconst:
Bug fix and speed update for validators.
-- we use a mat multiplication to speed up the gate calculations
-- we remove and noise terms to the shapely scoring which could possibly help non returning peers.


Eugene-hu (approved):

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -452,41 +450,39 @@ def forward (
...
         # We are taking the topk routing weights and returning their uids.
         # First we ensure topk is smaller than the network size then use the torch.topk.
         # topk_routing_weights: (torch.float64): scores of uids with highest scores.
-        # topk_routing_weights.shape = [ real_topk ]
+        # topk_routing_weights.shape = [ self.config.nucleus.topk ]
         # topk_routing_uids: (torch.LongTensor): uids with highest scores.
-        # topk_routing_uids.shape = [ real_topk ]
-        real_topk = min( len( active_uids ), self.config.nucleus.topk )
-        top_k_routing_weights, routing_uids = torch.topk( noisy_routing_weights, real_topk, dim=0)
+        # topk_routing_uids.shape = [ self.config.nucleus.topk ]
```
shibshib, line 470: Maybe let's just delete this if we're not using it?

shibshib (approved):

PR #705, opened 2022-03-16
Dataset revamp

isabella618033:
- better fallback logic
- can take a user-define dataset now
            - 1. add your dataset folder to ~/.bittensor/data/<dataset-name>
            -  2. set the flag --dataset.name <dataset-name>
- separate dataset from dataloader
- cleared all max_corpus_size check / attributes


bittensor/_dataset/dataset_impl.py
```diff
@@ -80,7 +81,7 @@ def requests_retry_session(
         session.mount('https://', adapter)
         return session
 
-    def retrieve_directory(self, address: str, params = None, action: str = 'post', timeout : int = 180):
+    def get_ipfs_file(self, address: str, file_meta: dict, action: str = 'post', timeout : int = 180):
```
shibshib: Is this now only exclusively getting files and not directories?

bittensor/_dataset/dataset_impl.py
```diff
@@ -169,46 +175,141 @@ def __del__(self):
     def close(self):
         self.data_queue.close()
 
-    def get_random_directories(self):
-        r""" Getting directories from a random dataset_hash
-        Where a directory could be leading to a data file or a directory file 
+    def get_folder_size(self, folder):
```
shibshib, line 178: Can we add some docustrings here? Other functions are documented, let's keep it consistent :)

bittensor/_dataset/dataset_impl.py
```diff
@@ -169,46 +175,141 @@ def __del__(self):
...
+
+        full_path = os.path.expanduser(os.path.join(self.data_dir, file_meta['Folder'], file_meta['Hash']))
+        if os.path.exists(full_path):
+            try:
+                with open(full_path, mode='r') as f:
+                    text = f.read()
+
+                logger.success("Loaded from disk:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
+            except Exception:
+                pass
```
shibshib: We should log the exception here, it'll help in debugging anyways.

bittensor/_dataset/dataset_impl.py
```diff
@@ -169,46 +175,141 @@ def __del__(self):
...
+
+            return text
+        else:
+            return None
```
shibshib: No need for an `else: return None` here, can just do `return None` outside the if statement.

bittensor/_dataset/dataset_impl.py
```diff
@@ -169,46 +175,141 @@ def __del__(self):
...
+
+        Return:
+            hashes (list):
+                The hashes from the dataset downloaded from disk or IPFS.     
+        """
+        # --- Load text from path
+        logger.success( f"Getting dataset: {file_meta['Name']}" )
         
-        return directories
+        hashes = None
```
shibshib: Is `hashes = None` necessary since you're setting it in the next line? `self.load_hash` should come back with a None if it doesn't work anyways

bittensor/_dataset/dataset_impl.py
```diff
@@ -222,26 +323,52 @@ def get_directories(self, keys: list):
                     directory: Map{ Name: str, Hash: str, Size: int }: 
                         A random directory that lead to a datafile.
         """
+        def get_hashes(dataset_meta):
+
+            if self.IPFS_fails > 10:
```
shibshib: should have a global variable or something to indicate the number of fails we are willing to tolerate, it's best not to use an integer in an if statement, will be hard to find later :)

bittensor/_dataset/dataset_impl.py
```diff
@@ -222,26 +323,52 @@ def get_directories(self, keys: list):
...
+                for sub_directory in sub_directories:
+                    sub_directory['Folder'] = dataset_meta['Name']
+            else:
+                sub_directories = self.get_dataset(dataset_meta)
+
+            if sub_directories != None:
+                return sub_directories
+
+            else:
+                return []
```
shibshib: Same as before, no need for `else:`, can just `return []`

bittensor/_dataset/dataset_impl.py
```diff
@@ -273,59 +400,56 @@ def extract_datafile_dir(self, directory):
...
-        if os.path.exists(full_path):
+        for text_file in files:
+            # --- Get text from the datafile directory
             try:
-                with open(full_path, mode='r') as f:
-                    text = f.read()
-                logger.success("Loaded:".ljust(20) + "<blue>{}</blue>".format(file_name))
-            except Exception:
-                logger.warning("Load failed:".ljust(20) + "<blue>{}</blue>".format(file_name))
+                text = self.load_hash(text_file)
```
shibshib: Not sure I understand the reasoning for try and catch here? Wouldn't `self.load_hash(text_file)` return None anyway? Or does it have the potential to throw an exception?

bittensor/_dataset/dataset_impl.py
```diff
@@ -359,34 +480,42 @@ def construct_text_corpus(self, min_data_len = 0):
...
                         pass
 
                     # --- Get text from the datafile directory
                     try:
                         text = self.get_text(random_datafile_dir)
+
                     except: 
                         text = None
-
+
```
shibshib: Same as earlier, try and except are needed? Or is get_text likely to just throw an exception?

shibshib:
Also, we need a description ;)

bittensor/_dataset/dataset_impl.py
isabella618033: Yea, try except is not needed, get_text itself is already secured. Removing this try except logic.

bittensor/_dataset/dataset_impl.py
isabella618033: Try except is not needed here, load_hash itself is already secured. Removing this try except logic.

bittensor/_dataset/dataset_impl.py
isabella618033: Okay, renaming it to get_ipfs_directory.

shibshib (approved):

PR #707, opened 2022-03-16
Adds multiprocessing to registration solving

camfairchild:
Adds multiprocessing to the registration solver. 
Note: 
    - We should modify the number of processes based on user input.
    - We can also modify the update interval to do smaller blocks of work,
        while still updating the block information after a different number of nonces,
        to increase the transparency of the process while still keeping the speed.
        
  Would be good if we had some flags for user configuration. Current default is 5 processes and update interval of 100,000 nonces before block information is re-polled

### DIFF
--- a/bittensor/utils/__init__.py
+++ b/bittensor/utils/__init__.py
@@ -10,8 +10,7 @@ import time
 import torch
 import numbers
 import pandas
-import sys
-from typing import Tuple, List, Union, Optional
+from typing import Any, Tuple, List, Union, Optional
 
 
 def indexed_values_to_dataframe ( 
@@ -98,12 +97,14 @@ def solve_for_difficulty( block_hash, difficulty ):
             break
     return nonce, seal
 
-def solve_for_difficulty_fast( subtensor, num_processes: int = 5, update_interval: int = 100000 ) -> None:
+def solve_for_difficulty_fast( subtensor, wallet, num_processes: int = 5, update_interval: int = 100000 ) -> Tuple[int, int, Any, int, Any]:
     """
     Solves the POW for registration using multiprocessing.
     Args:
         subtensor
             Subtensor to connect to for block information and to submit.
+        wallet:
+            Wallet to use for registration.
         num_processes: int
             Number of processes to use.
         update_interval: int
--- a/bittensor/utils/__init__.py
+++ b/bittensor/utils/__init__.py
@@ -133,7 +133,8 @@ def solve_for_difficulty_fast( subtensor, wallet, num_processes: int = 5, update
         best.raw = struct.pack("d", float('inf'))
         best_seal = multiprocessing.Array('h', 32, lock=True) # short array should hold bytes (0, 256)
         with multiprocessing.Pool(processes=num_processes, initializer=initProcess_, initargs=(solve_, found_solution, best, best_seal)) as pool:
-            while found_solution.value == -1: # no solution found
+            # while no solution found and wallet has not been registered
+            while found_solution.value == -1 and not wallet.is_registered(subtensor):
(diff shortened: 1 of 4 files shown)


bittensor/utils/__init__.py
```diff
@@ -95,42 +98,43 @@ def solve_for_difficulty( block_hash, difficulty ):
...
-            if product - limit < best:
-                best = product - limit
-                best_seal = seal
-
-            if product < limit:
-                return nonce, block_number, block_hash, difficulty, seal
-
-            if nonce % update_interval == 0:
+    with console.status("Solving") as status:        
+        found_solution = multiprocessing.Value('q', -1, lock=False) # int
```
unconst, line 131: I think the subtensor connection is not multiprocess safe. Have you checked this.

bittensor/utils/__init__.py
```diff
@@ -95,42 +98,43 @@ def solve_for_difficulty( block_hash, difficulty ):
...
+        best = multiprocessing.Value(ctypes.c_char_p, lock=True) # byte array to get around int size of ctypes
+        best.raw = struct.pack("d", float('inf'))
+        best_seal = multiprocessing.Array('h', 32, lock=True) # short array should hold bytes (0, 256)
+        with multiprocessing.Pool(processes=num_processes, initializer=initProcess_, initargs=(solve_, found_solution, best, best_seal)) as pool:
+            while found_solution.value == -1: # no solution found
```
unconst: Extra work, maybe not for this PR, but we should probably be checking to see if the wallet has been registered during this process. For instance, if another computer computing the registration solves it, then this should close down. However this would require adding this tool kit into the subtensor code base which is already a bit verbose. Food for thought.

unconst (approved):

bittensor/utils/__init__.py
unconst: If we did do this, we can add the registration arguments into --subtensor. which would give it a good namespace.

bittensor/utils/__init__.py
camfairchild: I have not.
However,
The multiprocessing stops every update_interval as the block of work is completed. The loop then checks if a solution was found (any solution) and submits the solution.
Therefore, this doesn't access the subtensor while multiprocessed.

bittensor/utils/__init__.py
camfairchild: Good idea. Though there's a registration check used before registration begins so is this not already in the tooling?

https://github.com/opentensor/bittensor/blob/36f4bdc902f6c4c581588d1fc73489d616dd16eb/bittensor/_subtensor/subtensor_impl.py#L489-L493

bittensor/utils/__init__.py
unconst: Ok, that's good. However you can also just create a new connection in each process if you need to.

bittensor/utils/__init__.py
unconst: Yes, but you can register from another place during this sequence. The check passes but should fail later when the POW is solved elsewhere.

bittensor/utils/__init__.py
```diff
@@ -95,42 +98,43 @@ def solve_for_difficulty( block_hash, difficulty ):
...
+                result = pool.starmap(solve_, iterable=[(nonce, block_bytes, difficulty, block_hash, block_number, limit) for nonce in range(update_interval)])
```
unconst, line 138: Ok, I see, this is very clean

bittensor/utils/__init__.py
unconst: i..e if you regsitered from another computer, then this falls into an infinite loop given the "neuron already in the actvie set error"

-- we should probably check the error responses also but this is a different PR

camfairchild:
May address #709 @isabella618033

unconst (approved):

PR #710, opened 2022-03-17
Adds coldkeygen from seed instead of mnemonic

camfairchild:
Allows generating a coldkey using regen_coldkey from a seed.
This is useful in allowing vanity key generation. 
Could be duplicated for hotkey generation, but I don't find this is particularly useful.


shibshib:
When do we get to review this one?

camfairchild:
> When do we get to review this one?

please review 🙏🏼

shibshib (approved):

PR #711, opened 2022-03-17
makes default network 'local'

camfairchild:
Changes the documentation as well
~~One-liner.~~
~~Couldn't find any tests to modify.~~


camfairchild:
looks like the metagraph test doesn't explicitly set a network so this change breaks it

tests/unit_tests/bittensor_tests/test_wandb.py
```diff
@@ -27,5 +27,7 @@ def test_dendrite():
     del dendrite
 
 def test_metagraph():
-    metagraph = bittensor.metagraph()
-    metagraph.to_wandb()
\ No newline at end of file
+    config = bittensor.subtensor.config()
+    subtensor = bittensor.subtensor(network='nakamoto', config=config, )
```
shibshib: It's not good practice to connect to the production network during testing. Let's either change this to `Nobunaga` which is our test net or change use the mocked network.

shibshib (approved):

README.md
```diff
@@ -184,11 +184,12 @@ $ btcli unstake
 ```
 
 ### 4.2. Selecting the network to join 
-There are two open Bittensor networks: Nobunaga, Akatsuki, Nakamoto.
+There are two open Bittensor networks: Nobunaga, Akatsuki, Nakamoto, Local.
```
shibshib: While we're at it, can we take out "akatsuki" this is pretty old and we don't have it anymore?

bittensor/_subtensor/__init__.py
```diff
@@ -137,7 +138,7 @@ def __new__(
             config.subtensor.network = bittensor.defaults.subtensor.network
            
         substrate = SubstrateInterface(
-            address_type = 42,
+            ss58_format = 42,
```
shibshib, line 141: Any reason for this change?

bittensor/_subtensor/subtensor_mock.py
```diff
@@ -62,7 +62,7 @@ def mock(cls):
         endpoint = bittensor.__mock_entrypoints__[0]
         port = int(endpoint.split(':')[1])
         substrate = SubstrateInterface(
-            address_type = 42,
+            ss58_format = 42,
```
shibshib, line 65: same comment as above.

bittensor/_subtensor/__init__.py
camfairchild: There was a deprecation warning for this by the substrate library. Figured I would update it.

README.md
camfairchild: Do you want me to remove it from the selection prompt too, or just the README?

PR #716, opened 2022-03-21
Fix for dockerfile

mrseeker:
Dockerfile does not install python package (bitcli does not work), this should fix it.


Dockerfile
```diff
@@ -30,5 +30,6 @@ WORKDIR /bittensor
 RUN pip install --upgrade numpy pandas setuptools "tqdm>=4.27,<4.50.0" wheel
 RUN pip install -r requirements.txt
 RUN pip install -e .
+RUN pip install .
```
shibshib, line 32: the previous command `RUN pip install -e .` does the same thing, not sure why we need this line?

mrseeker: This is a sneaky pip issue: the -e installs the application, but it does not install it as an importable module. When you call btcli, it will complain about bittensor not being installed as module.

shibshib: Interesting. Would it make more sense to just take out the `pip install -e .` and keep `pip install .`?

mrseeker: Just looking at it, the -e variation might indeed be better to take out, however I am not that familiar with setuptools and if it's installing it with or without the btcli.

shibshib: It should install it with regardless. Feel free  to take this out and I'll merge it :)

joeylegere (approved):

PR #721, opened 2022-03-24
multiprocessing proof of work fix

isabella618033:
- fixed the not speeding up problem 
- fixed the pool processes not starting up stably problem

### DIFF
--- a/tests/unit_tests/bittensor_tests/utils/test_utils.py
+++ b/tests/unit_tests/bittensor_tests/utils/test_utils.py
@@ -168,7 +168,7 @@ def test_solve_for_difficulty_fast_registered_already():
         block_hash = '0xba7ea4eb0b16dee271dbef5911838c3f359fcf598c74da65a54b919b68b67279'
         subtensor = MagicMock()
         subtensor.get_current_block = MagicMock( return_value=1 )
-        subtensor.difficulty = 2560000000 # set high to make solving take a long time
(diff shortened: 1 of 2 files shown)


shibshib (approved):

unconst (approved):

tests/unit_tests/bittensor_tests/utils/test_utils.py
```diff
@@ -168,7 +168,7 @@ def test_solve_for_difficulty_fast_registered_already():
         block_hash = '0xba7ea4eb0b16dee271dbef5911838c3f359fcf598c74da65a54b919b68b67279'
         subtensor = MagicMock()
         subtensor.get_current_block = MagicMock( return_value=1 )
-        subtensor.difficulty = 2560000000 # set high to make solving take a long time
+        subtensor.difficulty = 100000000000000000000000000 # set high to make solving take a long time
```
shibshib: I probably missed something here, why is difficulty changed so high?

isabella618033: This test is about if the pow can stop when the wallet got registrated somewhere else.

So the difficulty was supposed to be a number that is too big to get solved so that the pow can never complete and the registration can only get finished because the wallet got registered somewhere. Wanted to set it to inf, but python int is unbounded.

PR #723, opened 2022-03-24
fix aonx config

unconst:
The config is not being passed to the aonx


Eugene-hu (approved):
Nice catch!

shibshib:
Do you have something against spelling `axon` correctly?

shibshib (approved):

PR #724, opened 2022-03-24
Subtensor register update

Eugene-hu:
Major Changes:
- Retry with pow if within 3 blocks of solving
- Changed the max attempts from 10 attempts to 3
- To be merged after [https://github.com/opentensor/bittensor/pull/721](url)

### DIFF
--- a/bittensor/_subtensor/subtensor_impl.py
+++ b/bittensor/_subtensor/subtensor_impl.py
@@ -475,12 +475,13 @@ To run a local node (See: docs/running_a_validator.md) \n
             with bittensor.__console__.status(":satellite: Registering...({}/{})".format(attempts,max_allowed_attempts)) as status:
 
                 # pow failed
-                if pow_result is None:
+                if not pow_result:
                     # might be registered already
                     if (wallet.is_registered( self )):
                         bittensor.__console__.print(":white_heavy_check_mark: [green]Registered[/green]")
                         return True
                     
+                # pow successful, proceed to submit pow to chain for registration
                 else:
                     #check if pow result is still valid
                     while pow_result['block_number'] >= self.get_current_block() - 3:
@@ -512,7 +513,8 @@ To run a local node (See: docs/running_a_validator.md) \n
                                 time.sleep(1)
                                 continue
                             
-                            elif response.is_success:
+                            # Successful registration, final check for neuron and pubkey
+                            else:
                                 bittensor.__console__.print(":satellite: Checking Balance...")
                                 neuron = self.neuron_for_pubkey( wallet.hotkey.ss58_address )
                                 bittensor.__console__.print(":white_heavy_check_mark: [green]Registered[/green]")
--- a/bittensor/_subtensor/subtensor_impl.py
+++ b/bittensor/_subtensor/subtensor_impl.py
@@ -808,6 +808,10 @@ To run a local node (See: docs/running_a_validator.md) \n
                 response.process_events()
                 if response.is_success:
                     bittensor.__console__.print(":white_heavy_check_mark: [green]Finalized[/green]")
+                    block_hash = response.block_hash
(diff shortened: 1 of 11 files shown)


bittensor/_subtensor/subtensor_impl.py
```diff
@@ -467,61 +468,64 @@ def register (
...
-                            bittensor.__console__.print(":cross_mark: [red]Failed[/red]: error:{}".format(response.error_message))
-                            attempts += 1
-                            if attempts > max_allowed_attempts: 
-                                bittensor.__console__.print( "[red]No more attempts.[/red]" )
-                                return False
-                            else:
-                                status.update( ":satellite: Registering...({}/10)".format(attempts))
+                else:
+                    #check if pow result is still valid
+                    while pow_result['block_number'] >= self.get_current_block() - 3:
```
isabella618033, line 487: maybe add a sleep here? 
cause otherwise we could be sending hundred thousands of substrate calls in that 36 second

isabella618033 (approved):

bittensor/_subtensor/subtensor_impl.py
Eugene-hu: Nice catch! Added

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -467,61 +468,65 @@ def register (
...
             pow_result = bittensor.utils.create_pow( self, wallet )
-            with bittensor.__console__.status(":satellite: Registering...({}/10)".format(attempts)) as status:
-                # verify wallet not already registered
-                if (wallet.is_registered( self )):
-                    bittensor.__console__.print(":white_heavy_check_mark: [green]Registered[/green]")
-                    return True
+            with bittensor.__console__.status(":satellite: Registering...({}/{})".format(attempts,max_allowed_attempts)) as status:
+
+                # pow failed
+                if pow_result is None:
```
shibshib: nit: can probably get away with `if not pow_result:`

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -467,61 +468,65 @@ def register (
...
-                        response.process_events()
-                        if not response.is_success:
-                            bittensor.__console__.print(":cross_mark: [red]Failed[/red]: error:{}".format(response.error_message))
-                            attempts += 1
-                            if attempts > max_allowed_attempts: 
-                                bittensor.__console__.print( "[red]No more attempts.[/red]" )
-                                return False
-                            else:
-                                status.update( ":satellite: Registering...({}/10)".format(attempts))
+                else:
```
shibshib, line 485: no need for `else` here if we are returning True if wallet is registered. Can just proceed.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -467,61 +468,65 @@ def register (
...
-                        status.update( ":satellite: Registering...({}/10)".format(attempts))
-                        continue
-
-            if response.is_success:
-                with bittensor.__console__.status(":satellite: Checking Balance..."):
-                    neuron = self.neuron_for_pubkey( wallet.hotkey.ss58_address )
-                    bittensor.__console__.print(":white_heavy_check_mark: [green]Registered[/green]")
-                    return True
+                            
+                            elif response.is_success:
```
shibshib: why is this an `elif`? response.is_success can be None or have a value, no?

shibshib (approved):

bittensor/_subtensor/subtensor_impl.py
Eugene-hu: This else is actually for the pow, not for if wallet is registered. We skipped substrate call if pow fails and retry the pow instead. Let me add a comment to make it more clear

PR #730, opened 2022-03-31
Bit 390 shapely speed up

isabella618033:
- method: joining multiple masked contents to go through forward and calculate the loss together
- result: 50% speed up for the shapely calculation
- test: verified that the resulting shapely score is the same as the original implementation


Eugene-hu (approved):
LGTM!

PR #731, opened 2022-03-31
Bit 384 Validator optimization

isabella618033:
**Method**
(1) On the background, run n times forward functions multithreaded (default n = queue_size = 5)
(2) The results of the forward function would be put into a queue. 
(3) The results got from the queue would be used for shapely calculation. 
(4) After all results got piped out (after 5 times calculation of shapely), run backward and optimization to update the model parameters. 
(5) Repeat at (1)

**Result**
As shapely calculation is still the bottleneck, the speed up would be n times faster. 

**Test**
(not done) would like to test it overnight to make sure it runs properly for long
(not done) would like to add metric to measure if the speed up is really n times

**Note**
Actually I think it might be better if we can fix the dendrite/grpc first before merging this in, cause this pr is adding another level of complication, which we might not want before making sure that things are working fine. I will also move my priorities to grpc/dendrite before carrying on on this pr. But this pr is mostly ready, so feel free to check it.


shibshib (approved):

shibshib:
@isabella618033 approved you, but resolve conflicts first :)

PR #735, opened 2022-04-04
Adds random nonce initialization

unconst:


shibshib:
@unconst for the love of god a description :D

shibshib (approved):

camfairchild:
@unconst Seems tests are failing because of an integer overflow. 
mp package uses ctypes with the largest int long long, 32 bits.
Right now it's q for signed long long. So max would be 2^32 - 1

Solution is to limit the nonces below that, or use mp Array and some bit shifting to store nonces.

PR #740, opened 2022-04-08
Version checking

Eugene-hu:
- Quick PR that notifies the user if their bittensor is out of date. It will require us to update pip version. 
- Also adds in `btcli update` which can be used to pull the latest version of bittensor

image


isabella618033 (approved):
LGTM! 

I think even better, in the version_checking(), we might want to ask the user "do you want to update?" and wait for their respond and do an update for them, or else they might just skip reading it.

but it also create problems where the user might have to run the miner again :o

shibshib (approved):

shibshib:
LGTM but let's make sure that we update the pip version before we merge this

bittensor/utils/__init__.py
```diff
@@ -226,3 +227,12 @@ def create_pow( subtensor, wallet ):
         'block_hash': block_hash, 
         'work': binascii.hexlify(seal)
     }
+
+def version_checking():
+    response = requests.get(f'https://pypi.org/pypi/bittensor/json')
```
joeylegere: Why is this an f-string? Also this is probably a good candidate for something that should live in `constants.py` or something similar.

bittensor/_cli/cli_impl.py
```diff
@@ -212,6 +215,11 @@ def help ( self ):
             bittensor.neurons.multitron_server.neuron().run()
 
 
+    def update ( self ):
+        if self.config.no_prompt or self.config.answer == 'Y':
+            os.system(' (cd ~/.bittensor/bittensor/ ; git checkout master ; git pull --ff-only )')
+            os.system('pip install -e ~/.bittensor/bittensor/')
```
joeylegere, line 221: This is out of scope for this PR, but I see this path in a few places and I don't think it should be hardcoded or we'll have a lot of pain if we change it later (or make it OS-agnostic)

bittensor/utils/__init__.py
Eugene-hu: Oops, nice catch! Let me add that to init file

bittensor/_cli/cli_impl.py
Eugene-hu: Yea, I agree. Right now, bittensor installation is mostly done using the install script. As we move away, this hardcoded path is gonna bite us haha

PR #743, opened 2022-04-13
Blacklist update

Eugene-hu:
Improve the blacklist on both advanced and template server

### DIFF
--- a/bittensor/_neuron/text/template_server/run.py
+++ b/bittensor/_neuron/text/template_server/run.py
@@ -30,6 +30,7 @@ import pandas
 import datetime
 from threading import Lock
 from loguru import logger; logger = logger.opt(colors=True)
+from datetime import datetime,timedelta
 
 def serve( 
         config, 
@@ -68,6 +69,8 @@ def serve(
     )
     mutex = Lock()
 
+    timecheck = {}
+    n_topk_peer_weights = subtensor.min_allowed_weights
     def forward_text ( inputs_x ):
         r""" Single threaded version of the Forward function that is called when the axon recieves a forward request from other peers
         """ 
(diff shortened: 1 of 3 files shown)


shibshib (approved):

bittensor/_neuron/text/template_server/run.py
```diff
@@ -101,15 +104,61 @@ def blacklist(pubkey:str, request_type:bittensor.proto.RequestType) -> bool:
...
+            else:
+                timecheck[pubkey] = current_time
                 return False
+
+        # Black list or not
+        try:
+            registration_check()
+
+            stake_check()
+
```
isabella618033, line 154: hmmm no time check needed for template_server?

bittensor/_neuron/text/template_server/run.py
```diff
@@ -101,15 +104,61 @@ def blacklist(pubkey:str, request_type:bittensor.proto.RequestType) -> bool:
...
+            validator_check()
             
+            return False
+
+        except Exception as e:
```
isabella618033, line 159: the test would not bug you but ala might hahaha

isabella618033 (approved):
amazing job!

bittensor/_neuron/text/template_server/run.py
Eugene-hu: haha this is to catch exceptions from the tests, so tests can run in sequential. I think it should be fine?

bittensor/_neuron/text/template_server/run.py
Eugene-hu: Still debating if timecheck  is necessary, since multiprocessing would break the timechecks

isabella618033: But then the core validator is not multiprocessesd on sending dendrite calls rn.

bittensor/_neuron/text/template_server/run.py
shibshib: ugh. will allow.

PR #745, opened 2022-04-14
Added custom endpoints to metagraph cache network lookup

shibshib:
This allows people who are using a custom endpoint running their subtensor to also sync to a cached metagraph.

### DIFF
--- a/bittensor/_subtensor/__init__.py
+++ b/bittensor/_subtensor/__init__.py
@@ -125,7 +125,7 @@ class subtensor:
         # Select using config.subtensor.chain_endpoint
         elif config.subtensor.chain_endpoint != None:
             config.subtensor.chain_endpoint = config.subtensor.chain_endpoint
-            config.subtensor.network = network
+            config.subtensor.network = config.subtensor.network
          
         # Select using config.subtensor.network
         elif config.subtensor.network != None:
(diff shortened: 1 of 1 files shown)


isabella618033 (approved):

bittensor/_metagraph/metagraph_impl.py
```diff
@@ -429,7 +429,7 @@ def sync ( self, block: int = None, cached: bool = True ) -> 'Metagraph':
         logger.success(self.subtensor)
         if block == None:
             block = self.subtensor.get_current_block()
-            if cached and self.subtensor.network in ("nakamoto", "local"):
+            if cached and self.subtensor.network in ("nakamoto", "local", "endpoint"):
```
Eugene-hu: This would automatically assume Nakamoto, even though the endpoint might be connecting nobunaga or other networks. Do we want that?

Eugene-hu (approved):

bittensor/_metagraph/metagraph_impl.py
shibshib: not sure how you mean, why would it automatically assume Nakamoto? if you put in Nobunaga it will sync manually using Nobunaga.

PR #746, opened 2022-04-14
Check for replaced hotkey and reset the moving average score

isabella618033:
When syncing metagraph, check for a replaced hotkey and reset the moving average score.


Eugene-hu (approved):
LGTM

PR #750, opened 2022-04-21
Template miner hotfix

Eugene-hu:
Bring template miner in line with the core validator. This should also fix the issue with the increasing uids.

### DIFF
--- a/bittensor/_neuron/text/template_miner/__init__.py
+++ b/bittensor/_neuron/text/template_miner/__init__.py
@@ -413,8 +413,8 @@ class neuron:
             self.moving_avg_scores = torch.ones_like( self.metagraph.S ) * -1
 
         if self.metagraph.n > len(self.moving_avg_scores):
-            size_incerease = self.metagraph.n - len(self.moving_avg_scores)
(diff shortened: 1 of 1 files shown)


bittensor/_neuron/text/template_miner/__init__.py
```diff
@@ -397,6 +404,18 @@ def run_epoch( self ):
...
+    def metagraph_sync(self):
+        r""" Syncing metagraph together with other metagraph-size related objects
+        """
+        self.metagraph.sync()
+
+        if self.moving_avg_scores == None:
+            self.moving_avg_scores = torch.ones_like( self.metagraph.S ) * -1
+
+        if self.metagraph.n > len(self.moving_avg_scores):
+            size_incerease = self.metagraph.n - len(self.moving_avg_scores)
```
shibshib: nit: spelling. size_incerease should be size_increase,

shibshib (approved):

bittensor/_neuron/text/template_miner/__init__.py
Eugene-hu: Fixed!

PR #755, opened 2022-04-27
chain weight fix

Eugene-hu:
Uid issue caused by the metagraph not being updated properly. We should instead query the chain for the latest size of the network rather than depending on the metagraph.


bittensor/_neuron/text/advanced_server/run.py
```diff
@@ -252,11 +252,10 @@ def time_check():
...
 
         # --- creating our chain weights
-        chain_weights = torch.zeros(metagraph.n)
+        # --- query the chain for the most current number of peers on the network
+        chain_weights = torch.zeros(subtensor.n)
         uid = nn.uid
-       
-        if uid < len(chain_weights):
-            chain_weights[uid] = 1 
+        chain_weights[uid] = 1
```
shibshib, line 258: We need a better fix than this, because people's servers have been hanging because the UID doesn't exist in metagraph yet.

shibshib (approved):

bittensor/_neuron/text/advanced_server/run.py
Eugene-hu: We now query the chain for the number of peers on the network. If they are registered, their uid will be included as part of the set.

shibshib (approved):

PR #757, opened 2022-05-03
Catch CLI exceptions when no wallets are found

camfairchild:
Previously having no wallets meant `btcli list` and `btcli overview` would raise an uncaught exception. Now these should be handled with a useful message to the console.

image
image


shibshib:
Love the screenshots, well done!

tests/integration_tests/test_cli.py
```diff
@@ -106,6 +127,22 @@ def test_check_configs(self):
...
+        config.subtensor._mock = True
+        config.subtensor.network = "mock"
+        config.no_prompt = True
+
+        cli = bittensor.cli(config)
+        cli.run()
+
+    def test_overview_no_wallet( self ):
+        # Delete wallet if exists
+        self.rm_wallet()
```
joeylegere: Rather than calling `rm_wallet`, why not modify the config such that `config.wallet.path` points to an empty directory? When we run tests in parallel there's no guarantee that this will run at the same time as another test that relies on an extant wallet and mess with the results in a sporadic way.

tests/integration_tests/test_cli.py
```diff
@@ -84,6 +86,25 @@ def generate_wallet(coldkey : 'Keypair' = None, hotkey: 'Keypair' = None):
...
+        if os.path.exists('/tmp/pytest'):
+            shutil.rmtree('/tmp/pytest')
+    
+        the_wallet = bittensor.wallet (
+            path = '/tmp/pytest',
+            name = 'pytest',
+            hotkey = 'pytest',
+        )
+        
+        return the_wallet
```
joeylegere: This method already exists in `integration_tests/test_wallet.py`. Rather than redefining it here, why not move it into something like `tests/utils.py` and import into the tests that need it?

tests/integration_tests/test_cli.py
camfairchild: this is from some other code that does wallet stuff in the wallet tests. Instead I can just mock the IO, if you think that makes more sense.

bittensor/_cli/cli_impl.py
```diff
@@ -256,15 +256,23 @@ def set_weights( self ):
...
+            hotkey_files = next(os.walk(os.path.expanduser(hotkeys_path)))[2]
+        except StopIteration:
+            hotkey_files = []
         for hotkey_file_name in hotkey_files:
             hotkey_wallets.append( bittensor.wallet( path = wallet.path, name = wallet.name, hotkey = hotkey_file_name ))
         return hotkey_wallets
 
     def list(self):
         r""" Lists wallets.
         """
```
joeylegere, line 269: This comment isn't addressed at this PR, but this method is in dire need of some TLC and refactoring. We have 4 levels of indentation with static indexing into some lists, a bunch of bare try/except statements. I think a wallet named `"?"` could mean it's encrypted but it could also mean an unknown error was raised at some point and silently executed.

tests/integration_tests/test_cli.py
camfairchild: Removed in favour of mocking the IO

tests/integration_tests/test_cli.py
camfairchild: Mocked the IO https://github.com/opentensor/bittensor/pull/757/commits/adb54cf96387f546143966f658a3efee9d13f74b

camfairchild:
This PR also fixes an issue with coldkey regen where the default is `""` instead of `"default"` so the wallet hierarchy may get messed up

tests/integration_tests/test_cli.py
```diff
@@ -286,6 +325,73 @@ def test_inspect( self ):
...
+
+    def test_list_no_wallet( self ):
+        # Mock IO for wallet
+        with patch('bittensor.Wallet.coldkeypub_file', MagicMock(
+            exists_on_device=MagicMock(
+                return_value=False # Wallet doesn't exist
+            )
+        )):
+            config = self.config
+            config.wallet.path = '/tmp/pytest'
```
joeylegere: This might cause issues in parallel tests as this name is thrown around a lot. In a lazier implementation in a past PR, I just appended `time.time()` to the path. I've noticed this can still cause issues and collisions in the pathing.

Perhaps a better solution would be to give this a totally unique name (e.g. `/tmp/pytest_test_list_no_wallet`) if its completely encapsulated to this test.

joeylegere (commented):
This PR is from an external branch. nbd, i just updated the circleci settings to support this but I don't think it will trigger until the branch is updated again.

tests/integration_tests/test_cli.py
```diff
@@ -106,16 +110,51 @@ def test_check_configs(self):
...
+        # Mock IO for wallet
+        with patch('bittensor.Wallet.coldkeypub_file', MagicMock(
+            exists_on_device=MagicMock(
+                return_value=True # Wallet exists
+            )
+        )):
+            bittensor.subtensor.register = MagicMock(return_value = True)  
+            
+            config = self.config
+            config.wallet.path = '/tmp/walletpath'
```
joeylegere: see my other comment re: wallet path and parallel tests. Might not do anything right away, but could cause intermittent test failures down the road.

tests/integration_tests/test_cli.py
joeylegere: Could also add a function in `test_utils.py` that generates nonsense filepaths for us to use for testing (and keeps track of them akin to `get_random_unused_port`)

camfairchild: > Perhaps a better solution would be to give this a totally unique name (e.g. `/tmp/pytest_test_list_no_wallet`) if its completely encapsulated to this test.

Went with this solution as the IO should be mocked anyway.

joeylegere (approved):

PR #759, opened 2022-05-03
Constrain on the dataset to pull

isabella618033:
- Constrain on the dataset to pull from ipfs to be **Books3** corpus only, so it would be faster for the github action to pass.
- **BUT it would not solve the underlying issue** where by default when user trying to pull a mix of datasets from ipfs, it would still take them super long.
- Why would other dataset take longer? Because the **mountain hash** (the txt file that stores the hashes of all other txt files that belongs to that dataset) of other datasets are larger. Also, some **data file** of other dataset is so small that we have to send a lot more requests to IPFS to collect the same amount of text.


joeylegere:
Overall this LGTM, but I'd like to see a variable like `mock_dataset_for_tests=["Books3"]` defined in a `constant.py` or `config.yaml` or whatever so we can just import it. Then if we decide to change the dataset we use we only need to change one line of code rather than 6.

isabella618033:
> Overall this LGTM, but I'd like to see a variable like `mock_dataset_for_tests=["Books3"]` defined in a `constant.py` or `config.yaml` or whatever so we can just import it. Then if we decide to change the dataset we use we only need to change one line of code rather than 6.


Hey, nice suggestion! I have made an update to add `constant.py`, but rn it is rather empty with only constants for the dataset test. Lets move other constant over in some other PR if it is good

joeylegere (approved):

PR #761, opened 2022-05-03
Check transfer destination in API

camfairchild:
Previously, if the destination address was invalid, it was only caught by the config check in the cli.
Now, the check is done in the API as well.

image

Further, the previous implementation of the validation didn't support a public key passed as bytes to the API, whereas this new implementation does support bytes.


bittensor/utils/__init__.py
```diff
@@ -226,3 +228,84 @@ def create_pow( subtensor, wallet ):
...
+    Checks if the given address is a valid ss58 address.
+
+    Args:
+        address(str): The address to check.
+
+    Returns:
+        True if the address is a valid ss58 address for Bittensor, False otherwise.
+    """
+    try:
+        return ss58.is_valid_ss58_address( address, valid_ss58_format=42 )
```
joeylegere: I generally try to avoid 'magic numbers' like this where possible. Not that it's likely to change or anything (though it would be a pain to update every instance if this were a more dynamic variable), but it also helps with legibility. 

This would be another thing that would be nice in a constants file so that it has a semantic meaning associated with it when reading the code.

camfairchild: I moved the constant to the bittensor `__init__.py` file as `__ss58_format__`

joeylegere (approved):

PR #762, opened 2022-05-05
Parallelize unit tests

joeylegere:
Makes some changes that allow unit tests to run in parallel. This is imperfect and has some hacks, but the main changes are:

* Choosing random unused ports when instantiating instances of axons, subtensors, etc.
* Using unittest's `setUp` method to ensure objects that need to exist for tests are created within those tests.
* Basing any paths that are generated on the current timestamp (i.e. any wallet).
* Fun one: some of the tests need to spawn/kill mock subtensor subprocesses. These subprocesses are now identified by their parent's PID and the port that it uses is chosen using this as well. pretty cool, right?

One side effect of this PR is that I run `isort` on save in my IDE. This rearranges some of the imports and removes unused ones so any import code was likely just resorted.

### DIFF
--- a/bittensor/_subtensor/subtensor_mock.py
+++ b/bittensor/_subtensor/subtensor_mock.py
@@ -41,7 +41,7 @@ __type_registery__ = {
     }
 }
 
-GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME = f"node-subtensor-{os.getpid()}"
+GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME = "node-subtensor"
 print(GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME)
 
 
@@ -86,7 +86,7 @@ class mock_subtensor():
         r""" If subtensor is running a mock process this kills the mock.
         """
         for p in psutil.process_iter():
-            if p.name() == GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME and p.status() != psutil.STATUS_ZOMBIE and p.status() != psutil.STATUS_DEAD:
+            if p.name() == GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME and p.parent().pid == os.getpid() and p.status() != psutil.STATUS_ZOMBIE and p.status() != psutil.STATUS_DEAD:
                return True
         return False
 
@@ -95,7 +95,7 @@ class mock_subtensor():
         r""" Kills the global mocked subtensor process even if not owned.
         """
         for p in psutil.process_iter():
-            if p.name() == GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME:
+            if p.name() == GLOBAL_SUBTENSOR_MOCK_PROCESS_NAME and p.parent().pid == os.getpid() :
                 p.terminate()
                 p.kill()
 
--- a/bittensor/_subtensor/subtensor_mock.py
+++ b/bittensor/_subtensor/subtensor_mock.py
@@ -7,6 +7,7 @@ import time
 import os
 
 from . import subtensor_impl
+from tests.utils import get_random_unused_port
 
 __type_registery__ = {
     "runtime_id": 2,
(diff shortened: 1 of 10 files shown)


shibshib:
@joeylegere we do descriptions in our PRs, we're not heathens.

tests/integration_tests/test_keyfile.py
```diff
@@ -9,126 +9,151 @@
...
+        keyfile = bittensor.keyfile(path=os.path.join(self.root_path, "keyfile"))
+
+        mnemonic = bittensor.Keypair.generate_mnemonic(12)
+        alice = bittensor.Keypair.create_from_mnemonic(mnemonic)
+        keyfile.set_keypair(alice, encrypt=True, overwrite=True, password='thisisafakepassword')
+
+        bob = bittensor.Keypair.create_from_uri('/Bob')
+        keyfile.set_keypair(bob, encrypt=True, overwrite=True, password='thisisafakepassword')
+
+        return keyfile
```
joeylegere, line 49: This highlighted section is where most of the main changes are. The rest is a result of having converted to an OO style to facilitate using setUp/tearDown

bittensor/_subtensor/subtensor_mock.py
```diff
@@ -104,10 +106,13 @@ def create_global_mock_process(self):
         try:
             operating_system = "OSX" if platform == "darwin" else "Linux"
             path = "./bin/chain/{}/node-subtensor".format(operating_system)
-            port = int(bittensor.__mock_entrypoints__[0].split(':')[1])
-            print(port)
+            ws_port = int(bittensor.__mock_entrypoints__[0].split(':')[1])
+            print(ws_port)
+            print(os.getpid())
+            baseport = get_random_unused_port()
+            rpc = get_random_unused_port()
```
joeylegere, line 113: Turns out we only care about `ws_port`. Using `port+1` and `port+2` for the other two ports caused issues because the port in `mock_entrypoint` was generated by the PID. But often these processes are spawned sequentially, so the PIDs are as well, which means the subprocesses all squabble for the same port and everyone ends up losing.

tests/unit_tests/bittensor_tests/test_config.py
```diff
@@ -26,7 +26,12 @@ def test_loaded_config():
 
 def test_strict():
     parser = argparse.ArgumentParser()
-    parser.add_argument("arg", help="Dummy Args")
+
+    # Positional/mandatory arguments don't play nice with multiprocessing.
+    # When the CLI is used, the argument is just the 0th element or the filepath.
+    # However with multiprocessing this function call actually comes from a subprocess, and so there
+    # is no positional argument and this raises an exception when we try to parse the args later.
+    # parser.add_argument("arg", help="Dummy Args")
```
joeylegere, line 34: I welcome a better solution here but I just commented this out. rationale above.

tests/utils.py
```diff
@@ -0,0 +1,22 @@
+import socket
+from random import randint
+from typing import Set
+
+max_tries = 10
+
+
+def get_random_unused_port(allocated_ports: Set = set()):
```
joeylegere, line 8: This default argument is mutable by design and is necessary for the function to exist correctly. Honestly it seems like I'm trolling but I can't think of a better way to maintain this function's state between function calls (possibly because abusing this counterintuitive behaviour fills me with glee).

shibshib (approved):
Solid PR! Do we ignore the current tests failing?

joeylegere:
> Solid PR! Do we ignore the current tests failing?

Yeah, they're (surprisingly) unrelated to the PR. It was green [a few days ago](https://github.com/opentensor/bittensor/runs/6299680380?check_suite_focus=true), but the subtensor tests are failing because nobunaga is down. I've planned another PR so our unit tests don't rely on the test network in the future, but that's going to be separate and probably after we get the tests running quickly.

tests/utils.py
shibshib: an evil more terrible way to do this would be to keep an unencrypted text file that describes the state.

bittensor/_subtensor/subtensor_mock.py
shibshib: Nice catch!

PR #768, opened 2022-05-12
CircleCI

joeylegere:
Switches our CI from github actions to CircleCI so we can scale up as necessary. Build times are consistently <5 minutes as opposed to >15.

The main change is in the .circleci/config.yml, with other minor changes to facilitate multiprocessing, and the config_tests folder moved to a subfolder of unit_tests to avoid messy globbing when generating file paths.


tests/integration_tests/test_server_compression.py
```diff
@@ -80,10 +80,12 @@ def get_socket(bind_address='localhost',
...
             sock.close()
-            if os_error.errno in _UNRECOVERABLE_ERRNOS:
-                raise
-            else:
-                continue
+
+            # This variable was undefined--what were the previously unrecoverable error numbers?
+            # if os_error.errno in _UNRECOVERABLE_ERRNOS:
+            #    raise
+            # else:
```
shibshib, line 87: I think just take this out entirely.

shibshib (approved):

shibshib:
Looks good but what about coverage?

PR #769, opened 2022-05-13
Bit 440 advanced server finetuning

isabella618033:
Adding option to finetune only on the last few layer(s) which would be the proper way of finetuning. Also it will decrease a lot of memory usage. eg. it took 15+GB memory to train on gpt2-medium (345M parameters) for block_size 3 sequence length 512 tokens, now it would only take ~2GB for setting it to train on the last layer.


bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,54 @@ def __init__(self,
...
         # -- keeps track of gradients applied
         self.backward_gradients = 0 
-        
+        self.set_fine_tuning_params()
+
+    def set_fine_tuning_params(self):
+        r''' Set to tune only the parameter of the last layer
+        '''
+        def find_last_layer(model):    
+            r''' Recursively find the last layer in a nn.ModuleList
```
Eugene-hu, line 104: Would this work for models that have an additional head? Ex Causallm

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,54 @@ def __init__(self,
...
+            logger.success(f'Set to finetune layer {last_layer_name} and onwards')
+            for name, param in self.pre_model.named_parameters():
+                if last_layer_name in name or reached_last_layer == True:
+                    param.requires_grad = True
+                    reached_last_layer = True
+                else:
+                    param.requires_grad = False
+
+        if reached_last_layer == False:
+            if self.config.neuron.finetune.all:
```
Eugene-hu, line 151: Is this parameter being used? Do we assume that for all params requires_grad = True by default?

Eugene-hu (approved):
Looks fairly good to me! left you a few questions

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: yea, cause it look for the last layer of transformer, and if the last layer is a transformer with multiple heads, all of the heads would required grad.

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: When we pull the model from hugging face, all of the parameters requires grad, so that would be the default from hugging face. But then our default would be to set requires grad for the last layer only.

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,54 @@ def __init__(self,
         
         # -- keeps track of gradients applied
         self.backward_gradients = 0 
-        
+        self.set_fine_tuning_params()
+
+    def set_fine_tuning_params(self):
+        r''' Set to tune only the parameter of the last layer
+        '''
```
joeylegere: Can you add a test for this?

isabella618033: hey i have added some test to the file tests/unit_tests/bittensor_tests/test_neuron.py, please check:)

tests/unit_tests/bittensor_tests/test_neuron.py
```diff
@@ -0,0 +1,60 @@
+import bittensor
+import torch 
+import torch.nn as nn
+from torch.nn import TransformerEncoder, TransformerEncoderLayer
+
+adv_server = bittensor._neuron.text.advanced_server.server()
```
joeylegere: Is there a reason this is here, or can you instantiate this within the test function?

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,56 @@ def __init__(self,
...
+
+        # set the non-last layer parameters not to require grads
+        if (not self.config.neuron.finetune.all) and (last_layer_name != None):
+            logger.success(f'Set to finetune layer {last_layer_name} and onwards')
+            for name, param in self.pre_model.named_parameters():
+                if last_layer_name in name or reached_last_layer == True:
+                    param.requires_grad = True
+                    reached_last_layer = True
+                else:
+                    param.requires_grad = False
```
joeylegere: I find this section a bit confusing to follow, and I expect it will be challenging to refactor or debug later. Could we try to simplify the logic a bit?

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,56 @@ def __init__(self,
...
+
+        reached_last_layer = False
+
+        # set the non-last layer parameters not to require grads
+        if (not self.config.neuron.finetune.all) and (last_layer_name != None):
+            logger.success(f'Set to finetune layer {last_layer_name} and onwards')
+            for name, param in self.pre_model.named_parameters():
+                if last_layer_name in name or reached_last_layer == True:
+                    param.requires_grad = True
+                    reached_last_layer = True
```
joeylegere: don't need this; we can't enter this block if `reached_last_layer` isn't already `True`

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,56 @@ def __init__(self,
         
         # -- keeps track of gradients applied
         self.backward_gradients = 0 
-        
+        self.set_fine_tuning_params()
+
+    def set_fine_tuning_params(self):
```
joeylegere: Could we add return type annotations here?

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,56 @@ def __init__(self,
...
+        '''
+        def find_last_layer(model):    
+            r''' Recursively find the last layer in a nn.ModuleList
+            '''
+            reverted_child_list = [(name, child) for name, child in model.named_children()]
+            reverted_child_list.reverse()
+
+            for name, child in reverted_child_list:    
+                if isinstance(child, nn.ModuleList):
+                    if self.config.neuron.finetune.num_layers > len(child):
```
joeylegere, line 118: these could be inlined to avoid too much indentation. In python if you have `if x and y`, if `x` is falsey, then the interpreter actually won't even bother to check for condition `y`. Which is good in this case because it looks like you split it among two lines to avoid an error calling `len` on something that's not a `ModuleList`

bittensor/_neuron/text/advanced_server/nucleus_impl.py
```diff
@@ -87,7 +87,56 @@ def __init__(self,
         
         # -- keeps track of gradients applied
         self.backward_gradients = 0 
-        
+        self.set_fine_tuning_params()
+
+    def set_fine_tuning_params(self):
+        r''' Set to tune only the parameter of the last layer
+        '''
+        def find_last_layer(model):
```
joeylegere: Could we also have type hinting here?

tests/unit_tests/bittensor_tests/test_neuron.py
isabella618033: yup, it should be in the test function.

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: not really, cause it is an 'or' within the if case 
`if last_layer_name in name or reached_last_layer == True:`

basically the idea is for example, if there are layers like a-> b-> c-> d ->e, and layer c's name coincidence with the specified last_layer_name, then all c,d,e should required grad.

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: but the 2 if returns different things
now it is like `if a then (if b then return c) return d`, where both a and b was checked once only.
if we change it to inline then we might have to do it like `if (a and b) then return c elif a then return d` where we might have to check a for 2 times. 
or maybe there is some other better way

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: sure!

bittensor/_neuron/text/advanced_server/nucleus_impl.py
isabella618033: sure! my bad haha

bittensor/_neuron/text/advanced_server/nucleus_impl.py
joeylegere: Oh, I completely misread that it's an `or` not `and`, my bad.

joeylegere (approved):

PR #770, opened 2022-05-17
[BIT-398] Tokenizer utilities and unit tests

opentaco:
#### Add tokenizer utilities and unit tests

1. Adds tokenizer utilities, including tokenizer compatibility checking, cross-tokenizer pre-processing, and literal logit translation.
2. Adds unit tests for tokenizer compatibility checking and literal logit translation.
3. Server model logits are saved to pytorch compressed format (.pt) to bypass model calls for faster routine unit testing.

[BIT-398]: https://opentensor.atlassian.net/browse/BIT-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit_tests/bittensor_tests/utils/test_tokenizer_utils.py
```diff
@@ -0,0 +1,290 @@
...
+        print('Will first run server models (requires CUDA)...')
+
+        # === Run server models to obtain encoded logits ===
+        encodings = {}
+        for text_name, model_name, max_length in test_pairs:
+            result = tokenizer_translation(sample_text[text_name], model_name, max_length, topk=128)
+            original_loss, encoded_loss, translated_loss, enc_pre_logits = result
+            encodings[(text_name, model_name)] = (encoded_loss, translated_loss, enc_pre_logits)
+
+            print(text_name, model_name, original_loss, encoded_loss, translated_loss)
```
joeylegere: I don't think we should be handling this exception here; the test should just fail if we don't have cached embeddings, especially since they're in VCS.

When we run in circleCI we don't have the resources to compute these, and it leaves the test just hanging silently as it computes the embeddings. I think we should have this code in a separate script if we ever want to change what we're doing in this test

bittensor/utils/tokenizer_utils.py
```diff
@@ -0,0 +1,910 @@
...
+            probs_std (:obj:`torch.FloatTensor`, `required`):
+                [std_sequence_len, std_vocab_size] Output probability distribution over a target tokenizer vocabulary.
+                Reference that will be written in-place.
+            offset_mapping (:obj:`List[tuple]`, `required`):
+                Tokenizer offset mappings for a specific sequence [(left_0, right_0), (left_1, right_1), ...].
+            offset_mapping_std (:obj:`List[tuple]`, `required`):
+                Standard tokenizer offset mappings for a specific sequence [(left_0, right_0), (left_1, right_1), ...]
+            tokenizer (:obj:`PreTrainedTokenizerBase`, `required`):
+                Source tokenizer.
+            std_tokenizer (:obj:`PreTrainedTokenizerBase`, `required`):
```
joeylegere, line 534: Doesn't look like this actually is used in this function

bittensor/utils/tokenizer_utils.py
```diff
@@ -0,0 +1,910 @@
...
+                pairings[getattr(from_tokenizer, special_token)] = getattr(to_tokenizer, special_token)
+
+    return pairings
+
+
+def translate_special_token_text(text_batch: List[str], from_tokenizer: PreTrainedTokenizerBase,
+                                 to_tokenizer: PreTrainedTokenizerBase) -> Tuple[List[str],
+                                                                                 List[List[List[int]]],
+                                                                                 List[List[List[int]]],
+                                                                                 List[List[List[Any]]]]:
```
joeylegere, line 857: This return type is pretty wild. I think you should create a data structure to encapsulate it in a way that's easier to understand.

bittensor/utils/tokenizer_utils.py
opentaco: Need these parameters for debugging later on, so keeping it for now.

bittensor/utils/tokenizer_utils.py
opentaco: Noted, may simplify this later on.

PR #771, opened 2022-05-19
Add Cached Overview

camfairchild:
This PR adds a cached overview (from IPFS) by default, with the flag `--no_cache` to opt-out
- This significantly speeds up the overview command 
     - Assuming IPFS is fairly up-to-date, the accuracy issues for most people are likely less significant than the time overview would take

Edit:
This PR also adds:
- Hotkey sorting in overview via fuzzy-matching a column title `--wallet.sort_by <column_title>`
- Hotkey sort_order in overview via specifying `--wallet.sort_order <ascending|descending|reverse>`
- Hotkey specifying for overview `--wallet.hotkeys <hk0> <hk1> <hk2> ...`
- Optional width argument for the overview output `--width <int>`

### DIFF
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -21,6 +21,7 @@ from types import SimpleNamespace
 
 import bittensor
 from bittensor.utils.balance import Balance
+from fuzzywuzzy import fuzz
 from rich import print
 from rich.prompt import Confirm
 from rich.table import Table
@@ -511,6 +512,17 @@ class CLI:
         table.caption = "[white]Wallet balance: [green]\u03C4" + str(balance.tao)
 
         console.clear()
+
+        sort_by: str = self.config.wallet.sort_by
+        if sort_by is not None:
+            column_to_sort_by: int = 0
+            for index, column in zip(range(len(table.columns)), table.columns):
+                # Fuzzy match the column name. Default to the first column.
+                if fuzz.ratio(sort_by.lower(), column.header.lower().replace('[overline white]', '')) > 80:
+                    column_to_sort_by = index
+                    break
+            TABLE_DATA.sort(key=lambda row: row[column_to_sort_by])
+
         for row in TABLE_DATA:
             table.add_row(*row)
         table.box = None
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -514,7 +514,7 @@ class CLI:
         console.clear()
 
         sort_by: str = self.config.wallet.sort_by
-        if sort_by is not None:
+        if sort_by != "":
             column_to_sort_by: int = 0
             for index, column in zip(range(len(table.columns)), table.columns):
                 # Fuzzy match the column name. Default to the first column.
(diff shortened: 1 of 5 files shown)


bittensor/_cli/cli_impl.py
```diff
@@ -413,10 +413,31 @@ def overview(self):
...
-                nn = subtensor.neuron_for_pubkey( wallet.hotkey.ss58_address )
-                if not nn.is_null:
-                    neurons.append( (nn, wallet) )
+            try:
+                if self.config.subtensor.network not in ('local', 'nakamoto'):
+                    # We only cache neurons for local/nakamoto.
+                    raise Exception("This network is not cached, defaulting to regular overview.")
+            
+                if self.config.no_cache:
+                    raise Exception("Flag was set to not use cache, defaulting to regular overview.")
```
isabella618033: Wouldnt the program stop because the config is no_cache when you raise exception here? xd

camfairchild: Haha, it's in a try block already.

bittensor/_cli/cli_impl.py
```diff
@@ -410,13 +411,39 @@ def overview(self):
...
+                # Grab cached neurons from IPFS
+                all_neurons = metagraph.retrieve_cached_neurons()
+                # Map the hotkeys to uids
+                hotkey_to_neurons = {n.hotkey: n.uid for n in all_neurons}
+                for wallet in tqdm(all_hotkeys):
+                    uid = hotkey_to_neurons.get(wallet.hotkey.ss58_address)
+                    if uid is not None:
+                        nn = all_neurons[uid]
+                        neurons.append( (nn, wallet) )
+            except Exception as e:
```
joeylegere: what kind of exceptions can this block throw? we should try to avoid bare catch statements and make it more explicit why we're falling back to the code below

camfairchild: I modified it to catch only around the call to retrieve neurons. This is the way it's done in the retrieve function as well, so I'm not sure which exceptions it will actually throw, and it will likely catch them already within the function

joeylegere: Okay, thanks for the clarification. and I see they're already logged there, so no need to repeat the message here.

bittensor/_cli/cli_impl.py
```diff
@@ -484,9 +515,34 @@ def overview(self):
...
+            column_to_sort_by: int = 0
+            sort_descending: bool = False # Default sort_order to ascending
+
+            for index, column in zip(range(len(table.columns)), table.columns):
+                # Fuzzy match the column name. Default to the first column.
+                if fuzz.ratio(sort_by.lower(), column.header.lower().replace('[overline white]', '')) > 80:
+                    column_to_sort_by = index
+                    break
+            
+            if sort_order.lower() == 'desc' or sort_order.lower() == 'descending' or sort_order.lower() == 'reverse':
```
joeylegere: nit: would be easier to read if you're checking set membership i.e. `if sort_order.lower() in ('desc', .....)`

joeylegere (commented):
Overall LGTM, just need to resolve merge conflicts and it's good to go

bittensor/_cli/cli_impl.py
camfairchild: done

joeylegere (approved):

camfairchild:
Going to close and re-open this to try to get CI to work

PR #772, opened 2022-05-24
Fix Docker Build and Docker-Compose

rmaronn:
This PR:

### **- Irons out the Python version inconsistency between the install scripts and the Dockerfile**

Right now, the Docker image builds with Python3.7, but the `install.sh` script per the install instructions installs Python3.8. This is because Python3 is installed in the install script, but we specifically point out 3.7 in the Dockerfile. This causes the docker container to not be compatible with the new keccak update, breaking it.

### **- Fixes the Docker Compose spec to have the command correctly route and find the Bittensor Python library**

Previously, we were not specifying the path to the Bittensor Python library. By specifying the `PYTHONPATH` environment variable, we are able to tell it where to go to find it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,4 +10,4 @@ services:
       - ~/.bittensor:/root/.bittensor
 
     command: /bin/bash -c "
-      PYTHONPATH=/bittensor python3 /bittensor/miners/template_miner.py --subtensor.network akatsuki"
+      PYTHONPATH=/bittensor python3 /bittensor/miners/template_miner.py --subtensor.network nobunaga"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -8,7 +8,6 @@ services:
       - "8091:8091"
     volumes:
       - ~/.bittensor:/root/.bittensor
-    
+
     command: /bin/bash -c "
-      cd /bittensor
-      && python3.7 miners/text/template_miner.py --subtensor.network akatsuki"
+      PYTHONPATH=/bittensor python3 /bittensor/miners/template_miner.py --subtensor.network akatsuki"
```
shibshib: @rmaronn this should be Nobunaga, akatsuki is deprecated

shibshib (changes requested):

shibshib (approved):

PR #778, opened 2022-06-01
Synapse fix

isabella618033:
- saving previous synapse request and reuse it for the same input 
- debugged so the time it takes to complete 2 synapse come down from 8-12 sec to 2-3 sec


bittensor/_neuron/text/template_server/nucleus_impl.py
```diff
@@ -131,34 +132,39 @@ def encode_forward(self,inputs,tokenizer=None):
...
-        elif self.config.neuron.autocast and self.device == 'cuda':
-            pre_hidden = self.pre_model(inputs).last_hidden_state
+        if self.config.neuron.training or self.config.neuron.autocast and self.device == 'cuda':
+            output = self.pre_model(inputs, output_hidden_states=True)
+            pre_hidden = output.hidden_states[-1]
         else:
             with torch.no_grad():
-                pre_hidden = self.pre_model(inputs).last_hidden_state
-        
+                output = self.pre_model(inputs, output_hidden_states=True)
```
Eugene-hu, line 146: Would this always compute the logits layer as well? If so, do we know how much time the last layer takes?

bittensor/_neuron/text/template_server/nucleus_impl.py
```diff
@@ -131,34 +132,39 @@ def encode_forward(self,inputs,tokenizer=None):
...
+                pre_hidden = output.hidden_states[-1]
+
         if self.interpolate and sen_len[1] != pre_hidden.size()[1]:
             down= F.interpolate(pre_hidden.unsqueeze(1),size=[sen_len[1],pre_hidden.size()[2]],mode=self.inter_degree).squeeze(1)
```
Eugene-hu, line 150: We can probably remove the padding, and interpolation here and move them to encode and decode functions in the synapses

bittensor/_neuron/text/template_server/run.py
```diff
@@ -84,17 +84,22 @@ def forward_generate( inputs_x:torch.FloatTensor, synapse):
...
-
-    def forward_casual_lm(inputs_x, synapse):
-        output = model.pre_model(input_ids=inputs_x).logits
-        return output
+    def forward_hidden_state(inputs_x, synapse, model_output = None):
+        if model_output == None:
+            model_output = model.encode_forward(inputs_x.to(model.device))
+        
+        hidden = model_output.hidden
+        padding_r = (1024-hidden.size(2))
```
Eugene-hu: If we use the encode forward function, we can remove these padding functions.

Eugene-hu (approved):
Looks good to me! Left you a few comments

bittensor/_neuron/text/template_server/nucleus_impl.py
isabella618033: pre_model() always compute the whole model, always including the last layer, setting option output_hidden_states = True just add a return variable.

bittensor/_neuron/text/template_server/run.py
isabella618033: okie

PR #779, opened 2022-06-01
Hotfix: cached overview numeric sorting

camfairchild:
This PR fixes sorting numeric columns as numeric rather than strings

### DIFF
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -548,15 +548,26 @@ class CLI:
 
             for index, column in zip(range(len(table.columns)), table.columns):
                 # Fuzzy match the column name. Default to the first column.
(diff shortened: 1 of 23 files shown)


bittensor/_cli/cli_impl.py
```diff
@@ -548,15 +548,26 @@ def overview(self):
 
             for index, column in zip(range(len(table.columns)), table.columns):
                 # Fuzzy match the column name. Default to the first column.
-                if fuzz.ratio(sort_by.lower(), column.header.lower().replace('[overline white]', '')) > 80:
+                column_name = column.header.lower().replace('[overline white]', '')
+                match_ratio = fuzz.ratio(sort_by.lower(), column_name)
+                if  match_ratio > 70:
```
joeylegere: can you add to constants.py?

camfairchild: changed to use the maximum match rather than a threshold

joeylegere (approved):

PR #783, opened 2022-06-06
Only support modern python versions

joeylegere:
* Bumps various versions to support Python3.10 (and consequently Ubuntu 22.04, as that's the system-default).
* Removes support for EOL Python versions (we now only officially recognize >= 3.7).
* CircleCI now runs tests against all supported python versions in parallel.
    * We now try rerunning all tests twice, because 4/4 independent runs all need to pass and apparently saCRIfICiAL GOats aRe NoT A VAlID BUsineSS eXpeNSE


shibshib:
Sacrificial goats may only be purchased using official provider in Baghdad.

shibshib (approved):

PR #792, opened 2022-06-17
Bit 458 combine advanced with template server

isabella618033:
Created core server which does local training or remote training optionally 

1. parameter `--neuron.local_train` to control local training or not (default not to do local_train)
2. parameter `--neuron.remote_train` to control remote training or not (default not to do remote_train)
3. local_train does loss aggregation, so we averages the loss before doing backward to the loss(local loss) (line 285 in run.py), 
4. remote_train does gradient aggregation, so  when we have done n times backward, its gradient should be n times bigger then normal scale. Thus, before the optimisation, we divide the learning rate by n. (line 290 in run.py)
5. there is a response shape exception fix (line 324 nucleus.py) where we make sure that the resulted sequence length would be at least the input sequence length, otherwise padded by 0


bittensor/_neuron/__init__.py
```diff
@@ -20,18 +20,18 @@
...
-from .text import template_miner, template_server, advanced_server, core_validator, multitron_server
-
-
+from .text import template_miner, template_server, advanced_server, core_validator, multitron_server, core_server
 __all_neurons__ =  { 'text_template_miner': template_miner.neuron, 
                      'text_core_validator': core_validator.neuron,
                      'text_template_server':template_server.neuron,
                      'text_advanced_server':advanced_server.neuron,
-                     'multitron_server': multitron_server.neuron}
+                     'multitron_server': multitron_server.neuron,
```
Eugene-hu: Does the mulitron server still work?

bittensor/_neuron/text/core_server/run.py
```diff
@@ -0,0 +1,347 @@
...
+            #validator_check()
+            
+            return False
+
+        except Exception as e:
+            return True
+
+    def backward_callback(inputs_x:torch.FloatTensor, grads_dy:torch.FloatTensor, synapses=[] ):
+        """
+            The default forward callback when no callback is attached: Is used to call specific synapse functions
```
Eugene-hu: Oops, please help me change these doc strings

bittensor/_neuron/text/core_server/run.py
```diff
@@ -0,0 +1,347 @@
...
+                        response_messages.append('Success')
+                    else:
+                        response_tensors.append(None)
+                        response_codes.append(bittensor.proto.ReturnCode.NotImplemented)
+                        response_messages.append('Not Implemented')
+                except Exception as e:
+                    # --- Exception Hit in Synapse ---
+                    response_tensors.append(None)
+                    response_codes.append(bittensor.proto.ReturnCode.UnknownException)
+                    response_messages.append(str(e)+'kill yourself')
```
Eugene-hu: Uhh, not sure if this is what we want the messages to look like

bittensor/_neuron/text/core_server/run.py
```diff
@@ -0,0 +1,347 @@
...
+            for index, synapse in enumerate(synapses):
+                try:
+                    if synapse.synapse_type in axon.synapse_callbacks and axon.synapse_callbacks[synapse.synapse_type] != None:
+                        model_output, response_tensor = axon.synapse_callbacks[synapse.synapse_type](inputs_x[index], synapse)
+                        grads_dy_norm = grads_dy[index]/(grads_dy[index].sum() + 0.00001)
+                        torch.autograd.backward (
+                            tensors = [ response_tensor ],
+                            grad_tensors = [ grads_dy_norm ],
+                            retain_graph=True
+                        )
```
Eugene-hu, line 211: Similar to the advanced server, we apply the gradients and do one large step, correct?

Eugene-hu (approved):
LGTM! left you a few comments

bittensor/_neuron/text/core_server/run.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+    def blacklist(pubkey:str, request_type:bittensor.proto.RequestType) -> bool:
+        r"""Axon security blacklisting, used to blacklist message from low stake members
+            Args:
+                pubkey ( str, `required`):
+                    The public key of the caller.
+                request_type ( bittensor.proto.RequestType, `required`):
+                    the request type ('FORWARD' or 'BACKWARD').
+        """
```
joeylegere: What is this method doing here? It looks like all the local functions aren't actually used and it will just always return False.

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION 
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
+# DEALINGS IN THE SOFTWARE.
+
+""" Template server.
+
+Example:
+    $ import neurons
+    $ neurons.text.template_server.neuron().run()
+"""
```
joeylegere: could you update the docstrings? there are a few places you could probably just find/replace template_server with core_server

bittensor/_neuron/text/core_server/run.py
```diff
@@ -0,0 +1,347 @@
...
+            root_dir = config.neuron.full_path
+        )
+
+    last_set_block = subtensor.get_current_block()
+
+
+    # --- Run Forever.
+    while True:
+        
+        interation = 0
```
joeylegere: typo? iteration?

bittensor/_neuron/__init__.py
isabella618033: it should not work, and it is also not hard to make it work, 
but also it is not a priority to renew it and validate its performance, 
so lets depricate it for now

bittensor/_neuron/text/core_server/run.py
isabella618033: hahaha nice catch

bittensor/_neuron/text/core_server/run.py
isabella618033: previously we had some problem where nobunaga had no tao in it 
so the stake_check and validator_check has always fail, 
so we just commented out all the checks for testing purpose 

but we are now ready to  bring back the blacklist function, so yea, we can just uncomment it now

bittensor/_neuron/text/core_server/run.py
isabella618033: yea, that's right

PR #796, opened 2022-06-23
UI updates

Eugene-hu:
- Add several UI updates to the CLI for the synapse update.
- Adds an additional synapse check that can be turned on to protect against potential synapse attacks
- Contain some updates from master
- Small fix for the core_server

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -271,7 +271,7 @@ class axon:
         if not inspect.ismethod(forward_callback) and not inspect.isfunction(forward_callback):
             raise ValueError('The axon forward callback must be a function with signature Callable[inputs_x: torch.Tensor] -> torch.FloatTensor:, got {}'.format(forward_callback))   
         if len( inspect.signature(forward_callback).parameters) != 3:
-            raise ValueError('The axon forward callback must have signature Callable[ inputs_x: torch.Tensor, synapses] -> torch.FloatTensor:, got {}'.format(inspect.signature(forward_callback)))
+            raise ValueError('The axon forward callback must have signature Callable[ inputs_x: torch.Tensor, synapses, hotkey] -> torch.FloatTensor:, got {}'.format(inspect.signature(forward_callback)))
         if 'inputs_x' not in inspect.signature(forward_callback).parameters:
             raise ValueError('The axon forward callback must have signature Callable[ inputs_x: torch.Tensor] -> torch.FloatTensor:, got {}'.format(inspect.signature(forward_callback)))
         
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -54,6 +54,7 @@ class axon:
             synapse_last_hidden: 'Callable' = None,
             synapse_causal_lm: 'Callable' = None,
             synapse_seq_2_seq: 'Callable' = None,
+            synapse_checks: 'Callable' = None,
             thread_pool: 'futures.ThreadPoolExecutor' = None,
             server: 'grpc._Server' = None,
             port: int = None,
@@ -81,7 +82,9 @@ class axon:
                 synapse_causal_lm (:obj:`callable`, `optional`):
                     function which is called by the causal lm synapse
                 synapse_seq_2_seq (:obj:`callable`, `optional`):
-                    function which is called by the seq2seq synapse           
+                    function which is called by the seq2seq synapse   
+                synapse_checks (:obj:`callable`, 'optional'):
+                    function which is called before each synapse to check for stake        
                 thread_pool (:obj:`ThreadPoolExecutor`, `optional`):
                     Threadpool used for processing server queries.
                 server (:obj:`grpc._Server`, `required`):
@@ -141,6 +144,8 @@ class axon:
         synapses[bittensor.proto.Synapse.SynapseType.TEXT_CAUSAL_LM] = synapse_causal_lm
         synapses[bittensor.proto.Synapse.SynapseType.TEXT_SEQ_2_SEQ] = synapse_seq_2_seq
         
+        synapse_check_function = synapse_checks if synapse_checks != None else axon.default_synapse_check
+
         if priority != None:
             priority_threadpool = bittensor.prioritythreadpool(config=config)
         else: 
@@ -154,6 +159,7 @@ class axon:
             forward = forward_text,
             backward = backward_text,
             synapses = synapses,
+            synapse_checks = synapse_check_function,
             priority = priority,
             priority_threadpool = priority_threadpool,
             forward_timeout = config.axon.forward_timeout,
@@ -235,6 +241,12 @@ class axon:
         assert config.axon.port > 1024 and config.axon.port < 65535, 'port must be in range [1024, 65535]'
         bittensor.wallet.check_config( config )
 
+    @classmethod   
+    def default_synapse_check(cls, synapse, hotkey ):
+        """ default synapse check function
+        """
+        assert len(hotkey) == 48
+
     @staticmethod
     def check_backward_callback( backward_callback:Callable, pubkey:str = '_' ):
         """ Check and test axon backward callback function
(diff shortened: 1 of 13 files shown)


bittensor/_axon/__init__.py
```diff
@@ -54,6 +54,7 @@ def __new__(
             synapse_last_hidden: 'Callable' = None,
             synapse_causal_lm: 'Callable' = None,
             synapse_seq_2_seq: 'Callable' = None,
+            synapse_checks: 'Callable' = None,
```
joeylegere, line 57: nit: we can use straight up `typing.Callable` in our type annotations since we're pinned to Python >= 3.5 now.

bittensor/_axon/__init__.py
```diff
@@ -235,6 +241,12 @@ def check_config(cls, config: 'bittensor.Config' ):
         assert config.axon.port > 1024 and config.axon.port < 65535, 'port must be in range [1024, 65535]'
         bittensor.wallet.check_config( config )
 
+    @classmethod   
+    def default_synapse_check(cls, synapse, hotkey ):
+        """ default synapse check function
+        """
+        assert len(hotkey) == 48
```
joeylegere: what's 48? Can we put it in constants.py with a descriptive name?

bittensor/_axon/__init__.py
```diff
@@ -255,13 +267,13 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
         """
         if not inspect.ismethod(forward_callback) and not inspect.isfunction(forward_callback):
             raise ValueError('The axon forward callback must be a function with signature Callable[inputs_x: torch.Tensor] -> torch.FloatTensor:, got {}'.format(forward_callback))   
-        if len( inspect.signature(forward_callback).parameters) != 2:
+        if len( inspect.signature(forward_callback).parameters) != 3:
             raise ValueError('The axon forward callback must have signature Callable[ inputs_x: torch.Tensor, synapses] -> torch.FloatTensor:, got {}'.format(inspect.signature(forward_callback)))
```
isabella618033: should have shape 3 here

bittensor/_cli/__init__.py
```diff
@@ -776,6 +785,10 @@ def check_run_config( config: 'bittensor.Config' ):
         if config.model == 'None' and not config.no_prompt:
             model = Prompt.ask('Enter miner name', choices = list(bittensor.neurons.__text_neurons__.keys()), default = 'template_miner')
             config.model = model
+
+        if 'server' in config.model and not config.no_prompt:
```
isabella618033, line 815: what if the user want to server multiple synapse?

bittensor/_cli/cli_impl.py
```diff
@@ -189,8 +189,18 @@ def run_miner ( self ):
         # Run miner.
         if self.config.model == 'template_miner':
             bittensor.neurons.template_miner.neuron().run()
+
         elif self.config.model == 'core_server':
-            bittensor.neurons.core_server.neuron().run()
+
+            if self.config.synapse == 'TextLastHiddenState':
+                bittensor.neurons.core_server.neuron(lasthidden=True, causallm=False, seq2seq = False).run()
```
isabella618033, line 197: emmm maybe we should put the arg as a list, instead of having only either this or that synapse

bittensor/_axon/__init__.py
```diff
@@ -255,13 +267,13 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
         """
         if not inspect.ismethod(forward_callback) and not inspect.isfunction(forward_callback):
             raise ValueError('The axon forward callback must be a function with signature Callable[inputs_x: torch.Tensor] -> torch.FloatTensor:, got {}'.format(forward_callback))   
-        if len( inspect.signature(forward_callback).parameters) != 2:
+        if len( inspect.signature(forward_callback).parameters) != 3:
```
joeylegere, line 273: This is probably the sort of thing we could enforce with `mypy` or a static checker if we use [typing.Callable](https://docs.python.org/3/library/typing.html#typing.Callable). Then we also don't have to update this line (and any others) if the expected signature of the function changes.

bittensor/_neuron/text/core_server/run.py
```diff
@@ -156,17 +159,33 @@ def time_check():
...
 
         except Exception as e:
             return True
+    
+    def synapse_check(synapse, hotkey):
+        """
+        Custom synapse function to blacklist certain synapse functions depending on the stake
+        """
+        #TODO turn on synapse checking function
+        if synapse.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_LAST_HIDDEN_STATE:
```
isabella618033, line 180: is it just a place holder for now?

bittensor/_axon/__init__.py
joeylegere: I feel like the signature of this function (and the one in `bittensor/_neuron/text/core_server/run.py`) should return a `bool` indicating whether or not it passes, rather than raising an `Exception`.

This way if you have multiple checks within the function, you can (if necessary) log all the failed ones (rather than the first one that fails), and you can follow the logic in the outer scope a bit more cleanly with a `if synapse_check(synapse, hotkey)`.

bittensor/_synapse/text_seq2seq_impl.py
```diff
@@ -147,13 +147,12 @@ def check_forward_request_tensor     ( self, forward_request_tensor ):
     def check_forward_response_tensor    ( self, forward_request_tensor, forward_response_tensor ):
         if forward_response_tensor == None:
             raise ValueError('Empty Response')
-
         if (
              len( forward_response_tensor.shape ) != 2 or
              forward_response_tensor.size(0) != forward_request_tensor.size(0) or
-             forward_response_tensor.size(1) != self.num_to_generate
+             forward_response_tensor.size(1) > self.num_to_generate
```
joeylegere, line 153: what change necessitates the more lenient check here?

bittensor/_axon/__init__.py
Eugene-hu: Good point; we actually use that quite a bit. Let's add it as a ticket, we should change all of the 'Callable' to proper typing

bittensor/_axon/__init__.py
Eugene-hu: Nice Catch, let me add it as the length of the keys

bittensor/_axon/__init__.py
Eugene-hu: Yea, I agree. This check actually requires quite a lot of maintenance. We will need to make sure the static checker is compatible with `attach_forward_function` in the axon.

bittensor/_axon/__init__.py
Eugene-hu: Good point; it is probably cleaner and easier to understand rather than an `Exception`. Let me change that!

bittensor/_synapse/text_seq2seq_impl.py
Eugene-hu: The check was improperly formatted, num_to_generate only dictates the maximum length of the generation.

bittensor/_axon/__init__.py
Eugene-hu: Haha nice catch; let me change that

bittensor/_cli/__init__.py
Eugene-hu: Yea, Im still debating on what is the best design. Im kind of torn between forcing the user to type out all of the synapses it wants to support, or to prompt the user multiple times. Right now, the default value will serve all of the synapses

bittensor/_cli/cli_impl.py
Eugene-hu: hmm good point, it can get too verbose, but I could see both cases being useful. Let's add both options!

bittensor/_neuron/text/core_server/run.py
Eugene-hu: Yeap, we will turn on blacklist, and checking before the launch

bittensor/_cli/__init__.py
isabella618033: yea i think type out all the synapses is better

bittensor/_cli/cli_impl.py
isabella618033: later on, maybe lets revisit our UI, cause we are having soooo many options now, maybe time to bring back the use of config file?

bittensor/_neuron/text/core_server/run.py
isabella618033: :o so what is the difference between blacklist and synapse check?

bittensor/_cli/cli_impl.py
Eugene-hu: Yeap, sounds good. A config file is a good idea! I believe we should still support that, but the code is quite old

bittensor/_neuron/text/core_server/run.py
Eugene-hu: the blacklist is checked based on the metadata of the grpc request while the check is based on individual synapses. So blacklist will reject entire requests, while synapse checks will reject individual synapse calls

PR #798, opened 2022-06-27
Synapse fix dend test

isabella618033:
- dendrite test fixes
- use [EOS] as the pad token


tests/integration_tests/test_dendrite.py
```diff
@@ -33,11 +33,19 @@
     modality = 0
 )
 
+synapses = [bittensor.synapse.TextLastHiddenState(),bittensor.synapse.TextCausalLM(), bittensor.synapse.TextSeq2Seq(num_to_generate=70)]
+
+def check_resp_shape(resp, num_resp, block_size, seq_len):
+    assert len(resp) == num_resp
+    assert list(resp[0][0].shape) == [block_size, seq_len, bittensor.__network_dim__]
+    assert list(resp[0][1].shape) == [block_size, seq_len, bittensor.__vocab_size__]
+    assert list(resp[0][2].shape) == [block_size, 70]
```
joeylegere: what does 70 represent here?

isabella618033: welp you caught me, 
it is the number of tokens to generate for seq2seq
just moved it to constant.py

PR #800, opened 2022-06-27
Feature/dont ask if flag set

camfairchild:
Previously, unset flags for `subtensor.network`, `wallet.name`, and `wallet.hotkey` had a default value that was not `None`
We would check the config to see if the defaults are set, in order to determine if we should prompt for a new value.
This is an issue because if a user wanted the default as the value, and set it with a flag, the user would still be prompted.

This PR changes the default flag value to `argparse.SUPPRESS`, which removes the value from the config namespace, and allows us to check if the flag was set at all.

To accommodate this, we also now check those values with `config.<parent>.get('<child>', <bittensor.default.<parent>.<child>)` to have the same effect as keeping the default

### DIFF
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -163,7 +163,7 @@ class CLI:
         # Check coldkey.
         wallet = bittensor.wallet( config = self.config )
         if not wallet.coldkeypub_file.exists_on_device():
-            if Confirm.ask("Coldkey: [bold]'{}'[/bold] does not exist, do you want to create it".format(self.config.wallet.name)):
+            if Confirm.ask("Coldkey: [bold]'{}'[/bold] does not exist, do you want to create it".format(self.config.wallet.get('name', bittensor.defaults.wallet.name))):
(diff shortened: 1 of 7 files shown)


bittensor/_cli/cli_impl.py
```diff
@@ -130,9 +130,9 @@ def inspect ( self ):
         dendrite = bittensor.dendrite( wallet = wallet )
 
         
-        with bittensor.__console__.status(":satellite: Looking up account on: [white]{}[/white] ...".format(self.config.subtensor.network)):
+        with bittensor.__console__.status(":satellite: Looking up account on: [white]{}[/white] ...".format(self.config.subtensor.get('network', bittensor.defaults.subtensor.network))):
             
-            if self.config.wallet.hotkey == 'None':
+            if self.config.wallet.get('hotkey', bittensor.defaults.wallet.hotkey) == 'None':
```
joeylegere: why is `None` interpolated to a string here / where does that happen?

joeylegere (approved):

PR #810, opened 2022-06-29
Bit 490 synapse one fail drag all down

isabella618033:
Some bug fixes
- axon side: synapse response default to be synapse.empty()
- axon side: even if one of the synapse failed, the running result should always be added to response_synapses.
- if core server is not training locally or remotely, then it should sleep through the block time
- update model only if we do local train or remote train
- remove empty reshape in core validator

### DIFF
--- a/bittensor/_neuron/text/core_server/run.py
+++ b/bittensor/_neuron/text/core_server/run.py
@@ -284,6 +284,12 @@ def serve(
             
             if iteration != 0:
                 (losses/iteration).backward()
+        
+        else:
+            while end_block >= current_block:
+                time.sleep(12)
+                current_block = subtensor.get_current_block()
+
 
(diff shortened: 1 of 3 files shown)


bittensor/_neuron/text/core_server/run.py
```diff
@@ -305,6 +305,12 @@ def backward_callback(inputs_x:torch.FloatTensor, grads_dy:torch.FloatTensor, sy
             
             if iteration != 0:
                 (losses/iteration).backward()
+        
+        else:
+            while end_block >= current_block:
+                time.sleep(12)
```
Eugene-hu, line 311: can we use `bittensor.__blocktime__ ` instead?

Eugene-hu (approved):

PR #811, opened 2022-06-29
Read coldkey password from environment variables

joeylegere:
Allows reading in the password for btcli operations from an environment variable rather than explicitly from the commandline. If you don't pass the `--no_prompt` flag, you will still be prompted before performing a stake/unstake/transfer etc. as a safeguard.

This will search your current environment variables for something formatted as `BT_COLD_PW_XXXX` where `XXXX` is the name of your coldkey (in order to enable automatic staking for those with multiple cold keys on the same machine).

Note that coldkey names are *not* case sensitive. What kind of monster has keys with the same name with different capitalization to differentiate them?? why would we even support that in the first place?

example usage:
```
btcli stake --amount 10  --subtensor.network nobunaga --wallet.name my_test_wallet --wallet.hotkey default --no_prompt 
```

### DIFF
--- a/bittensor/_keyfile/keyfile_impl.py
+++ b/bittensor/_keyfile/keyfile_impl.py
@@ -191,28 +191,11 @@ def encrypt_keyfile_data ( keyfile_data:bytes, password: str = None ) -> bytes:
 
 def get_coldkey_password_from_environment(coldkey_name: str) -> Optional[str]:
 
-    # hey bro i heard u like list comprehensions
-    try:
-        return str(
-            os.environ[
-                [
-                    x
-                    for x in list(
-                        map(
-                            lambda x: x if coldkey_name in x.lower() else None,
-                            {
-                                env_var
-                                for env_var in os.environ
-                                if "bittensor_cold_pw_" in env_var.lower()
-                            },
-                        )
-                    )
-                    if x is not None
-                ][0]
-            ]
-        )
-    except IndexError:
-        return None
+    candidate_pw_vars = set()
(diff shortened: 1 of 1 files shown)


bittensor/_keyfile/keyfile_impl.py
```diff
@@ -186,7 +188,17 @@ def encrypt_keyfile_data ( keyfile_data:bytes, password: str = None ) -> bytes:
         vault = Vault( password )
     return vault.vault.encrypt ( keyfile_data )
 
-def decrypt_keyfile_data( keyfile_data: bytes, password: str = None) -> bytes:
+
+def get_coldkey_password_from_environment(coldkey_name: str) -> Optional[str]:
+
+    candidate_pw_vars = set()
+    for env_var in os.environ:
+        if env_var.lower().startswith("bittensor_cold_pw_") and coldkey_name.lower() in env_var.lower():
```
Eugene-hu: So to use it, the environment variable must be "bittensor_cold_pw_(wallet_name)"?

Eugene-hu: Small nit, can we change it to "BT_(...)"? This put it in line with the other environment variables that we use for the defaults

Eugene-hu (approved):

bittensor/_keyfile/keyfile_impl.py
joeylegere: it's actually case-insensitive because i'm comparing it to a call with `.lower()`, but I could do `.upper()` instead to make it match.

Issue #804, opened 2022-06-28
Change default action so that it does not automatically attempt to re-register hotkeys


hueymahl:
Currently when a hotkey is deregistered, bittensor starts a re-registration process for that hotkey by default.  The problem with this is that if you are running multiple miners on the same hardware, the stress of the re-registration attempt will max out the machine's resources and negatively impact the performance of the other miners.  This, in-turn, can lead to a cascade of hotkeys being deregistered.

I would prefer for the default action to be that the miner simply terminates and does not attempt to re-register when there is a deregistration event.  For those that wish to keep this feature, there can be a command that turns it on.

Alternatively, create a command to turn off auto-reregistration if the default action is left as-is.

Not re-registering was the default action previously.  If my memory serves, the justification for changing it was to make it easier to run a miner.  Then it made sense because the registration difficulty was negligible and did not negatively impact other running miners. That is no longer the case.

joeylegere:
Closing this because the feature is done and will be merged into master as part of Synapse.

PR #819, opened 2022-07-04
Adds flag --wallet.reregister <bool>

camfairchild:
This PR addresses #804 
If the wallet is not registered, by default, it will attempt to re-register the wallet
if the flag `--wallet.reregister` is set to `false`, it will exit with `0` instead.

If using pm2, the restart of the miner can be prevented using:  
https://pm2.keymetrics.io/docs/usage/restart-strategies/#skip-auto-restart-for-specific-exit-codes

Edit: changed the flag to `--wallet.reregister`


joeylegere:
I think this PR could be simplified if the flag were `--wallet.reregister` and the logic was contained in the `Wallet.register` method instead. It's a bit more honest about what's happening (because technically the wallet object is what gets registered, not the neuron), and it'll also make our lives easier if/when we add/modify new neurons in the future (so we won't have to remember to include the logic when we check if we're registered or not).

camfairchild:
> I think this PR could be simplified if the flag were `--wallet.reregister` and the logic was contained in the `Wallet.register` method instead. It's a bit more honest about what's happening (because technically the wallet object is what gets registered, not the neuron), and it'll also make our lives easier if/when we add/modify new neurons in the future (so we won't have to remember to include the logic when we check if we're registered or not).

Done

joeylegere (approved):
LGTM!

PR #826, opened 2022-07-11
[Fix] tensors on two devices issue, core-validator 

camfairchild:
This fixes the tensors on two devices issue that happens with core validators using cuda devices on master


camfairchild:
another bug with this. don't merge yet

PR #827, opened 2022-07-11
Null synapse

Eugene-hu:
Changes the default synapse to the Null synapse; this will act as the main communication tool when a server is given a synapse, it is not familiar with.

Also added parameters to the seq2seq proto such as temperature, repetition_penalty, length_penalty, max_time, num_beam_groups.


bittensor/_axon/axon_impl.py
```diff
@@ -334,6 +334,8 @@ def finalize_codes_stats_and_logs():
                 else:
                     synapse_responses [ index ] = synapse.empty()
 
+                response_synapses.append(synapse.serialize_to_wire_proto(code = synapse_codes[index], message= synapse_messages[index] ))
```
isabella618033: (1) indent error maybe (?)
(2) there were case where when it failed somehow, response_synapse did not got appended and  there is one short response_synapse and it throw some weird error

isabella618033 (approved):

bittensor/_axon/axon_impl.py
Eugene-hu: I think the ident seems fine? haha maybe its just display funny. (2) Good point, there should be an .append when an exception is hit. Let me revert the change, its probably cleaner before.

opentaco (approved):

PR #828, opened 2022-07-12
Bit 495 synapse more unit test

isabella618033:
added test cases
- on the dendrite <-> axon level
(1) when there is 0-3 missing synapse
(2) when there is 0-3 synapse with error

- on the receptor / receptor-pool
(1) when the number of return tensor < number of synapse
(2) when the number of return tensor > number of synapse
(3) when part of the return synapse having shape error
(4) when part of the return synapse has unknown error


bittensor/_receptor/receptor_impl.py
```diff
@@ -573,7 +573,7 @@ def finalize_stats_and_logs():
         # ======================================
         # ==== Check response length ====
         # ======================================
-        if len( grpc_response.tensors ) != len( synapses ):
+        if ( len( grpc_response.tensors ) != len( grpc_response.synapses ) )or ( len( grpc_response.tensors ) != len( synapses ) ):
```
Eugene-hu: Mega small nit: I think you are missing a space before the `or` XD

Eugene-hu (approved):
LGTM

bittensor/_receptor/receptor_impl.py
isabella618033: waaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

PR #831, opened 2022-07-12
[BIT 487] btcli regen coldkey with pubkey only

camfairchild:
This PR adds the ability to regen a coldkeypub.txt only instead of the whole coldkey
This allows users to avoid exposing their coldkey private part to an untrusted server provider, for instance.

`btcli regen_coldkeypub --wallet.name newtestpub --ss58_address 5HY<...>H`


bittensor/utils/__init__.py
```diff
@@ -308,20 +308,24 @@ def is_valid_destination_address( address: Union[str, bytes] ) -> bool:
         # Check if ed25519
         if address.startswith('0x'):
             if not is_valid_ed25519_pubkey( address ):
-                bittensor.__console__.print(":cross_mark: [red]Invalid Destination Public Key[/red]: {}".format( address ))
+                if not no_prompt:
```
joeylegere: why if not no_prompt? will this potentially fail silently if the user passes `--no_prompt` along with a destination key?

joeylegere (approved):
left a question. LGTM though

bittensor/utils/__init__.py
camfairchild: it will fail silently, but it will still return False so then it's up to the calling function to give output.

The addition there is because I'm using it to validate the public key/ss58_address which is not a destination address. So perhaps I should move the output to the transfer function.

PR #833, opened 2022-07-13
Remove template miner

Eugene-hu:
-Deprecate Template miner
-Adds priority for the core server


scripts/update.sh
```diff
@@ -0,0 +1,232 @@
+
```
isabella618033: why is this file here?

tests/unit_tests/bittensor_tests/test_neuron.py
```diff
@@ -237,89 +237,6 @@ def exit_early(*args, **kwargs):
         # Should try to register the neuron
         mock_register.assert_called_once()
 
-def test_templateminer_reregister_flag_false_exit():
```
isabella618033, line 240: i think this is sth vune added recently, and it checkes for registration, so maybe convert that for sth core_server?

isabella618033: same for the tst below

isabella618033 (commented):
LGTM, left you small comments xd

tests/unit_tests/bittensor_tests/test_neuron.py
Eugene-hu: Theres one for core_server and core_validator already. So I just removed the tests for template miner

scripts/update.sh
Eugene-hu: Oops nice catch, that shouldnt be there

bittensor/_neuron/text/core_server/run.py
```diff
@@ -166,7 +186,7 @@ def time_check():
         try:
             registration_check()
 
-            stake_check()
+            #stake_check()
```
shibshib, line 189: if we don't need this let's just delete it entirely.

shibshib (approved):

bittensor/_neuron/text/core_server/run.py
Eugene-hu: This is temporary; we will need to turn on these checks when we get to Nakamoto

PR #836, opened 2022-07-13
seq2seq defaults updates

Eugene-hu:
- Longer server computation time
- lower seq2seq default

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -206,7 +206,7 @@ class axon:
             parser.add_argument('--' + prefix_str + 'axon.backward_timeout', type=int,
                 help='Number of seconds to wait for backward axon request', default=2*bittensor.__blocktime__)
             parser.add_argument('--' + prefix_str + 'axon.forward_timeout', type=int,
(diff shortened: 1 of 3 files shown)


bittensor/_axon/__init__.py
```diff
@@ -206,7 +206,7 @@ def add_args( cls, parser: argparse.ArgumentParser, prefix: str = None  ):
             parser.add_argument('--' + prefix_str + 'axon.backward_timeout', type=int,
                 help='Number of seconds to wait for backward axon request', default=2*bittensor.__blocktime__)
             parser.add_argument('--' + prefix_str + 'axon.forward_timeout', type=int,
-                help='Number of seconds to wait for forward axon request', default=bittensor.__blocktime__)
+                help='Number of seconds to wait for forward axon request', default=10*bittensor.__blocktime__)
```
isabella618033: ermmmmmmmmmm are you sure this is good....? cause some bad guy can just send huge request and use up 2 mins of the server

Eugene-hu: good point, let push it down to 1 minute, we will add back some of the blacklisting and checking feature before nakamoto launch

isabella618033 (approved):

PR #838, opened 2022-07-14
Model Parallelism

robertalanm:
Utilizing the Accelerate package from HuggingFace allows us to run larger models than 6.7B.


joeylegere:
should this be based off of `Synapse` and have the changes in `core_server` instead?

PR #848, opened 2022-07-20
[BIT-484] Add TextCausalLMNext

opentaco:
Adds the TextCausalLMNext Synapse, as well as the tokenizer_utils needed to encode/decode the synapse. Both core_server and core_validator are running successfully on Nobunaga with this branch.

TextCausalLMNext sends a user-defined topk of server token phrases for input context continuation, which validator/client uses for validated generation where server phrase probabilities are evaluated against a groundtruth continuation. Supports backward gradients from validator/client to server.

Check detailed commit explanations.

### DIFF
--- a/bittensor/_dendrite/dendrite_impl.py
+++ b/bittensor/_dendrite/dendrite_impl.py
@@ -436,7 +436,80 @@ class Dendrite(torch.autograd.Function):
         self.update_stats( formatted_endpoints, synapses, formatted_inputs, outputs, codes, times )
         return outputs[0], codes[0], times[0]
 
-    def text_last_hidden_state( 
+    def text_causal_lm_next(
+            self,
+            endpoints: Union[torch.LongTensor, List[torch.LongTensor], List['bittensor.Endpoint'], 'bittensor.Endpoint'],
+            inputs: Union[str, List[str], List[torch.LongTensor], torch.LongTensor],
+            synapse: Optional['bittensor.synapse.TextCausalLMNext'] = synapse.TextCausalLMNext(),
+            timeout: Optional[int] = None,
+            requires_grad: Optional[bool] = None,
+    ) -> Tuple[Union[List[torch.FloatTensor], torch.FloatTensor], torch.LongTensor, torch.FloatTensor]:
+        r""" Forward text inputs to a list of neuron endpoints and returns logit encodings or timeout.
+
+                Args:
+                    endpoints (:obj:`Union[torch.LongTensor, List[torch.LongTensor], List[bittensor.Endpoint], bittensor.Endpoint]` of shape :obj:`(num_endpoints)`, `required`):
+                        Endpoints to send inputs to. Endpoint can be one of the following types:
+                            - a single endpoint tensor shape [250]
+                            - a set of endpoint tensors shape [n, 250]
+                            - a list of endpoints tensors each of shape [250]
+                            - a single endpoint object. Inputs will be sent to this endpoint alone.
+                            - a list of endpoint objects. All inputs will be sent to these endpoints.
+
+
+                    inputs (:obj:`Union[str,  List[str], List[torch.LongTensor], torch.LongTensor]` of shape :obj:`(num_endpoints * [batch_size, sequence_len])`, `required`):
+                        Tokenized sentences to send on the wire. Inputs can be one of the following types:
+                            - a single string: the string will be tokenized using the bittensor tokenizer.
+                            - a list of strings: the strings will be tokenized using the bittensor tokenizer.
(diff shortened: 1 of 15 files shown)


bittensor/_dendrite/dendrite_impl.py
```diff
@@ -436,7 +436,80 @@ def text_causal_lm (
         self.update_stats( formatted_endpoints, synapses, formatted_inputs, outputs, codes, times )
         return outputs[0], codes[0], times[0]
 
-    def text_last_hidden_state( 
+    def text_causal_lm_next(
```
Eugene-hu, line 439: Should we expect this synapse to be called on its own or always in conjunction with CausalLM?

bittensor/_synapse/__init__.py
```diff
@@ -101,6 +102,38 @@ def TextCausalLM (
...
+        topk: int = 4096,
+        forward_request_serializer_type: 'bittensor.proto.Serializer.Type' = bittensor.proto.Serializer.MSGPACK,
+        forward_response_serializer_type: 'bittensor.proto.Serializer.Type' = bittensor.proto.Serializer.MSGPACK,
+        backward_request_serializer_type: 'bittensor.proto.Serializer.Type' = bittensor.proto.Serializer.MSGPACK,
+        backward_response_serializer_type: 'bittensor.proto.Serializer.Type' = bittensor.proto.Serializer.MSGPACK,
+    ) -> TextCausalLMNext:
+        """ Factory function which returns a TextCausalLMNext synapse adapter given arguments.
+            Args:
+                topk (:obj:`int`):
+                    Specifies the number of topk server token phrases to return.
```
Eugene-hu, line 116: A little bit confused about the topk variable. Are they the topk logits of the server's tokenizer?

bittensor/_synapse/text_causallmnext_impl.py
```diff
@@ -0,0 +1,170 @@
...
+        ):
+            raise ValueError(f"backward_request_gradient.shape must be in "
+                             f"[>={forward_request_tensor.shape[0]} x (2 x {self.topk} + 1)], "
+                             f"got: {backward_request_gradient.size(0)} for synapse: {self}")
+
+    def encode_forward_request_tensor(self, forward_request_tensor: torch.Tensor) -> torch.Tensor:
+        return forward_request_tensor
+
+    def decode_forward_request_tensor(self, forward_request_tensor: torch.Tensor) -> torch.Tensor:
+        return forward_request_tensor
```
Eugene-hu, line 140: Do we need a encoding or decoding function?

bittensor/_synapse/text_causallmnext_impl.py
```diff
@@ -0,0 +1,170 @@
...
+                len(forward_request_tensor.shape) != 2 or
+                forward_request_tensor.shape[0] == 0 or
+                forward_request_tensor.shape[1] == 0
+        ):
+            raise ValueError(f"forward_request_tensor.shape must be in [-1, -1], "
+                             f"got: {list(forward_request_tensor.shape)} for synapse: {self}")
+
+    def check_forward_response_tensor(self, forward_request_tensor, forward_response_tensor):
+        # forward_request_tensor: [batch_size, sequence_len]
+        # forward_response_tensor: [ >= batch_size * (2 * topk + 1)]
```
Eugene-hu, line 113: The response tensor is a 1-D tensor?

Eugene-hu (approved):
Approved, left you a few questions

bittensor/_dendrite/dendrite_impl.py
opentaco: TextCausalLMNext can be called independently of TextCausalLM, but for initial tests they'll both run so comparisons can be performed.

bittensor/_synapse/__init__.py
opentaco: Yes, topk logits/phrases, under the assumption that each logit represents a text phrase (the server represents this with a single token with associated probability, but it could be multiple std tokens (a token phrase) at the validator).

bittensor/_synapse/text_causallmnext_impl.py
opentaco: Currently, encoding and decoding is best performed at the origin, that is server and validator forward() functions respectively. The reason being that variable length phrases need to otherwise be converted to padded tensors, simply for the purpose of passing it to the synapse encode_forward_request_tensor(), where it would need to be decoded and then re-encoded and compacted, thus duplicating work best done immediately at the origin.

bittensor/_synapse/text_causallmnext_impl.py
opentaco: Yes, the response is 1-D to efficiently compact variable length token phrases by omitting padding that would otherwise be required to make it n-D.

PR #852, opened 2022-07-24
[BIT-484] Server/validator improvements (TextCausalLMNext)

opentaco:
### [BIT-484] Server/validator improvements (TextCausalLMNext)

Various synapse improvements, highlights include:

1. Critical tokenizer fixes like setting server tokenizer padding_side.
2. Support for facebook/opt-* models and other non-fast tokenizer models.
3. Validator checks for NaNs in neuron_stats.
4. Gradual EMA from zero for neuron_stats used for weighting (to improve mean estimation).
5. Added normal validation loss to TextCausalLMNext.

Tested branch successfully for core_server and core_validator on CUDA, including running facebook/opt-13b.

[BIT-484]: https://opentensor.atlassian.net/browse/BIT-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -78,6 +84,9 @@ def __init__(self,
         if self.config.neuron.autocast and self.device[:4] == 'cuda':
             self.pre_model.half()
 
+        # Generative default expects most recent token on right-hand side with padding on left. https://github.com/huggingface/transformers/pull/10552
+        self.tokenizer.padding_side = "left"
```
Eugene-hu, line 88: Can we use the bittensor.tokenizer.prep_tokenizer function? Just for consistency and in case we write more miners in the future.

bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -362,42 +356,37 @@ def encode_forward_causallm(self, token_batch, tokenizer=None, encode_len=bitten
...
+                                                      self.tokenizer, self.std_tokenizer,
+                                                      self.split_map_cache,
+                                                      self.to_translation_map, self.from_translation_map,
+                                                      tokens['input_ids'], token_batch)
+            probs_std = probs_std.to(self.device)
+            logits_std = torch.log(probs_std + 1e-40)
+
+            original_loss = self.get_loss_fct(pre_logits, tokens['input_ids'])
+            translated_loss = self.get_loss_fct(logits_std, token_batch)
+            logger.info(f'TextCausalLM \t| Server loss: {original_loss: .2f} \t| Translated loss: {translated_loss: .2f}')
```
Eugene-hu, line 381: Out of curiosity, how much is the difference in loss?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -60,32 +60,33 @@
 neuron_stats_columns = [
     ['UID', 'uid', '{:.0f}', 'cyan'],  # neuron UID
```
Eugene-hu, line 61: Maybe we should put these on the docs haha

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -462,17 +463,12 @@ def neuron_stats_update(self, neuron_stats: Dict[int, Dict[str, Any]]):
                 stats.setdefault('updates!', 1)  # number of EMA zeroing updates init to zero
 
             for key in self.synapse_keys:
-                zkey = key + '!'
-                if zkey in stats:
-                    if key in _stats:
-                        stats[zkey] = (1 - self.alpha) * stats[zkey] + self.alpha * _stats[key]
-                    else:
-                        stats[zkey] = (1 - self.alpha) * stats[zkey]  # + self.alpha * 0
+                zkey = key + '!'  # zeroing key
```
Eugene-hu, line 466: What is this zkey? Is that how we are indexing the stats?

Eugene-hu (approved):
LGTM! Amazing work!

bittensor/_neuron/text/core_server/nucleus_impl.py
opentaco: Yes, we can use prep_tokenizer, will do it in another pull request.

bittensor/_neuron/text/core_server/nucleus_impl.py
opentaco: Translation loss can vary on average between 0.1 - 0.3 nats above original loss.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, we'll do that once the table columns have been finalized.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: The zeroing key is a copy of a stat where unsuccessful responses mean a zero will EMA into the value, thus decaying the neuron's key statistic according to which it is weighted. The original key statistic does not push a zero on unsuccessful responses, but only on valid responses, such that the original key is an accurate representation of neuron performance when responsive.

PR #854, opened 2022-07-25
[feature] cpu register faster

camfairchild:
Previously, multiprocessed CPU registration created a new pool of workers (processes) every update interval of nonces. This has some overhead and is thus inefficient.

This PR fixes that and instead keeps the pool until a potential solution is found. 
From some limited testing (1-core, 6-core, 16-core) I saw a 20-30% increase in hashrate.


bittensor/utils/__init__.py
```diff
@@ -102,7 +106,126 @@ def solve_for_difficulty( block_hash, difficulty ):
...
+        if (product - limit) < best_local: 
+            best_local = product - limit
+            best_seal_local = seal
+
+    # Send best solution to best queue.
+    solver.best_queue.put((best_local, best_seal_local))
+    return None, time.time() - start
+
+
+def solve_for_difficulty_fast( subtensor, wallet, num_processes: int = None, update_interval: int = 50_000 ) -> Optional[POWSolution]:
```
joeylegere: Probably better form if we have `update_interval: Optional[int] = None` since we override to the default value below anyways. (And if we change it from 50k, then it only changes in one spot)

joeylegere: also nit: typing on `num_processes` should be `Optional[int]` as well given its default value.

camfairchild: Added

camfairchild:
Also added https://github.com/opentensor/bittensor/pull/854/commits/5ac648b5762533ba4a23f16978bc0356eda9ec07
which changes from `cpu_count()` to the number of allowed processes
See: https://stackoverflow.com/a/55423170

joeylegere (approved):

bittensor/utils/__init__.py
```diff
@@ -102,7 +105,126 @@ def solve_for_difficulty( block_hash, difficulty ):
...
+def millify(n: int):
+    millnames = ['',' K',' M',' B',' T']
+    n = float(n)
+    millidx = max(0,min(len(millnames)-1,
+                        int(math.floor(0 if n == 0 else math.log10(abs(n))/3))))
+
+    return '{:.0f}{}'.format(n / 10**(3 * millidx), millnames[millidx])
+
+@dataclass
+class POWSolution:
```
Eugene-hu, line 126: Is this a simple storage class for the solution?

bittensor/utils/__init__.py
```diff
@@ -102,7 +105,126 @@ def solve_for_difficulty( block_hash, difficulty ):
...
+    time_queue: multiprocessing.Queue
+    solution_queue: multiprocessing.Queue
+    newBlockEvent: multiprocessing.Event
+    stopEvent: multiprocessing.Event
+    curr_block: multiprocessing.Array
+    curr_block_num: multiprocessing.Value
+    curr_diff: multiprocessing.Array
+    check_block: multiprocessing.Lock
+    limit: int
+
```
Eugene-hu, line 191: Would love some documentation on this. This inherits from a multiprocessing process, correct? Would love some explanation of the different queues.

bittensor/utils/__init__.py
```diff
@@ -115,126 +237,155 @@ def solve_for_difficulty_fast( subtensor, wallet, num_processes: int = None, upd
...
+    # Establish communication queues
+    stopEvent = multiprocessing.Event()
+    stopEvent.clear()
+    best_queue = multiprocessing.Queue()
+    solution_queue = multiprocessing.Queue()
+    time_queue = multiprocessing.Queue()
+    check_block = multiprocessing.Lock()
+    
+    # Start consumers
+    solvers = [ Solver(i, num_processes, update_interval, best_queue, time_queue, solution_queue, stopEvent, curr_block, curr_block_num, curr_diff, check_block, limit)
```
Eugene-hu, line 330: I see... so all of these objects are shared across all of the processes and each process has a random starting point?

Eugene-hu (approved):
Looks good to me! Nice job Vune! Left you a few comments; lets push to Nobunaga first and master later

bittensor/utils/__init__.py
camfairchild: yes

bittensor/utils/__init__.py
camfairchild: added some to the class https://github.com/opentensor/bittensor/pull/854/commits/a976934c97e568ef19eb967261cddd9f5482b11d

shibshib (approved):

PR #856, opened 2022-07-27
[BIT-525] Tokenizer improvements (whitespace-preserving)

opentaco:
### [BIT-525] Tokenizer improvements (whitespace-preserving)

Add support for non whitespace-preserving tokenizers.

Prepends strings with a space, in the case of non whitespace-preserving tokenizers like BERT, which should mimic whitespace-preserving token strings more often than not.

Adds error catching on per-UID basis in shapley_base for synapse validation to be more robust.

[BIT-525]: https://opentensor.atlassian.net/browse/BIT-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_tokenizer/__init__.py
+++ b/bittensor/_tokenizer/__init__.py
@@ -19,7 +19,7 @@
 
 from transformers import AutoTokenizer
 import bittensor
-from bittensor.utils.tokenizer_utils import set_vocab_len, set_whitespace_preserving
+from bittensor.utils.tokenizer_utils import prep_tokenizer
 
 
 class tokenizer:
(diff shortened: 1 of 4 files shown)


bittensor/_tokenizer/__init__.py
```diff
@@ -18,7 +18,9 @@
 # DEALINGS IN THE SOFTWARE.
 
 from transformers import AutoTokenizer
-import bittensor 
+import bittensor
+from bittensor.utils.tokenizer_utils import prep_tokenizer
```
Eugene-hu, line 22: Would we expect users to use the prep_tokenizer function a lot? If so, it might be good to have it as a class function as well

Eugene-hu (approved):
LGTM!

bittensor/_tokenizer/__init__.py
opentaco: Not really, it's just to prepare the server tokenizer (and the bittensor.tokenizer gets prepared automatically). So clients wouldn't need to run it, for instance.

PR #860, opened 2022-08-02
[BIT-527] Update validator neuron stats display

opentaco:
### [BIT-527] Update validator neuron stats display

Update validator's neuron_stats display, fix sorting, and add weights column to stats update table, and a params estimate column.

[BIT-527]: https://opentensor.atlassian.net/browse/BIT-527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -563,7 +563,8 @@ class nucleus( torch.nn.Module ):
         super(nucleus, self).__init__()
         self.config = config
         self.device = device
-        self.max_n = subtensor.max_n 
+        self.max_n = subtensor.max_n
+
         tokenizer = bittensor.tokenizer()
         self.pad_token = tokenizer(tokenizer.pad_token)['input_ids'][0]
 
@@ -604,6 +605,7 @@ class nucleus( torch.nn.Module ):
         parser.add_argument('--nucleus.importance', type=float, help='hyperparameter for the importance loss', default=3)
         parser.add_argument('--nucleus.noise_multiplier', type=float, help='Standard deviation multipler on weights', default=2 )
         parser.add_argument('--nucleus.dendrite_backward', action='store_true', help='Pass backward request to the server side or not', default=False )
+        parser.add_argument('--nucleus.scaling_law_power', type=float, help='Power for modified scaling law, powered down to improve dynamic range, e.g. 3 → 6 nats for 0.5.', default=0.5)
 
     @classmethod
     def config ( cls ):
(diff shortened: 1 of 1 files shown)


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -511,9 +510,12 @@ def calculate_weights(self):
                                                                            multiple=max_allowed_ratio)
         return topk_uids, topk_weights
 
-    def weights_table(self, topk_uids, topk_weights):
+    def weights_table(self, topk_uids, topk_weights, include_uids=None, num_rows: int = None):
         r""" Prints weights table given topk_uids and topk_weights.
         """
+
+        assert len(include_uids) <= num_rows, f'{len(include_uids)} > {num_rows}'  # ensure all include_uids fit in num_rows
```
Eugene-hu, line 517: What happens if this assertion breaks? Would it stop the validator?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -588,6 +599,7 @@ def add_args( cls, parser ):
         parser.add_argument('--nucleus.importance', type=float, help='hyperparameter for the importance loss', default=3)
         parser.add_argument('--nucleus.noise_multiplier', type=float, help='Standard deviation multipler on weights', default=2 )
         parser.add_argument('--nucleus.dendrite_backward', action='store_true', help='Pass backward request to the server side or not', default=False )
+        parser.add_argument('--nucleus.scaling_law_power', type=float, help='Power for modified scaling law, powered down to improve dynamic range, e.g. 3 → 6 nats for 0.5.', default=0.5)
```
Eugene-hu, line 602: Question: so to improve the dynamic range, we lower this power factor? Ie.  what happens when I set this factor to 2 instead of 0.5?

Eugene-hu (approved):
LGTM! Nice job, Taco!

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, lower power factor to improve dynamic range. Increase power factor to amplify the scaling law, so a factor of 2 would square the parameter count estimate and drastic enlarge the incentive differences.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, it will stop the validator, unless self.config.neuron.restart_on_failure.

I'll comment out the assert in another pull request with some other minor improvements.

PR #861, opened 2022-08-02
Additional Checks and Protections

Eugene-hu:
- Adds protections for servers in anticapation for the upcoming launch
- Seq2seq (generation) synapse now requires a validator that is actively setting a weight on it
- Turns back on Timecheck which protects the server from constant spamming


bittensor/_neuron/text/core_server/run.py
```diff
@@ -208,16 +202,28 @@ def synapse_check(synapse, hotkey):
...
+                return False
+
         elif synapse.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_SEQ_2_SEQ:
-            pass
-        
+
+            if (metagraph.S[incoming_uid] < config.neuron.seq2seq_stake) and (metagraph.S[incoming_uid,  uid]):
+                return False     
+
         return True
```
isabella618033, line 240: should add an else return false here for the case where there is no matching synapse type

bittensor/_neuron/text/core_server/run.py
```diff
@@ -160,44 +160,38 @@ def registration_check():
...
+            # Only check if the request are forward requests
+            if request_type == 'FORWARD':
+                if pubkey in timecheck.keys():
+                    prev_time = timecheck[pubkey]
+                    if current_time - prev_time >= timedelta(seconds=config.neuron.blacklist.time):
+                        timecheck[pubkey] = current_time
+
+                    else:
+                        timecheck[pubkey] = current_time
+                        raise Exception('blacklist')
```
isabella618033: I think last time someone mentioned maybe it would be better if we just return false for these functions instead of raising exception.

bittensor/_neuron/text/core_server/run.py
```diff
@@ -160,44 +160,38 @@ def registration_check():
...
-            if pubkey in timecheck.keys():
-                prev_time = timecheck[pubkey]
-                if current_time - prev_time >= timedelta(seconds=config.neuron.blacklist.time):
-                    timecheck[pubkey] = current_time
-                    return False
+            # Only check if the request are forward requests
+            if request_type == 'FORWARD':
+                if pubkey in timecheck.keys():
+                    prev_time = timecheck[pubkey]
+                    if current_time - prev_time >= timedelta(seconds=config.neuron.blacklist.time):
```
isabella618033: Regarding the time check, 
(1) could we again take a moving average to sample across multiple requests from the same validator, so that it would be a more fair estimation?
(2) maybe it is better to correlate it with stake? ie. if you have higher stake, then the server would allow more frequency. But higher stake =/= more weight set to the server, sooooo it might not be a good metric.

bittensor/_neuron/text/core_server/run.py
```diff
@@ -160,44 +160,38 @@ def registration_check():
...
         # Check for time
         def time_check():
             current_time = datetime.now()
-            if pubkey in timecheck.keys():
-                prev_time = timecheck[pubkey]
-                if current_time - prev_time >= timedelta(seconds=config.neuron.blacklist.time):
-                    timecheck[pubkey] = current_time
-                    return False
+            # Only check if the request are forward requests
+            if request_type == 'FORWARD':
```
isabella618033: if request type is backward, then return false, which means that the validator/client can spam with backward request?

bittensor/_neuron/text/core_server/run.py
```diff
@@ -208,16 +202,28 @@ def synapse_check(synapse, hotkey):
         """
```
isabella618033, line 204: why is the synapse check not part of the blacklist check? what is the difference between these two?

bittensor/_neuron/text/core_server/run.py
Eugene-hu: Yea, the idea was that backwards requests are often much larger to send, but that's still a vulnerability. Let me change it.

bittensor/_neuron/text/core_server/run.py
Eugene-hu: (1) haha wait I'm confused, a moving average of the time between requests? So we reject requests if the moving average drops below a certain value?

Eugene-hu: (2) Yea, I think it might be better to separate the two. Blacklist is designed primarily for protection rather than the priority of service.

bittensor/_neuron/text/core_server/run.py
Eugene-hu: That was for the synapse checks; we need the function to raise an expectation to reject the requests in the interceptor :)

bittensor/_neuron/text/core_server/run.py
Eugene-hu: So the blacklist function operates in the interceptor before the message reaches the axon where the synapse checks are meant to protect individual synapses. For instance, a server may want to protect its seq2seq synapse with strong protections compared to its casuallm, since it requires a lot more compute.

bittensor/_neuron/text/core_server/run.py
Eugene-hu: Haha nice catch! Let me add that

isabella618033 (approved):

PR #868, opened 2022-08-05
[Feature] [cubit] CUDA registration solver

camfairchild:
This PR adds a CUDA solver for the registration POW.
- Requires: [cubit](https://github.com/opentensor/cubit/)
    - see requirements for cubit
- installed as extra `pip install -e .[cubit]`
- will throw error if the user doesn't have a CUDA device and/or cubit
### New Flags
- `--cuda (default: False)` to use CUDA registration (if available)
- `--cuda.dev_id <int> (default: 0)`
- `--cuda.TPB <int, 128, 256, 512> (default: 256)`
    - the number of threads per block in the CUDA kernel
        - should be left at default or raised to 512
        - may crash the registration process if too high
            - power of 2 only
- `--cuda.update_interval <int> (default: 50_000)`
    - the number of nonces to solve between chain updates
        - can be raised or lowered.
            - having a higher value may mean less-frequent chain updates which may lead to trying to submit a solution outside of the valid solve window for that block (not efficient)


shibshib:
Tested this and overall looks good! But looks like when running using btcli run it won't use cuda to register, so these flags don't quite apply there correct?

shibshib (changes requested):
Merge conflicts still to be fixed

camfairchild:
> Tested this and overall looks good! But looks like when running using btcli run it won't use cuda to register, so these flags don't quite apply there correct?

I'll have to add those flags.

shibshib (approved):

PR #869, opened 2022-08-05
Synapse timeout fix

Eugene-hu:
Fixes for the timeout issue in Synapse:
- Includes a fix for the priority thread; now will reject stale requests
- Includes a fix for the axon; now cancels futures once timeout has been reached
- Removed loss calculation for stability

### DIFF
--- a/bittensor/_neuron/text/core_server/nucleus_impl.py
+++ b/bittensor/_neuron/text/core_server/nucleus_impl.py
@@ -373,8 +373,8 @@ class server(torch.nn.Module):
             probs_std = probs_std.to(self.device)
             logits_std = torch.log(probs_std + 1e-40)
 
-            original_loss = self.get_loss_fct(pre_logits, tokens['input_ids'])
-            translated_loss = self.get_loss_fct(logits_std, token_batch)
+            original_loss = self.get_loss_fct(pre_logits, tokens['input_ids']).item()
+            translated_loss = self.get_loss_fct(logits_std, token_batch).item()
             message = f'Loss: {original_loss:.2f} → {translated_loss:.2f}'
             # logger.info(f'TextCausalLM \t| Server loss: {original_loss: .2f} \t| Translated loss: {translated_loss: .2f}')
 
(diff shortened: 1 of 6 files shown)


bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -351,16 +353,14 @@ def encode_forward_causallm(self, token_batch, tokenizer=None, encode_len=bitten
...
         """
-
         tokens = self.token_remap(token_batch, std_tokenizer=tokenizer, return_offsets_mapping=True)  # remap to server tokenizer
 
         def _forward(_model_output=model_output):
             if _model_output is None:
                 # transformer models like gerpt2 typically perform worse with left-side attention mask, so turning it off
                 _model_output = self.pre_model(input_ids=tokens['input_ids'],
-                                               # attention_mask=tokens['attention_mask'],
+                                                attention_mask=tokens['attention_mask'],
```
opentaco: Just note that transformer models like gerpt2 typically perform worse with left-side attention mask, so turning it off is generally advised. On the server-side retokenization there is padding introduced for tokenizers not equivalent to std_tokenizer.

bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -439,8 +440,9 @@ def _forward(_model_output=model_output):
             # then compact new token phrases and probabilities into 1-D tensor
             topk_tensor = topk_token_phrases(last_logits, self.tokenizer, topk=topk)  # [batch_size, (topk + 1), max_len]
 
-            original_loss = self.get_loss_fct(_model_output.logits, tokens['input_ids'])
-            message = f'Loss: {original_loss:.2f}'
```
opentaco, line 443: Showing the original_loss server_side is generally useful feedback to know if the model is performing as expected. Would it be possible to bring this back in future, or was it causing some specific issue?

bittensor/_neuron/text/core_server/nucleus_impl.py
Eugene-hu: Nice catch Taco! That wasnt suppose to be there. I testing different parts of the model

bittensor/_neuron/text/core_server/nucleus_impl.py
Eugene-hu: Let me see if I can bring it back!

opentaco (approved):

PR #872, opened 2022-08-09
[BIT-540] Choose responsive UIDs for setting weights in validator + validator save/load

opentaco:
### [BIT-540] Choose responsive UIDs for setting weights in validator

Prefer current epoch's responsive UIDs, randomly sample min_allowed_weights number of UIDs to set weights to.

Discard max_allowed_ratio in favor of filtering non-zero weights, expecting consensus-levels (number of validators setting weights for UID) to be equal on average for responsive nodes. This means max_allowed_ratio is no longer strictly required for boosting consensus.

If there are less than min_allowed_weights non-zero weights, then set_weights should fail.

---

### [BIT-536] Validator save and load from filesystem

1. Uses core_server save/load format and args.
2. Saves neuron_stats and associated hotkeys.
3. Checks for hotkey changes and updates neuron_stats / neuron_hotkeys (and saves).

[BIT-540]: https://opentensor.atlassian.net/browse/BIT-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[BIT-536]: https://opentensor.atlassian.net/browse/BIT-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -174,7 +174,7 @@ class neuron:
 
         # load last saved validator values from the file system
         if not config.neuron.restart:
-            self.load(config.neuron.full_path)
+            self.load()
 
     @classmethod
     def check_config( cls, config: 'bittensor.Config' ):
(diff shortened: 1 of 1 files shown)


Eugene-hu (approved):
LGTM

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -513,39 +569,60 @@ def neuron_stats_update(self, neuron_stats: Dict[int, Dict[str, Any]]):
...
-        topk_weights = bittensor.utils.weight_utils.normalize_max_multiple(x=topk_weights,
-                                                                           multiple=max_allowed_ratio)
-        return topk_uids, topk_weights
+        # === Filter to non-zero weights ===
+        neuron_weights = neuron_weights[preferred_uids]  # rearrange neuron_weights to match preferred_uids order
+        preferred_uids = preferred_uids[neuron_weights > 0]  # filter to non-zero weights
+        neuron_weights = neuron_weights[neuron_weights > 0]  # filter to non-zero weights
+
+        # === Slice min_allowed_weights UIDs ===
+        sample_uids = preferred_uids[:min_allowed_weights]  # slice to min_allowed_weights
```
isabella618033, line 607: actually i have always been confused, cause `min_allowed_weight` is just the minimum network requirement, 
so shouldnt we set it to `max(min_allowed_weight, len(responsive_uids))` ?

isabella618033 (approved):

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Previously and currently validators always fixed the number of weights set to min_allowed_weight, likely because the demands on subtensor are more predictable then and because validators operate more similarly the consensus is improved.

PR #874, opened 2022-08-11
SynapseType fix in dendrite

robertalanm:
- Fixed an issue where the SynapseType was TextCausalLMNext, when it should be TEXT_CAUSAL_LM_NEXT.


robertalanm:
noticed its the same in other functions, applying fix to those

joeylegere (approved):

shibshib (approved):

PR #876, opened 2022-08-12
[hotfix] fix flags for multiproc register limit

camfairchild:
btcli run calls cli.register() without the correct config flags, so I added them to subtensor args

### DIFF
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -239,7 +239,7 @@ class CLI:
         """
         wallet = bittensor.wallet( config = self.config )
         subtensor = bittensor.subtensor( config = self.config )
-        subtensor.register( wallet = wallet, prompt = not self.config.no_prompt, num_processes = self.config.num_processes, update_interval = self.config.update_interval )
+        subtensor.register( wallet = wallet, prompt = not self.config.no_prompt, num_processes = self.config.get('num_processes', None), update_interval = self.config.get('update_interval', None) )
 
     def transfer( self ):
         r""" Transfer token of amount to destination.
--- a/bittensor/_cli/cli_impl.py
+++ b/bittensor/_cli/cli_impl.py
@@ -239,7 +239,7 @@ class CLI:
         """
         wallet = bittensor.wallet( config = self.config )
         subtensor = bittensor.subtensor( config = self.config )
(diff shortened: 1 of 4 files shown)


opentaco (approved):

bittensor/_cli/cli_impl.py
```diff
@@ -239,7 +239,7 @@ def register( self ):
         """
         wallet = bittensor.wallet( config = self.config )
         subtensor = bittensor.subtensor( config = self.config )
-        subtensor.register( wallet = wallet, prompt = not self.config.no_prompt, num_processes = self.config.num_processes, update_interval = self.config.update_interval )
+        subtensor.register( wallet = wallet, prompt = not self.config.no_prompt, num_processes = self.config.get('num_processes', None), update_interval = self.config.get('update_interval', None) )
```
isabella618033: shouldnt we set a default for num_process and update_interval at the config so that we dont have to handle none?
cause as i rmb, the num_process and update interval will be overrided if they are none withing the registration function anyways

camfairchild: yes, they override at the registration function. We can add a flag to the run command as well, but I figured it best to address the fix in this PR and then add the flag

bittensor/_subtensor/__init__.py
```diff
@@ -197,6 +199,10 @@ def add_defaults(cls, defaults ):
         defaults.subtensor.chain_endpoint = os.getenv('BT_SUBTENSOR_CHAIN_ENDPOINT') if os.getenv('BT_SUBTENSOR_CHAIN_ENDPOINT') != None else None
         defaults.subtensor._mock = os.getenv('BT_SUBTENSOR_MOCK') if os.getenv('BT_SUBTENSOR_MOCK') != None else False
 
+        defaults.subtensor.register = bittensor.Config()
+        defaults.subtensor.register.num_processes = os.getenv('BT_SUBTENSOR_REGISTER_NUM_PROCESSES') if os.getenv('BT_SUBTENSOR_REGISTER_NUM_PROCESSES') != None else None # uses processor count by default within the function
```
isabella618033, line 204: isnt the default from the OS? 
if so we can remove the non check in the registration etc

camfairchild: if we pass None as an arg to the register function, it doesn't use a default arg, it gets None as the arg.
So, I do a None check and default to the number of CPUs the OS allows for the process.
I figure it's better to leave this logic (using OS) in the cpu registration as the flag will be used for cuda too, potentially

isabella618033 (approved):

PR #867, opened 2022-08-04
Fix/move overview args to cli

camfairchild:
- Moves overview-specific args to cli parsers
- Fixes overview when using `--hotkeys` or `--wallet.hotkeys`
- modifies (un)stake multiple hotkeys output and adds a new function for each

PR #877, opened 2022-08-12
[hotfix] fix flags for run command, fix hotkeys flag for overview, and [feature] CUDA reg

camfairchild:
#876 Fixes the flags 

The CPU register optimization introduced a bug where the btcli run command lacked some config flags that were used in CLI.register.
This PR adds them in

Edit:

This PR also adds
#868 and #867


camfairchild:
I can't make the merge myself. master has to be merged back into nobunaga first

shibshib (approved):

camfairchild:
Wait for merge of #878

camfairchild:
Wait for #868 to merge in

bittensor/_cli/__init__.py
```diff
@@ -819,6 +829,28 @@ def check_register_config( config: 'bittensor.Config' ):
             hotkey = Prompt.ask("Enter hotkey name", default = bittensor.defaults.wallet.hotkey)
             config.wallet.hotkey = str(hotkey)
 
+        if not config.no_prompt and config.subtensor.register.cuda.use_cuda == bittensor.defaults.subtensor.register.cuda.use_cuda:
+            # Ask about cuda registration only if a CUDA device is available.
+            if torch.cuda.is_available():
+                cuda = Confirm.ask("Would you like to try CUDA registration?\n")
```
isabella618033: Sounds like a promotion for cuda registration XDD 
maybe just "Detected CUDA device, use CUDA for registration?" sth like that XD 

Or, do we have functionalities where if we see that a CUDA device is already being used, we avoid assigning the registration task for the same device?

bittensor/_cli/__init__.py
```diff
@@ -819,6 +829,28 @@ def check_register_config( config: 'bittensor.Config' ):
...
+                config.subtensor.register.cuda.use_cuda = cuda
+                # Only ask about which CUDA device if the user has more than one CUDA device.
+                if cuda and config.subtensor.register.cuda.dev_id == bittensor.defaults.subtensor.register.cuda.dev_id and torch.cuda.device_count() > 0:
```
isabella618033: Do not understand the 2nd check, so the dev_id in the config has to be == dev_id in default? 
Then how can we change the dev_id through config?

bittensor/_cli/__init__.py
```diff
@@ -819,6 +829,28 @@ def check_register_config( config: 'bittensor.Config' ):
...
+                    devices: List[str] = [str(x) for x in range(torch.cuda.device_count())]
+                    device_names: List[str] = [torch.cuda.get_device_name(x) for x in range(torch.cuda.device_count())]
+                    console.print("Available CUDA devices:")
+                    choices_str: str = ""
+                    for i, device in enumerate(devices):
+                        choices_str += ("  {}: {}\n".format(device, device_names[i]))
+                    console.print(choices_str)
+                    dev_id = Prompt.ask("Which GPU would you like to use?", choices=devices, default=str(bittensor.defaults.subtensor.register.cuda.dev_id))
```
isabella618033, line 846: Maybe enclose the whole "cuda registration? " -> "which cuda device to use?" thing in a while loop, so that the user dont have to start again when he choose the wrong device.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -488,52 +511,62 @@ def register (
...
-                                    'coldkey': wallet.coldkeypub.ss58_address
-                                } 
-                            )
-                            extrinsic = substrate.create_signed_extrinsic( call = call, keypair = wallet.hotkey )
-                            response = substrate.submit_extrinsic( extrinsic, wait_for_inclusion=wait_for_inclusion, wait_for_finalization=wait_for_finalization )
-                            
-                            # We only wait here if we expect finalization.
-                            if not wait_for_finalization and not wait_for_inclusion:
-                                bittensor.__console__.print(":white_heavy_check_mark: [green]Sent[/green]")
+                    if pow_result['block_number'] < self.get_current_block() - 3:
```
isabella618033: This part was removed from a while loop, any reason why? 
As I rmb, we had the while loop previously for the case where sometime we fail at submitting the extrinsic to subtensor, while the pow result was still valid, not sure if this is still the case.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -729,6 +762,197 @@ def add_stake(
...
+                    continue
+
+                neurons.append( neuron )
+
+        # Remove existential balance to keep key alive.
+        ## Keys must maintain a balance of at least 1000 rao to stay alive.
+        total_staking_rao = sum([amount.rao if amount is not None else 0 for amount in amounts])
+        if total_staking_rao == 0:
+            # Staking all to the first wallet.
+            if old_balance.rao > 1000:
```
isabella618033, line 839: Not sure if I have mistaken it for this if and elif check , but any reason why we need special handling when rao > or < 1000? 

and I am confused with the `# Staking all to the first wallet.` comment

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -729,6 +762,197 @@ def add_stake(
...
+                old_balance -= bittensor.Balance.from_rao(1000)
+
+        elif total_staking_rao < 1000:
+            # Staking less than 1000 rao to the wallets.
+            pass
+        else:
+            # Staking more than 1000 rao to the wallets.
+            ## Reduce the amount to stake to each wallet to keep the balance above 1000 rao.
+            percent_reduction = 1 - (1000 / total_staking_rao)
```
isabella618033, line 848: Correct me if I am wrong, but I suppose `keep the balance above 1000 rao` is for some check to see if some wallet have some existential issue. 
So what we should actually do is to check if the total_staking_rao left the user no more then 1000 rao left, if so, ask the user to input a new amount value, instead of doing all these stuff on the bkg without the user's knowledge.
( although I understand that 1000 rao is a small amount and kind of neglectable =="

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -729,6 +762,197 @@ def add_stake(
...
+            wallet._coldkey = wallet_0._coldkey
+            staking_all = False
+            # Convert to bittensor.Balance
+            if amount == None:
+                # Stake it all.
+                staking_balance = bittensor.Balance.from_tao( old_balance.tao )
+                staking_all = True
+
+            elif not isinstance(amount, bittensor.Balance ):
+                staking_balance = bittensor.Balance.from_tao( amount )
```
isabella618033: amount could be a list, right? but i dont think bittensor.Balance.from_tao handles a list

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -729,6 +762,197 @@ def add_stake(
...
+                    call = substrate.compose_call(
+                    call_module='SubtensorModule', 
+                    call_function='add_stake',
+                    call_params={
+                        'hotkey': wallet.hotkey.ss58_address,
+                        'ammount_staked': staking_balance.rao
+                        }
+                    )
+                    payment_info = substrate.get_payment_info(call = call, keypair = wallet.coldkey)
+                    if payment_info:
```
isabella618033, line 888: Like this check! Do we have this check for the add_stake as well?

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -729,6 +762,197 @@ def add_stake(
         
         return False
 
+    def add_stake_multiple (
```
isabella618033, line 767: Seems to have some overlapping function with add_stake, maybe we can generalise them tgt later...

bittensor/utils/__init__.py
```diff
@@ -405,31 +418,139 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
+            The number of nonces to try before checking for more blocks
+        TPB: int
+            The number of threads per block. CUDA param that should match the GPU capability
+        dev_id: int
+            The CUDA device ID to execute the registration on
+    """
+    if not torch.cuda.is_available():
+        raise Exception("CUDA not available")
+
+    if update_interval is None:
```
isabella618033, line 489: update_interval default is 50000 already

bittensor/utils/__init__.py
```diff
@@ -405,31 +418,139 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
+        update_interval = 50_000
+        
+    block_number = subtensor.get_current_block()
+    difficulty = subtensor.difficulty
+    block_hash = subtensor.substrate.get_block_hash( block_number )
+    while block_hash == None:
```
isabella618033: should add some time.sleep() here etc, 
cause if the network is down and subtensor keep failing, then you will end up with busy waiting
possibly add some exit requirement as well?

setup.py
```diff
@@ -69,4 +69,7 @@
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
     python_requires='>=3.7',
+    extras_requires={
+        'cubit': ['cubit>=1.0.5 @ git+https://github.com/opentensor/cubit.git']
```
isabella618033, line 73: make sure cubit.git is open to public:)

bittensor/_cli/__init__.py
camfairchild: if the dev_id is the default, it'll ask for a CLI input of the dev_id

bittensor/_cli/__init__.py
camfairchild: added

bittensor/_cli/__init__.py
camfairchild: the Prompt package takes care of that actually. I think the error catch isn't necessary

bittensor/_subtensor/subtensor_impl.py
camfairchild: this was in the add_stake already. Something to do with a minimum balance for the chain to keep the coldkey account

camfairchild: > and I am confused with the `# Staking all to the first wallet.` comment  

This is from the if check, if the total staking amount from the args is 0, we assume all balance will be staked.
And this has to go to the first wallet because not every wallet can get all the stake.

bittensor/utils/__init__.py
camfairchild: if we pass None to the arg, python doesn't use the default, it's only if we don't specify the arg at all
So this is necessary unless I do a different call or pass a kwargs to the function call

bittensor/_subtensor/subtensor_impl.py
camfairchild: Yea, it was like that before, we can potentially overload the single stake or call the multiple stake from the single

bittensor/_subtensor/subtensor_impl.py
camfairchild: it's in the for-loop so it can't be. must be a float or a balance

bittensor/_cli/__init__.py
isabella618033: it is not a big issue, just that the flow is a bit weird where in the user's point of view, 
if they have set the default, and they have specified the dev_id in the command line, lt means that they have to confirm again for the 3rd time here XD

bittensor/_subtensor/subtensor_impl.py
camfairchild: I actually don't think this check is necessary, anyway

bittensor/_subtensor/subtensor_impl.py
camfairchild: yes, it's already there

bittensor/_subtensor/subtensor_impl.py
isabella618033: I think the user should specify where should the money go instead of having us to make assumption that the money is going to the first wallet (and how do you define the first wallet?)

bittensor/_subtensor/subtensor_impl.py
isabella618033: ohh right right

bittensor/utils/__init__.py
isabella618033: yea, but in what case should we pass none to the arg? Shouldnt we just not specify the arg?

bittensor/_subtensor/subtensor_impl.py
camfairchild: I'll change that back

isabella618033 (approved):

bittensor/utils/__init__.py
camfairchild: added a backoff to grab the new block

bittensor/utils/__init__.py
camfairchild: the None is the default of the config flag. The CLI will always pass None if the user doesn't specify that flag.

bittensor/_cli/__init__.py
camfairchild: true, I'll fix it

PR #878, opened 2022-08-14
Fix/diff unpack bit shift

camfairchild:
Turns out the bit shift to unpack the difficulty was also wrong
- Fixed the bit shift
- Added a test to catch this
- Added a process restriction on the registration tests so they finish quicker

### DIFF
--- a/bittensor/utils/__init__.py
+++ b/bittensor/utils/__init__.py
@@ -269,6 +269,14 @@ def solve_for_nonce_block(solver: Solver, nonce_start: int, nonce_end: int, bloc
     solver.best_queue.put((best_local, best_seal_local))
     return None, time.time() - start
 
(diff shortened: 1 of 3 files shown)


shibshib (approved):

bittensor/utils/__init__.py
```diff
@@ -220,7 +220,7 @@ def run(self):
                 with self.check_block:
                     block_number = self.curr_block_num.value
                     block_bytes = bytes(self.curr_block)
-                    block_difficulty = int(self.curr_diff[0] >> 32 | self.curr_diff[1])
+                    block_difficulty = int(self.curr_diff[0] << 32 | self.curr_diff[1])
```
eduardogr: is this covered with some unit tests??
Seems like it's the place to catch these details

eduardogr: From my point of view, when a piece of code is that critical, it's worthy to extract into a new function a cover it with unit tests. In this way, integration tests could avoid knowing about its domain.

camfairchild: the integration test covers it. Perhaps I should move the test to the unit test folder as it's more of a unit test.

I'll also refactor it into two functions unpack/pack

PR #879, opened 2022-08-15
Add/address CUDA reg changes

camfairchild:
This PR addresses the comments from #877


tests/unit_tests/bittensor_tests/utils/test_utils.py
```diff
@@ -235,17 +235,29 @@ def test_is_valid_ed25519_pubkey():
...
+    
+    mock_diff = multiprocessing.Array('Q', [0, 0], lock=True) # [high, low]
+    
+    bittensor.utils.registration_diff_pack(fake_diff, mock_diff)
+    assert bittensor.utils.registration_diff_unpack(mock_diff) == fake_diff
+
+    fake_diff = pow(2, 32) * pow(2, 4) # this should be too large if the bit shift is wrong (32 + 4 bits)
+    
+    bittensor.utils.registration_diff_pack(fake_diff, mock_diff)
+    assert bittensor.utils.registration_diff_unpack(mock_diff) == fake_diff
```
eduardogr: These 2 asserts covers the same scenario? or are they from different scenarios??

From xUnit principles, it's a good practice to keep assertions group by scenario/usecase, so, if one group of aserts fail we must know what's the failure about (a group could be of one in this example).

If they are identifying different cases, i would split them in two test functions.

bittensor/utils/__init__.py
```diff
@@ -449,6 +450,18 @@ def millify(n: int):
...
+                            Exception,
+                            interval=1,
+                            max_tries=3)
+def get_block_with_retry(subtensor: 'bittensor.Subtensor') -> Tuple[int, int, bytes]:
+    block_number = subtensor.get_current_block()
+    difficulty = subtensor.difficulty
+    block_hash = subtensor.substrate.get_block_hash( block_number )
+    if block_hash is None:
+        raise Exception("Network error. Could not connect to substrate to get block hash")
+    return block_number, difficulty, block_hash
```
eduardogr, line 469: I am missing the happy case test when we have no exception. Do we have it?

eduardogr: btw, good movement extracting this.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -510,62 +510,64 @@ def register (
...
-                                bittensor.__console__.print(":white_heavy_check_mark: [green]Already Registered[/green]")
-                                return True
-
-                            bittensor.__console__.print(":cross_mark: [red]Failed[/red]: error:{}".format(response.error_message))
-                            time.sleep(1)
-                        
-                        # Successful registration, final check for neuron and pubkey
+                    with bittensor.__console__.status(":satellite: Submitting POW..."):
+                        # check if pow result is still valid
+                        while pow_result['block_number'] >= self.get_current_block() - 3:
```
eduardogr: I can see this change of condition is correct.

Just a comment around this kind of changes:
- In terms of maintain the code it would be helpful to have the check of this condition encapsulated in a single function and also provide a unit test with a data provider with the set of inputs that lead us to have true/false. That would lead us to have a change of that test along with this condition change that would be helpful in terms of maintainability and also to read PRs around.

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -865,10 +867,9 @@ def add_stake_multiple (
                 # Stake it all.
                 staking_balance = bittensor.Balance.from_tao( old_balance.tao )
                 staking_all = True
-
-            elif not isinstance(amount, bittensor.Balance ):
-                staking_balance = bittensor.Balance.from_tao( amount )
```
eduardogr, line 870: don't we need this anymore?

eduardogr (approved):

tests/unit_tests/bittensor_tests/utils/test_utils.py
camfairchild: Good point. I'll split these. One tests less than 32 bits, the other more than 32 bits

bittensor/_subtensor/subtensor_impl.py
camfairchild: extracted and wrote up some unit tests

bittensor/_subtensor/subtensor_impl.py
camfairchild: We cast above to Balance so they are all Balances

bittensor/utils/__init__.py
camfairchild: added some success tests for this

PR #882, opened 2022-08-17
Fixed GPU registration bug

MythicFrost:
The block was indented one tab too far.


camfairchild:
Thanks for the PR. I added a test and opened #883

PR #884, opened 2022-08-18
[Fix] --help command

camfairchild:
This PR:
- fixes the `btcli --help` output
- adds a test to verify the help output is correct
- modifies the help output with the version

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -227,13 +227,13 @@ class axon:
     def add_defaults(cls, defaults):
         """ Adds parser defaults to object from enviroment variables.
         """
-        defaults.axon = bittensor.config()
+        defaults.axon = bittensor.Config()
         defaults.axon.port = os.getenv('BT_AXON_PORT') if os.getenv('BT_AXON_PORT') != None else 8091
         defaults.axon.ip = os.getenv('BT_AXON_IP') if os.getenv('BT_AXON_IP') != None else '[::]'
         defaults.axon.max_workers = os.getenv('BT_AXON_MAX_WORERS') if os.getenv('BT_AXON_MAX_WORERS') != None else 10
         defaults.axon.maximum_concurrent_rpcs = os.getenv('BT_AXON_MAXIMUM_CONCURRENT_RPCS') if os.getenv('BT_AXON_MAXIMUM_CONCURRENT_RPCS') != None else 400
         
-        defaults.axon.priority = bittensor.config()
+        defaults.axon.priority = bittensor.Config()
         defaults.axon.priority.max_workers = os.getenv('BT_AXON_PRIORITY_MAX_WORKERS') if os.getenv('BT_AXON_PRIORITY_MAX_WORKERS') != None else 10
         defaults.axon.priority.maxsize = os.getenv('BT_AXON_PRIORITY_MAXSIZE') if os.getenv('BT_AXON_PRIORITY_MAXSIZE') != None else -1
 
--- a/bittensor/_config/__init__.py
+++ b/bittensor/_config/__init__.py
@@ -95,8 +95,7 @@ class config:
                 print('Error in loading: {} using default parser settings'.format(e))
 
         # 2. Continue with loading in params.
-        if strict:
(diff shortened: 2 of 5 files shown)


bittensor/_axon/__init__.py
```diff
@@ -227,13 +227,13 @@ def add_args( cls, parser: argparse.ArgumentParser, prefix: str = None  ):
     def add_defaults(cls, defaults):
         """ Adds parser defaults to object from enviroment variables.
         """
-        defaults.axon = bittensor.config()
+        defaults.axon = bittensor.Config()
```
Eugene-hu, line 230: Is there a reason that the axon config needs to be the actual class rather than the factory class?

bittensor/_config/__init__.py
```diff
@@ -37,13 +40,15 @@ class InvalidConfigFile(Exception):
         """ In place of YAMLError
         """
 
-    def __new__( cls, parser: ArgumentParser = None, strict: bool = False ):
+    def __new__( cls, parser: ArgumentParser = None, strict: bool = False, args: Optional[List[str]] = None ):
```
Eugene-hu, line 43: Is there a reason for this change? I will be hesitant to change the base config class unless necessary. Every single bittensor object initiates a base config object. So ideally, we should keep it as lightweight as possible. The config class is just a whole bag of worms XD

Eugene-hu: I also think the config class is in need of a cleanup, so it might be good for the next update.

Eugene-hu (commented):
The implementation looks good, but we just need to be very careful about the config class. Let's run a few tests on nobunaga to make sure nothing breaks :)

bittensor/_axon/__init__.py
camfairchild: this is what every other class does in add_defaults. Basically the config factory instantiates a new argument parser instance.
This messed up the argparser creation and makes btcli --help not work.

 What you're supposed to do is create a new Config object (Munch object) so config.axon works after add_defaults (is a Munch object)

bittensor/_config/__init__.py
camfairchild: this is an inversion of control of the sys.argv from the command line. 

Normally argparser gets it internally, if you don't specify which args to parse, by calling sys.argv the same way I added in the neuron `__main__`

I added it as an optional arg and lower down it defaults to calling sys.argv if None, so the class will function as normal without the arg specified.

The IoC is important for testing if `btcli --help` works correctly and has the text we expect. Otherwise I would have to patch argparser, which is subject to API changes. This is recommended from some SO posts I read.

Edited: dep injection -> IoC

eduardogr: Since this sounds like a temporary change before the cleanup, i would generate some issues in this repository to not forget about the tech debt we have around this code.

eduardogr:
I do not know much this codebase, but from my point of view this change seems simple enough to be able to merge it.
It looks direct and effective with its purpose. 

I understand that changing functions' interfaces is not the best and can bring chaos to the codebase.

I would like that before we close this PR we create issues tagged as **tech-debt** exposing the debt we have related with CLI/Config/Args. Because, from my point of view, seems like the bug was also introduce for the existence of the tech-spec.

Finally, since this is going to the nobunaga branch and with no release deadline I see we can merge this and continue learning and improving the codebase from this point if something fails again. By the way, we should have the necessary test coverage for this, do we have it?

eduardogr (approved):
I do not know much this codebase, but from my point of view this change seems simple enough to be able to merge it.
It looks direct and effective with its purpose.

I understand that changing functions' interfaces is not the best and can bring chaos to the codebase.

I would like that before we close this PR we create issues tagged as tech-debt exposing the debt we have related with CLI/Config/Args. Because, from my point of view, seems like the bug was also introduced for the existence of the tech-spec.

Finally, since this is going to the nobunaga branch and with no release deadline I see we can merge this and continue learning and improving the codebase from this point if something fails again. By the way, we should have the necessary test coverage for this, do we have it?

unconst (approved):

PR #886, opened 2022-08-22
Bit 553 bug fixes

isabella618033:
- axon priority_threadpool not passed through
- remove server's error message on validator side
- server set weight block time to be independent with optimisation step
- axon wasnt timing out at 12 sec
-  responsive uid not counted properly

### DIFF
--- a/bittensor/_axon/axon_impl.py
+++ b/bittensor/_axon/axon_impl.py
@@ -201,7 +201,7 @@ class Axon( bittensor.grpc.BittensorServicer ):
         # ==============================================================
         # ==== Function which prints all log statements per synapse ====
         # ==============================================================
-        def finalize_codes_stats_and_logs():
+        def finalize_codes_stats_and_logs( message = None):
             for index, synapse in enumerate( synapses ):
                 request.synapses [ index ].return_code = synapse_codes[ index ] # Set synapse wire proto codes.
                 request.synapses [ index ].message = synapse_messages[ index ] # Set synapse wire proto message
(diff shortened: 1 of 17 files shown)


bittensor/_threadpool/priority_thread_pool_impl.py
```diff
@@ -190,6 +193,7 @@ def weakref_cb(_, q=self._work_queue):
         if num_threads < self._max_workers:
             thread_name = '%s_%d' % (self._thread_name_prefix or self,
                                      num_threads)
+            print('thread_name', self._max_workers, thread_name)
```
Eugene-hu: Might want to remove this ;)

bittensor/_threadpool/priority_thread_pool_impl.py
```diff
@@ -166,11 +166,14 @@ def submit(self, fn, *args, **kwargs):
             start_time = time.time()
             if 'priority' in kwargs:
                 del kwargs['priority']
+            
+            timeout = kwargs.get('timeout', None)
+            if 'timeout' in kwargs:
```
Eugene-hu: Can you add tests for this individually? There should be a test_priority_thread_pool.py

bittensor/_axon/axon_impl.py
```diff
@@ -280,7 +281,8 @@ def finalize_codes_stats_and_logs():
                     inputs_x = deserialized_forward_tensors, 
                     synapses = synapses,
                     priority = priority,
-                    hotkey= request.hotkey
+                    hotkey = request.hotkey,
+                    timeout = self.forward_timeout - (clock.time() - start_time)
```
Eugene-hu: One thing here; we will need a higher timeout for seq2seq requests. So we need to figure out which synapse type and set the corresponding timeout

Eugene-hu (commented):
Looks good! Left a few comments that we need to be careful about

bittensor/_axon/axon_impl.py
isabella618033: i think one of the issue is that it is possible for the validator (or client) to send through multiple synapse at the same time, and the way we handle it is to send multiple requests back at one time
so the timeout should be the min of the timeouts if there is multiple synapses, what do you think?

Eugene-hu: yeap, that sounds good

Eugene-hu (approved):
LGTM!

PR #887, opened 2022-08-22
[feature] external axon flags

camfairchild:
This PR adds two flags: `--axon.external_ip` and `--axon.external_port`  

The flags are used to specify to the chain axon networking info that doesn't match your internal networking, or can't be found automatically. 

For instance, this is useful for both runpod.io and vast.ai servers which configure the internal container ports to usually mismatch with a mapped external port. And occasionally a proxied IP address.   

These flags make it much easier to run a bittensor server on their platforms.


bittensor/_axon/__init__.py
```diff
@@ -199,9 +210,13 @@ def add_args( cls, parser: argparse.ArgumentParser, prefix: str = None  ):
         prefix_str = '' if prefix == None else prefix + '.'
         try:
             parser.add_argument('--' + prefix_str + 'axon.port', type=int, 
-                    help='''The port this axon endpoint is served on. i.e. 8091''', default = bittensor.defaults.axon.port)
+                    help='''The local port this axon endpoint is binds. i.e. 8091''', default = bittensor.defaults.axon.port)
```
unconst: Is bound to

bittensor/_axon/axon_impl.py
```diff
@@ -2,6 +2,7 @@
 """
 # The MIT License (MIT)
 # Copyright © 2021 Yuma Rao
+# Copyright © 2022 Opentensor Foundation
```
unconst, line 5: Is this a thing now?

unconst (approved):

bittensor/_axon/axon_impl.py
camfairchild: I feel like we should, no?

PR #888, opened 2022-08-26
normalize_max_multiple update

Eugene-hu:
Previously, the normalization incorrectly assumes that the inflection point of the tanh (x=0) normalization occurs at the beginning of the data.  This incorrect assumption led to more weight being placed on the lower end of the distribution.

image

This update corrects this assumption by shifting the inflection point to a specific quantile of the weight distribution. This correctly shifts the distribution and makes use of the full range of the tanh normalization. The update should allow validators to set higher number of weights without worry that the top performers will be diluted.

image


Eugene-hu:
The issue that this pr was trying to fixed was solved by another fix.

PR #889, opened 2022-08-29
[BIT-552] Validator improvements (nucleus permute, synergy avg)

opentaco:
### [BIT-552] Validation improvements

Minor validator fixes for rich table display delimitation, counting responsive UID when first encountered, and skipping repeat metagraph syncs.

Major validator improvements removing factors of variability:

1. Nucleus permutes `metagraph.n` (e.g. 4096) and consumes subsets from order to ensure each UID is queried once.
1. Neuron prolongs epoch until e.g. 4096 unique UIDs are queried once, at least.
1. Synergy is now averaged to ensure consistency across different subset (nucleus.topk) sizes.
1. `synergy_scaling_law_power` parameter is introduced.
1. `synergy_scaling_law_power` is increased from 0.5 to 0.6 to compensate for averaging producing much lower synergy values compared to base Shapley values.

[BIT-552]: https://opentensor.atlassian.net/browse/BIT-552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -543,6 +563,7 @@ def neuron_stats_update(self, neuron_stats: Dict[int, Dict[str, Any]]):
                 zkey = key + '!'  # zeroing key
                 stats.setdefault(zkey, 0.)  # initialize zkey val to zero to gradually increase with observations
                 if key in _stats and not math.isnan(_stats[key]):
+                    responsive_uids += [_uid]
```
Eugene-hu, line 566: Does moving this line up change the set of responsive uids?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -794,18 +816,26 @@ def forward(
...
-        request_start_time = time.time()
+        # === Ensure each UID is queried once ===
+        # Persist object variable self.permute_uids across forward calls.
+        # Reset to new permutation of all UIDs once empty.
+        if len(self.permute_uids) == 0:  # no more UIDs to query
+            self.permute_uids = torch.randperm(metagraph.n)  # reset to new permutation of all UIDs
 
         # === Randomly select num_endpoints UIDs ===
-        random_uids = torch.randperm(metagraph.n)[:num_endpoints]
+        random_uids = self.permute_uids[:num_endpoints]  # newest selection of UIDs to query
```
Eugene-hu, line 826: self.permute_uids is a torch tensor, correct? If len(permute_uids) < num_endpoints, it wouldn't break?

Eugene-hu (approved):
LGTM!

isabella618033 (approved):

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: i think prev set of responsive uid was not complete, and this change makes it complete

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, it will now include a responsive UID also on first encounter, where previously it required that a UID already existed in `neuron_stats`. So this fixes it.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: If `len(permute_uids) < num_endpoints` with a torch tensor it should only provide the available slice, same as with a normal list slice index that exceeds actual len. This is tested to not break e.g. during final 16 UIDs in 4096 indexed in subsets of 20, where it only slices the 16 available UIDs.

unconst (approved):
Beautiful.

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -1229,6 +1267,7 @@ def shapley_synergy(stats: Dict, synergy: Callable, ext: str, target: torch.Tens
                 measured_loss = synergy(first, second, target, ext)
 
                 loss_diff_share = torch.clamp(expected_loss - measured_loss, 0) / 2  # record direct loss diff
+                loss_diff_share /= len(responsives)  # average over responsives
```
shibshib, line 1270: Are we 100% certain the `responsives` won't ever be 0?

shibshib (approved):

shibshib:
Great PR @opentaco , should we merge this to Nobunaga if we're testing on Nobunaga testnet?

bittensor/_neuron/text/core_validator/__init__.py
opentaco: `len(responsives)` can be 0, but then it won't enter the loop and the division by zero won't occur.
Note how the responsives list comprehension is mirroring the loop below, so if responsives is empty the loop won't be entered.

```python
    responsives = [uid for uid, stat in stats.items() if 'loss' + ext in stat]
    for _first, first in stats.items():
        if 'loss' + ext not in first:
            continue
```

PR #900, opened 2022-09-02
fix help overview, fix tests, change to --config.strict and --config

unconst:
Fixes --help for the btcli

The issue was caused because by bittensor calls a parse_args on import. 
The fix was ensuring bittensor only calls pars_args when a parser has been created by the API.

Features:
-- changes --config to --config.file
-- changes --strict to --config.strict

Updates: 
This will require a documentation change to reflect these new arguments


camfairchild:
Is `--help` not working on nobunaga still?

 #884 should have fixed this

PR #903, opened 2022-09-05
Adds no_private command to cli and wallet

unconst:
User can now specify that they only want to regenerate the coldkeypub.txt file during regeneration or on the creation of the coldkey. This helps the user keep private information off of potentially insecure servers.

btcli regen_coldkey --no_private

btcli new_coldkey --no_private

### DIFF
--- a/bittensor/_cli/__init__.py
+++ b/bittensor/_cli/__init__.py
@@ -324,6 +324,13 @@ class cli:
             help='Seed hex string used to regen your key i.e. 0x1234...'
         )
         regen_coldkey_parser.add_argument(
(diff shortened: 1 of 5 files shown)


camfairchild (changes requested):
`btcli regen_coldkeypub` does this already
see #831

Further, you should only be able to do this securely if you never have to provide the private portion. Meaning using the address/public key to regen.

camfairchild:
We can probably incorporate this into the regular btcli regen command, but not new coldkey. That still exposes the insecure server  to the private part

PR #904, opened 2022-09-05
adds --n_keys to new_hotkey command

unconst:
Allows the user to specify --n_keys during the new_hotkey command allowing them to create multiple hotkeys at once. 

Example:
btcli new_hotkey --n_keys 10                       
Enter wallet name (default): 
Enter hotkey name (default): 
Create hotkeys: '['default0', 'default1', 'default2', 'default3', 'default4', 'default5', 'default6', 'default7', 'default8', 'default9']'? [y/n]:

### DIFF
--- a/bittensor/_cli/__init__.py
+++ b/bittensor/_cli/__init__.py
@@ -310,7 +310,7 @@ class cli:
             help='''Creates a new hotkey (for running a miner) under the specified path.'''
         )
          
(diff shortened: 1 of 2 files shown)


camfairchild (changes requested):
why not allow user input of the names? Something like, the name flag is an alias of the names flag, with nargs=+
Then only ask for the name if the array is empty, otherwise make one hotkey per name in the array

PR #905, opened 2022-09-06
[fix] fixes unstake with max-stake flag

camfairchild:
The balance class equality didn't support `Balance == None`, this PR adds a catch for this and returns False.
This fixes an error on `unstake_multiple` with a `wallet.max_stake` flag set

### DIFF
--- a/bittensor/utils/balance.py
+++ b/bittensor/utils/balance.py
@@ -68,6 +68,9 @@ class Balance:
         return self.__str__()
 
     def __eq__(self, other: Union[int, float, "Balance"]):
+        if other is None:
+            return False
+            
         if hasattr(other, "rao"):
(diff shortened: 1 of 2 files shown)


bittensor/utils/balance.py
```diff
@@ -68,6 +68,9 @@ def __repr__(self):
         return self.__str__()
 
     def __eq__(self, other: Union[int, float, "Balance"]):
+        if other is None:
```
eduardogr, line 71: what about balances that are None??

eduardogr (approved):

bittensor/utils/balance.py
camfairchild: Balances can't be None, as they then aren't an instance of Balance.

PR #910, opened 2022-09-08
Update old docs link to new one

quac88:
This change deletes the old gitbooks documentation link and replaces it with the new one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ### Internet-scale Neural Networks <!-- omit in toc -->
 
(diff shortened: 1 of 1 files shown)


camfairchild (changes requested):
Can you modify the network link to be `https://bittensor.com/network` as well

camfairchild:
Also, we should merge this into nobunaga branch.

camfairchild:
#915 implements the proposed changes and is based on nobunaga instead

PR #912, opened 2022-09-12
Parachain registration

shibshib:
This enables the parachain registration and mining on Bellagene Parachain testnet. Should be possible to simply run

```
btcli run --subtensor.network bellagene --subtensor.chain_endpoint "wss://parachain.opentensor.ai:443" --wallet.name bellagene_mac
```

This PR also includes some fixes to the cpu registration optimization (namely, we were using APIs that are not available on all linux distress or MacOS).


bittensor/_subtensor/__init__.py
```diff
@@ -232,6 +236,9 @@ def determine_chain_endpoint(network: str):
         elif network == "nobunaga": 
             # Staging network.
             return bittensor.__nobunaga_entrypoints__[0]
+        elif network == "bellagene":
+            # Parachain test net
+            return bittensor.__bellagene_entrypoints__[0]
```
Eugene-hu: Do we still use multiple entrypoints? I think we can probably depreciate that structure

Eugene-hu (approved):
Looks good! Does this need to go straight to master? If not, it should go Nobunaga, and be released in the next version

bittensor/_subtensor/__init__.py
shibshib: I think so too. let me fix it.

isabella618033 (approved):

PR #923, opened 2022-09-21
[BIT 584] [feature] btcli register output stats not in place

camfairchild:
Previously the registration statistics were always output in-place, overwriting the old information.  
To help with logging performance, this PR adds a flag `--subtensor.register.output_in_place <bool>`  

Setting this flag to false (default true) makes the logging send output to new lines instead.   

Screen Shot 2022-09-21 at 12 17 34 PM  

This PR also addresses an issue with the flag `--subtensor.register.cuda.use_cuda <bool>` and `--wallet.reregister <bool>` where they were parsed as `bool(str)` instead of `strtobool(str)`, meaning `"" == False` and `"False" == True` and, importantly `--cuda.use_cuda False` was parsed as `True`


shibshib (approved):

bittensor/_wallet/__init__.py
```diff
@@ -114,7 +115,7 @@ def add_args(cls, parser: argparse.ArgumentParser, prefix: str = None ):
             
             parser.add_argument('--' + prefix_str + 'wallet.hotkeys', '--' + prefix_str + 'wallet.exclude_hotkeys', required=False, action='store', default=bittensor.defaults.wallet.hotkeys, type=str, nargs='*', help='''Specify the hotkeys by name. (e.g. hk1 hk2 hk3)''')
             parser.add_argument('--' + prefix_str + 'wallet.all_hotkeys', required=False, action='store_true', default=bittensor.defaults.wallet.all_hotkeys, help='''To specify all hotkeys. Specifying hotkeys will exclude them from this all.''')
-            parser.add_argument('--' + prefix_str + 'wallet.reregister', required=False, action='store', default=bittensor.defaults.wallet.reregister, type=bool, help='''Whether to reregister the wallet if it is not already registered.''')
+            parser.add_argument('--' + prefix_str + 'wallet.reregister', required=False, action='store', default=bittensor.defaults.wallet.reregister, type=strtobool, help='''Whether to reregister the wallet if it is not already registered.''')
```
isabella618033, line 119: Why do we need it to be strtobool instead of bool?

bittensor/_cli/__init__.py
```diff
@@ -832,32 +832,38 @@ def check_overview_config( config: 'bittensor.Config' ):
...
-                    try:
-                        # replace the commas with spaces then split over whitespace.,
-                        # then strip the whitespace and convert to ints.
-                        dev_id = [int(dev_id.strip()) for dev_id in dev_id.replace(',', ' ').split()]
-                    except ValueError:
-                        console.error(":cross_mark:[red]Invalid GPU device[/red] [bold white]{}[/bold white]\nAvailable CUDA devices:{}".format(dev_id, choices_str))
-                        sys.exit(1)
-                config.subtensor.register.cuda.dev_id = dev_id
+            if not config.no_prompt:
+                if config.subtensor.register.cuda.get('use_cuda', 'NA') == 'NA': # flag not set
```
isabella618033: Why do we need it as a 'NA' instead of a None?

isabella618033 (approved):
LGTM, left you some questions.

bittensor/_wallet/__init__.py
camfairchild: if it's just bool, it'll cast the string input to a bool literally. So we get `bool("False") == True`

strtobool matches the word in the string with the wanted bool

bittensor/_cli/__init__.py
camfairchild: I'll have to check, but this can probably be `None`

bittensor/_wallet/__init__.py
isabella618033: can we change the action to 'store_true' or 'store_false' instead? Just like the rest if it.

camfairchild: those only allow for true or false. The idea here is to allow `--wallet.reregister False` and `--wallet.reregister True`

unconst (approved):

PR #930, opened 2022-09-29
changed local train and remote train descriptions

quac88:
local and remote train sometimes cause bugs if set to false. The default is already false, so an extra note is added to remind users to only use this flag when passing true


camfairchild:
I think the description should be more like "Set this flag to enable [...]".

Good PR, though! Definitely needed.

camfairchild:
Also, make sure to merge PRs into nobunaga instead of master

shibshib (approved):

unconst (approved):

shibshib:
@quac88 I just noticed this is effectively trying to merge your repository to this repository :D 

We can't merge this since this is your personal repository. You need to do a git checkout on the opentensor foundation's repo, and then make a new branch with these change and push it up.

camfairchild:
imo this should be merged into nobunaga first so it doesn't mess with our branching

PR #933, opened 2022-09-29
[feature] [BIT 578] speed up metagraph storage query

camfairchild:
This PR speeds up the storage call made by`subtensor.neurons` in the `subtensor.use_neurons_fast` function.  

This feature works by bundling a nodejs binary with the [polkadotjs API](https://github.com/polkadot-js/api).  
This binary is a CLI that implements the `sync_and_save --filename <default:~/.bittensor/metagraph.json> --block_hash <default:latest>` command.   
This syncs the metagraph at the blockhash and saves it to a json file.  

The speed-up is quite significant, below is a test run of the manual without the fix, with the ipfs cache, and with the fix.  
output

And below is the IPFS cache sync versus the manual sync (with fix)

output_cach_vs_fixed_manual

A pro of this is that it removes the need for a centralized IPFS cache of the metagraph.  

A downside of this fix is that the binaries with nodejs bundled use ~50MB each (one linux, one macos).  
There is currently no binary for windows, but I'm not certain this should be included anyway, as we only support linux/macos.  

 
Another pro of this fix is it works on both nobunaga and nakamoto, and can be adapted to any network.
This also leaves room for adding other large substrate queries and working further with the polkadot js api.


setup.py
```diff
@@ -36,6 +38,25 @@
...
+        # remove linux binaries
+        os.remove('bin/subtensor-node-api-linux')
+    else:
+        # neither linux or macos
+        # remove both binaries
+        os.remove('bin/subtensor-node-api-linux')
+        os.remove('bin/subtensor-node-api-macos')
+except Exception as e:
+    print('Failed to remove unsupported subtensor node api binaries. Error: {}'.format(e))
+    pass
```
eduardogr: Why do we need to remove binaries?

I agree with the intention of providing just one binary. What about just packaging one binary conditionally and not removing the unused? Would that be possible?

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1500,6 +1513,91 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+                "priority": str(int),
+                "last_update": int,
+                "weights": [
+                    [int, int],
+                ],
+                "bonds": [
+                    [int, str(int)],
+                ],
+            }
+            """
```
eduardogr: This is helpful given that you are including an external binary.

What about moving this specification to a document in the bin file and translate this comment also to a set of unit tests?

eduardogr: Both documents; the specification and the tests could be referenced here with links

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1500,6 +1513,91 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+            if platform == "linux" or platform == "linux2":
+                # linux
+                os_name = "linux"
+            elif platform == "darwin":
+                # OS X
+                os_name = 'macos'
+            else:
+                raise Exception('Unsupported platform for subtensor-node-api')
+            bittensor.__console__.print("Using subtensor-node-api for neuron retrieval...")
+            path_to_bin = os.path.join(os.path.dirname(__file__), f"../../bin/subtensor-node-api-{os_name}")
```
eduardogr: Since this is something independent and has importance here I would say that it would be a good idea to extract it to an independent function and test it with unit tests

bittensor/_subtensor/subtensor_impl.py
eduardogr: Adding unit tests for the binary file is good to:
- localize files and ensure that you have all of them in the repo
- write a specification

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1500,6 +1513,91 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+            # will write to ~/.bittensor/metagraph.json by default
+            subprocess.run([path_to_bin, "sync_and_save", "-u", endpoint_url, '-b', block_hash], check=True, stdout=subprocess.PIPE)
+            with open(os.path.join(os.path.expanduser('~'), '.bittensor/metagraph.json')) as f:
+                data = json.load(f)
```
eduardogr: What about errors here? for ex:

- errors executing the binary
- possible expected errors of the execution
- errors reading the file

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1500,6 +1513,91 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+        """
+        neurons = []
+        if block is None:
+            block: int = self.get_current_block()
+
+        block_hash: str = self.substrate.get_block_hash( block )
+        endpoint_url: str = self.chain_endpoint
+        # format endpoint with ws://
+        if "ws://" not in endpoint_url and "wss://" not in endpoint_url:
+            endpoint_url = "ws://" + endpoint_url
```
eduardogr: nice candidate to extract to a function and test it with unit tests to ensure a well-formed endpoint for wss

eduardogr:
Very nice results @camfairchild  !

where do binaries come from? It would be nice to have information about that and visibility about the code that originate that bins, What do you think about that?

setup.py
camfairchild: The only way I see this being possible is if we download the binary on-demand. The idea here was to conserve space because they're 50MB. 
Not sure users can build the binaries themselves as they require node/npm, might be inconvenient.

camfairchild:
> 

The binaries are from https://github.com/opentensor/subtensor-node-api
Not sure where to write docs about this though

bittensor/_subtensor/subtensor_impl.py
camfairchild: good idea, added

setup.py
eduardogr: I see your point.

About this approach, does this means that we are going to maintain one pip package for each platform now?

bittensor/_subtensor/subtensor_impl.py
eduardogr: great!

bittensor/_subtensor/__init__.py
```diff
@@ -134,11 +138,11 @@ def __new__(
         # make sure it's wss:// or ws://
         # If it's bellagene (parachain testnet) then it has to be wss
         endpoint_url: str = config.subtensor.chain_endpoint
-        if endpoint_url[0:6] != "wss://" and endpoint_url[0:5] != "ws://":
-            if config.subtensor.network == "bellagene":
-                endpoint_url = "wss://{}".format(endpoint_url)
-            else:
-                endpoint_url = "ws://{}".format(endpoint_url)
+        if config.subtensor.network == "bellagene":
```
eduardogr, line 143: should `bellagene` be a constant?

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1490,16 +1491,56 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+        neurons = []
+        if block is None:
+            block: int = self.get_current_block()
+
+        block_hash: str = self.substrate.get_block_hash( block )
+        endpoint_url: str = self.chain_endpoint
+        endpoint_url = bittensor.utils.networking.get_formatted_ws_endpoint_url(endpoint_url)
+    
+        # check if fast sync is available
+        FastSync.verify_fast_sync_support()
```
eduardogr, line 1593: Could we name the function in such a way that we do not need the comment?

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1490,16 +1491,56 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+        # try to fast sync
+        fast_sync: FastSync = FastSync(endpoint_url)
```
eduardogr, line 1595: If this is a _try_, should we check some expected error cases? If not, is the comment needed?

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1490,16 +1491,56 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+        # get neurons
+        fast_sync.sync_neurons(block_hash)
+        # load neurons
+        neurons = fast_sync.load_neurons()
```
eduardogr, line 1599: given that comment is the name of the function i would say that we do not need it

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1490,16 +1491,56 @@ def neurons(self, block: int = None ) -> List[SimpleNamespace]:
...
+        block_hash: str = self.substrate.get_block_hash( block )
+        endpoint_url: str = self.chain_endpoint
+        endpoint_url = bittensor.utils.networking.get_formatted_ws_endpoint_url(endpoint_url)
+    
+        # check if fast sync is available
+        FastSync.verify_fast_sync_support()
+        # try to fast sync
+        fast_sync: FastSync = FastSync(endpoint_url)
+        # get neurons
+        fast_sync.sync_neurons(block_hash)
```
eduardogr, line 1597: could we improve the name of the function to get_neurons? or sync_and_get_neurons? Or something that's aligned with its purpose?

Seems like we could improve the name of the function and remove the comment

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1667,3 +1711,7 @@ def neuron_for_wallet( self, wallet: 'bittensor.Wallet', block: int = None ) ->
                 neuron object associated with uid or None if it does not exist.
         """
         return self.neuron_for_pubkey ( wallet.hotkey.ss58_address, block = block )
+
+class NeuronPullException(Exception):
+    """Raised when pullin a Neuron from the chain fails."""
+    pass
```
eduardogr, line 1808: Could we use a Python built-in exception here?

Is python idiomatic to use built-in exceptions.

check out this list: https://docs.python.org/3/library/exceptions.html

Maybe **LookupError** ?

```
exception LookupError

    The base class for the exceptions that are raised when a key or index used on a mapping 
    or sequence is invalid: IndexError, KeyError. This can be raised directly by codecs.lookup().
```

- You can check [codecs.lookup() here](https://docs.python.org/3/library/codecs.html#codecs.lookup)

Or maybe:
- [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError)
- [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError)

setup.py
camfairchild: Iirc, the way it works is we have one wheel per platform under the same package. pip should automatically request and install the correct wheel.

tests/unit_tests/bittensor_tests/utils/test_fast_sync.py
```diff
@@ -0,0 +1,267 @@
...
+                _ = FastSync.verify_os_support()
+
+    def test_os_not_supported_freebsd(self):
+        with patch("bittensor.utils.fast_sync.FastSync.get_platform", return_value="freebsd"): # freebsd is not supported
+            with pytest.raises(FastSyncOSNotSupportedException):
+                _ = FastSync.verify_os_support()
+    
+    def test_os_supported_linux(self):
+        with patch("bittensor.utils.fast_sync.FastSync.get_platform", return_value="linux"): # linux is supported
+            _ = FastSync.verify_os_support()
```
eduardogr: i think you can avoid the result for this kind of tests. But it's nice to point out that the function returns something...mmm... nice anyways!

tests/unit_tests/bittensor_tests/utils/test_fast_sync.py
```diff
@@ -0,0 +1,267 @@
+import json
```
eduardogr, line 1: nice job here @camfairchild , congrats!

bittensor/_subtensor/subtensor_impl.py
camfairchild: I'll modify the comment because it only syncs the neurons from the network. It doesn't return any neurons, you have to load neurons after

bittensor/utils/networking.py
```diff
@@ -199,3 +199,19 @@ def upnpc_create_port_map(port: int):
 
     except Exception as e:
         raise UPNPCException(e) from e
+
+def get_formatted_ws_endpoint_url(endpoint_url: str) -> str:
+    """
+    Returns a formatted websocket endpoint url.
+    Note: The port (or lack thereof) is left unchanged
```
eduardogr, line 206: Given this note It sounds that it would be nice to comment examples of the cases you are formatting

eduardogr: Given that you have tests.... you can just delegate this responsibility of the cases from this code covers to the test file

bittensor/_subtensor/subtensor_impl.py
camfairchild: I think it's usually best to create your own exceptions so you can catch them properly.

in this case, the exception inherits from a base exception for fast sync, so I can do `except FastSyncException` but I can also be more specific and except for the exact exception type.

If I made it a built-in exception, then a bug in my code may also raise this built-in and I would unknowingly handle it with the wrong assumption

tests/unit_tests/bittensor_tests/utils/test_network_utils.py
```diff
@@ -110,6 +111,40 @@ def test_upnpc_create_port_map(mocked_upnp):
     with pytest.raises(UPNPCException):
         upnpc_create_port_map(port=port)
     
+class TestGetFormattedWSEndpointURL(unittest.TestCase):
```
eduardogr, line 114: This kind of test suite is a very nice candidate to use a data provider so we can read in a few lines the different cases  of input/outputs we have here.

Take a look at: https://pypi.org/project/unittest-data-provider/

I am not suggesting that exactly library but the intention itself

eduardogr:
> > 
> 
> The binaries are from https://github.com/opentensor/subtensor-node-api Not sure where to write docs about this though

If we are getting binaries from there I would suggest to comment it somewhere so we can wire the components and generate the knowledge

@camfairchild very nice job!!

camfairchild:
> > > 
> > 
> > The binaries are from https://github.com/opentensor/subtensor-node-api Not sure where to write docs about this though
> 
> If we are getting binaries from there I would suggest to comment it somewhere so we can wire the components and generate the knowledge
> 
> @camfairchild very nice job!!

Good idea. I'll add it in the `setup.py` file. Perhaps though we should maintain a docs website in future.

bittensor/_subtensor/__init__.py
camfairchild: this is the name of the new network that @shibshib added

setup.py
eduardogr: ah i see, perfect. Thanks @camfairchild

tests/unit_tests/bittensor_tests/utils/test_network_utils.py
camfairchild: that's a good idea. I'll modify it to use something similar

bittensor/_subtensor/subtensor_impl.py
eduardogr: I see your point.

I like this consideration you can find in [PEP recommendations](https://peps.python.org/pep-0008/#programming-recommendations):

> Design exception hierarchies based on the distinctions that code catching the exceptions is likely to need, rather than the locations where the exceptions are raised. Aim to answer the question “What went wrong?” programmatically, rather than only stating that “A problem occurred” (see [PEP 3151](https://peps.python.org/pep-3151) for an example of this lesson being learned for the builtin exception hierarchy)

so, if this case is something that we need to know to differentiate the way we handle the error, it makes sense, but if not...are we using this to trace the error in case of debugging?

tests/unit_tests/bittensor_tests/utils/test_network_utils.py
eduardogr: Great!

bittensor/_subtensor/subtensor_impl.py
camfairchild: The checks are done in the lower calls and a `FastSyncException` gets raised if there was a problem during the sync
This is handled in the calling-method (e.g. `Subtensor.neurons`)

bittensor/_subtensor/subtensor_impl.py
shibshib: Agree, @camfairchild this should be moved to a proper documentation as well. Nicely done!

bittensor/_subtensor/subtensor_impl.py
camfairchild: I believe I'm using it in the former way.

The idea is that if an exception occurs, I know what it's caused by (i.e. what went wrong), so I can handle it correctly
And that way I don't handle an exception raised by something that I'm not aware of (e.g. a `LookupError` that isn't caused by having an neuron pull)

unconst (approved):

shibshib:
Perhaps I misunderstood your graphs, but looks like the ifs cache is still a little faster despite being centralized, is that right?

Also, you have conflicts ;)

camfairchild:
> Perhaps I misunderstood your graphs, but looks like the ifs cache is still a little faster despite being centralized, is that right?
> 
> Also, you have conflicts ;)

Nope, you're right. The IPFS is faster, but only when it's up ;)

shibshib:
> > Perhaps I misunderstood your graphs, but looks like the ifs cache is still a little faster despite being centralized, is that right?
> > Also, you have conflicts ;)
> 
> Nope, you're right. The IPFS is faster, but only when it's up ;)

ouch. 🗡️

bittensor/_subtensor/subtensor_impl.py
eduardogr: fair enough 👍🏽

shibshib:
@camfairchild why is this still marked do not merge? any blockers?

camfairchild:
Going to refactor into an external pypi package so the binary can be distributed easier

PR #934, opened 2022-09-29
Update dockerfile to current on dockerhub

camfairchild:
This dockerfile reflects the current docker image for `opentensorfdn/bittensor` on dockerhub.

The current image and tag is `opentensorfdn/bittensor:3.3.3-torch1.12.0-cuda11.3-pm2` 

This image contains `pm2 htop tmux curl` along with the bittensor dependencies.  

This PR also adds `netcat` to the image as I think it's useful.


Dockerfile
```diff
@@ -14,22 +15,28 @@ ARG DEBIAN_FRONTEND=noninteractive
...
+RUN apt update && apt upgrade -y
+# Install bittensor
+## Install dependencies
+RUN apt install -y curl sudo nano git htop netcat wget unzip python3-dev python3-pip tmux apt-utils cmake build-essential
+## Upgrade pip
+RUN pip3 install --upgrade pip
 
-RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y apt-utils curl git cmake build-essential unzip python3-pip  wget iproute2 software-properties-common
+# Install nvm and pm2
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
```
eduardogr: Can you check here the expected hash of the content you are downloading?
An alternative could be pinning here the file you need

shibshib: I think we can just do pip install now, no?

eduardogr: yes yes, we can generate the docker image after generating the wheel

eduardogr (approved):

PR #940, opened 2022-10-07
[Fix] multi cuda fix

camfairchild:
This PR addresses the issues with CUDA registration introduced by v3.3.4
This fixes:
- Improper termination
And this should fix:
- Reduced actual hashrate (does not match reported)
    - `nonce_start += self.update_interval * self.num_proc`
    - to
    - `nonce_start += self.update_interval * self.TPB`


eduardogr:
Is this going to be hotfixed? or following the pipeline of the release 3.3.5?

bittensor/utils/__init__.py
```diff
@@ -518,6 +510,11 @@ def solve_for_difficulty_fast( subtensor, wallet, num_processes: Optional[int] =
     stopEvent.set() # stop all other processes
     status.stop()
 
+    # terminate and wait for sovlers to exit
```
eduardogr: typo: sovlers

bittensor/utils/__init__.py
```diff
@@ -518,6 +510,11 @@ def solve_for_difficulty_fast( subtensor, wallet, num_processes: Optional[int] =
...
+    for w in solvers:
+        w.terminate()
+        w.join()
```
eduardogr: Could `w` have a more meaningful name?
i'm thinking about having the code self-explained so we don't need the previous comment. That's something we could consider as a good practice. What do you think about this?

bittensor/utils/__init__.py
```diff
@@ -671,41 +676,47 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
-            stopEvent.set() # stop all other processes
-            status.stop()
+        # exited while, solution contains the nonce or wallet is registered
+        stopEvent.set() # stop all other processes
+        status.stop()
 
-            return solution
+        for w in solvers: # terminate and wait for all solvers
+            w.terminate()
+            w.join()
```
eduardogr: given that you are repeating this little piece of code.. Would it have sense to extract into a function and use it?

eduardogr:
Looks good to me. I do not understand the whole context but seems nice.

Just a little note; if you see that you can improve readability and maintainability around the context of the bug when you are hotfixing it's a nice thing to do, if you are not in a hurry. (depending how hot is the hotfix)

eduardogr (approved):

camfairchild:
> Is this going to be hotfixed? or following the pipeline of the release 3.3.5?
@eduardogr 
this should be hotfixed

bittensor/utils/__init__.py
```diff
@@ -268,14 +264,13 @@ def __init__(self, proc_num, num_proc, update_interval, time_queue, solution_que
...
 
     def run(self):
-        block_number: int
-        block_bytes: bytes
-        block_difficulty: int
-        nonce_limit = int(math.pow(2,64)) - 1
+        block_number: int = 0 # dummy value
+        block_bytes: bytes = b'0' * 32 # dummy value
+        block_difficulty: int = int(math.pow(2,64)) - 1 # dummy value
+        nonce_limit = int(math.pow(2,64)) - 1 # U64MAX
```
shibshib, line 271: nicely done

bittensor/utils/__init__.py
```diff
@@ -635,17 +630,25 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
         
-        start_time = time.time()
-        time_since = 0.0
         solution = None
-        itrs_per_sec = 0
+        hash_rate = 0 # EWMA hash_rate (H/s)
+
+        n = 5 # number of samples to keep
+        alpha_ = 0.70 # EWMA alpha for hash_rate
+
```
shibshib: why are these numbers locally set? perhaps they should either be global or params.

bittensor/utils/__init__.py
```diff
@@ -669,43 +672,55 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
 
                 except Empty:
                     break
             
+            time_since = time.time() - start_time
             if num_time > 0:
-                time_avg = time_total / num_time
-                itrs_per_sec = TPB*update_interval*num_processes / time_avg
-                time_since = time.time() - start_time
+                time_since_last = time.time() - time_last
```
shibshib: you might run into inconsistencies here. calling time.time() and then one more time two lines later might give you a millisecond difference. Might be better to give it to a variable and use that instead. 

```
time_now = time.time()
time_since = time_now - start_time

if num_time > 0:
     time_since_last = time_now - time_last
```

bittensor/utils/__init__.py
```diff
@@ -669,43 +672,55 @@ def update_curr_block(block_number: int, block_bytes: bytes, diff: int, lock: mu
...
+                # create EWMA of the hash_rate to make measure more robust
+                hash_rate_ = ((num_time * TPB * update_interval) / time_since_last)
+                hash_rates.append(hash_rate_)
+                hash_rates.pop(0) # remove the 0th data point
```
shibshib, line 798: might be safer to put this on a try/catch.

PR #947, opened 2022-10-12
Remove locals from cli and bittensor common

unconst:
Adding install(show_locals=False) to cli/__init__.py and bittensor/__init__.py

This stops the error messages from becoming incredibly large.


camfairchild (approved):
looks good. Good change

shibshib (approved):

PR #950, opened 2022-10-13
[feature] Improve dataloader performance

joeylegere:
Quick and dirty improvement to dataloader performance. This is often a bottleneck during fine-tuning or any use case where the dataset is queried more often than a standard validator would do so.


bittensor/_dataset/dataset_impl.py
```diff
@@ -473,25 +476,22 @@ def construct_text_corpus(self, min_data_len = 0):
...
-                        total_dataset_size += int(random_datafile_dir['Size'])
-                        total_dataset_len += len(text_list)
-
-                    i += 1
-                    
-                    if (total_dataset_len > min_data_len) or self.IPFS_fails > self.IPFS_fails_max:
-                        break
+                with concurrent.futures.ThreadPoolExecutor(max_workers=cpu_count()) as executor:
+                    future_map = {}
+                    for idx, call_arg in enumerate(directories[:500]):
```
shibshib: Why the last 500 here?

shibshib (approved):
Approving to unblock, but let me know why we are using the last 500 there

bittensor/_dataset/dataset_impl.py
joeylegere: Horrible hack!

Basically, the first 500 directories (max) should have more than sufficient data for any reasonable query. 100 might also work. Didn't test it. However if you don't limit the amount you submit to the threadpool you hit some performance issues (bandwidth, and ram on smaller machines).

PR #954, opened 2022-10-13
[Hotfix] Fix CUDA Reg update block

camfairchild:
This PR hotfixes the update block portion of the CUDA registration


eduardogr:
I think that, even if this is a _hotfix_, we need a test that could verify that this change is safe.
Do we have it? Do you think it's worthy to write a test here?

eduardogr (approved):
LGTM, great job @camfairchild , nice catch!

Eugene-hu (approved):

shibshib (approved):

PR #955, opened 2022-10-13
Minor fixes

unconst:
1) Removes the parser generation on the config with messes with --help when using a parser
2) Turns off console-rich local logging (which sucks)

### DIFF
--- a/bittensor/_config/__init__.py
+++ b/bittensor/_config/__init__.py
@@ -54,7 +54,7 @@ class config:
                 Nested config object created from parser arguments.
         """
         if parser == None:
-            parser = ArgumentParser()
+            return config_impl.Config()
 
         # Optionally add config specific arguments
         try:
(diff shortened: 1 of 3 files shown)


eduardogr (approved):
LGTM

bittensor/_config/__init__.py
```diff
@@ -54,7 +54,7 @@ def __new__( cls, parser: ArgumentParser = None, strict: bool = False, args: Opt
                 Nested config object created from parser arguments.
         """
         if parser == None:
-            parser = ArgumentParser()
+            return config_impl.Config()
```
camfairchild, line 57: Why is this changed?

unconst: Because it messes with --help when importing bittensor and using argparse anywhere else.

unconst: This is because we create an argument parser on bittensor import

PR #956, opened 2022-10-15
Fallback subtensor

unconst:
Adds fallback endpoints to bittensor.subtensor allowing the user to pass a list of fallback endpoints in the event that the default fails.

btcli overview --subtensor.chain_endpoint badendpoint --subtensor.fallback_endpoints AtreusLB-2c6154f73e6429a9.elb.us-east-2.amazonaws.com:9944 --no_prompt --logging.debug


shibshib:
Would it make sense to have the Atreus load balancer a default fallback and allow folks to also specify their own fallbacks?

bittensor/_subtensor/__init__.py
```diff
@@ -136,23 +139,55 @@ def __new__(
...
-        substrate = SubstrateInterface(
+        endpoint_url = subtensor.format_endpoint(endpoint_url)
+
+        # Get fallback endpoints.
+        if fallback_endpoints == None:
+            if config.subtensor.fallback_endpoints != None:
+                fallback_endpoints = config.subtensor.fallback_endpoints
+            else:
+                fallback_endpoints = []
+        fallback_endpoints = [ subtensor.format_endpoint(end) for end in fallback_endpoints ]
```
shibshib: Can combine all of this into: 

```
fallback_endpoints = []
if fallback_endpoints == None and config.subtensor.fallback_endpoints != None:
               fallback_endpoints = config.subtensor.fallback_endpoints

fallback_endpoints = [ subtensor.format_endpoint(end) for end in fallback_endpoints ]
```

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -14,6 +14,7 @@
 # THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION 
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
 # DEALINGS IN THE SOFTWARE.
+from matplotlib.pyplot import connect
```
shibshib: is this being used?

shibshib (approved):

shibshib:
I think this requires changes on all the endpoint tests here.

eduardogr:
> Would it make sense to have the Atreus load balancer a default fallback and allow folks to also specify their own fallbacks?

@shibshib @unconst 

+1 to this, I think we should have that as default in order to have a friendly update of the bittensor nodes. That's an important point to care about.

bittensor/_subtensor/__init__.py
eduardogr: +1 to this structure, I like the order bottom-up to build the scenarios we cover programmatically. We need less "resources" to understand this code. It adds readability and maintainability

bittensor/_subtensor/subtensor_impl.py
eduardogr: seems not

eduardogr (changes requested):
LGTM ! 

I am missing tests here. For code changes we should have either the creation of new tests or the modification of existing ones.

unconst:
@eduardogr changes are added

camfairchild (approved):
LGTM

shibshib:
After the discussion we had, I'm approving this @unconst , just fix the unit tests

shibshib (approved):

tests/integration_tests/test_subtensor.py
```diff
@@ -75,19 +75,15 @@ def test_network_overrides( self ):
         assert subtensor.endpoint_for_network() == bittensor.__nobunaga_entrypoint__
 
     def test_connect_no_failure( self ):
-        self.subtensor.connect(timeout = 1, failure=False)
-
+        self.subtensor.substrate
+        assert True
```
eduardogr, line 79: What is this testing?

sounds like just no unexpected errors, is this test for that?

tests/integration_tests/test_subtensor.py
```diff
@@ -75,19 +75,15 @@ def test_network_overrides( self ):
...
-    def test_connect_fail( self ):
-        self.subtensor.substrate=None
-        with pytest.raises(RuntimeError):
-            self.subtensor.connect()
-
+        try:
+            self.subtensor.substrate
+            assert True
+        except:
+            assert False
```
eduardogr, line 86: Can this test pass with the previous test passing?

Sounds a little bit confusing for me. 
What do you think about adding assertions with more semantic to see what  `self.subtensor.substrate` is about?

unconst:
you have no idea how much of a mess this simple PR has caused in our tests

tests/integration_tests/test_subtensor.py
camfairchild: I'm pretty certain you don't need the assert
It should fail on error that isn't caught

PR #957, opened 2022-10-18
Bit 590 backward fix

isabella618033:
- Keep count of remote loss on server, saving model based on most 20 recent average loss.
- When doing tokenization remap, make sure the input size and output size is the same.
- Ensure encode_forward_causallmnext deterministic by set seed.
- When both local_train and remote_train are on: do local_train only when the server is free.
- Validator default to do backward train


bittensor/_neuron/text/core_server/run.py
```diff
@@ -366,15 +370,19 @@ def backward_callback(inputs_x:torch.FloatTensor, grads_dy:torch.FloatTensor, sy
...
-                    loss, _ = model( next( dataset ).to(model.device) )
-                    if iteration > 0 : 
-                        losses += loss
-                    else:
-                        losses = loss
-                    iteration += 1
-                    current_block = subtensor.get_current_block()
-                    logger.info(f'local training\titeration: {iteration}\tloss: {loss}')
+                if current_block != subtensor.get_current_block() and axon.priority_threadpool.is_empty:
+                    with mutex:
```
shibshib, line 374: Just a clarification required: you're making it so that the entire thread pool must be completely empty to do the backward callback? Would it make a difference if it contained one thread or two threads?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -996,6 +996,7 @@ def forward(
                 neuron_stats.setdefault(_uid, {})
                 neuron_stats[_uid].update(_stats)  # gather neuron synapse validation measures and statistics
 
+        loss.backward()
```
shibshib: we're running a backward call during a forward? That's interesting!

shibshib (approved):
looks good! Nice job using those mutexes!

bittensor/_neuron/text/core_server/run.py
isabella618033: Basically we cannot do a local_train and a remote_train at the same time. So the work around is to do a local_train only when the server is free, cause we want to put the remote_train (serving a validator) as the first priority

shibshib: Understood. So `axon.priority_threadpool` contains all threads for remote and local train jobs, correct?

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: welp you caught me :p just now when i was double checking it, the backward is actually at somewhere else, so it is not needed here

bittensor/_neuron/text/core_server/run.py
isabella618033: `axon.priority_threadpool` contains all the remote jobs, so when it is empty, it means that the server is free then we can do the local train

Eugene-hu (approved):
LGTM

PR #959, opened 2022-10-20
No set weights

unconst:
Adds flag

--neuron.no_set_weights

If set the model does not set weights.


unconst:
@camfairchild please review

bittensor/_neuron/text/core_server/run.py
```diff
@@ -442,17 +442,18 @@ def backward_callback(inputs_x:torch.FloatTensor, grads_dy:torch.FloatTensor, sy
                 # --- query the chain for the most current number of peers on the network
                 chain_weights = torch.zeros(subtensor.n)
                 chain_weights [ uid ] = 1
```
camfairchild: Should also encapsulate these lines and `try` should surround just the inner bit of the new `if`

camfairchild (approved):
Just that one change and then LGTM

PR #961, opened 2022-10-26
Bit 578 speed up metagraph storage query

camfairchild:
Note: this feature requires using https://github.com/opentensor/subtensor/pull/26 in the subtensor node that you query as the buffer size needs to be changed for the fast sync to work
<hr />  

This PR adds the [subtensorapi](https://pypi.org/project/subtensorapi/) package to grab live storage from the chain in a faster manner.  

The current live-sync takes around ~8m (See #933) using only pure-python.  

The subtensorapi package wraps a nodejs binary in python and utilizes the [`@polkadot/api`](https://www.npmjs.com/package/@polkadot/api) npm library to sync from the chain.  

This sync outputs as JSON to `~/.bittensor/metagraph.json` (by default) and then is read into python before being returned to the bittensor package.  

Below is a current graph of the performance of subtensorapi (sapi) vs the ipfs sync (current cached-sync).   
The results may be worse than average as the request times are very node-dependent. This node is hosted on a cheap contabo VPS with heavy traffic. 
I expect request times to be similar to this, if not better.

Screen Shot 2022-10-26 at 2 42 38 PM

Further, subtensorapi can be extended to support other storage values. Currently it also supports the `subtensorModule.blockAtRegistration` map using `Subtensor.blockAtRegistration_all_fast()`


requirements.txt
```diff
@@ -50,3 +50,4 @@ wandb>=0.11.1,<0.13.4
 ansible_vault>=2.1
 substrate-interface==1.2.4
 markupsafe==2.0.1
+subtensorapi>=0.0.12; sys_platform != "win32"
```
eduardogr: 😍

tests/unit_tests/bittensor_tests/utils/test_fastsync.py
```diff
@@ -0,0 +1,48 @@
+import unittest
+from unittest.mock import patch
+import pytest
+import bittensor
+try:
+    from subtensorapi.exceptions import *
+    fastsync_imported = True
+except ImportError:
+    fastsync_imported = False
```
eduardogr, line 9: why do we have this exception case?

eduardogr: wouldn't be better to isolate the test code from this library??? We could use some test double, e.g.: spy, dummy obj, mocks,..

camfairchild: This is for the exceptions only. I wanted to test the handling of exceptions caused during the interface with the subtensor API library

camfairchild: I have more tests with the class itself in the library

requirements.txt
```diff
@@ -50,3 +50,4 @@ wandb>=0.11.1,<0.13.4
 ansible_vault>=2.1
 substrate-interface==1.2.4
 markupsafe==2.0.1
+subtensorapi>=0.1.0; sys_platform != "win32"
```
eduardogr: let's use == for requirements, new library version integrated should trigger a new PR for this repository.

We do not want users to use a different version of the library if they build the code by themselves

tests/unit_tests/bittensor_tests/utils/test_fastsync.py
eduardogr: So we can avoid this here right??? What's the scenario where the `ImportError` is raised?

camfairchild: if the whole test suite is run on a machine without the package installed

shibshib (approved):

shibshib:
overall looks LGTM @camfairchild  but we need to work with @eduardogr before merging this yet, since the AWS endpoints need to support this change as well (right now they're missing the ` --ws-max-out-buffer-capacity 1024 --rpc-max-payload 1000` flags

tests/unit_tests/bittensor_tests/utils/test_fastsync.py
eduardogr: i would say that if the test suite is running without this package installed, it will fail. 
So, we do not need this and It does not seem like a good practice to me.

What do you think about this?

camfairchild: Just let the tests fail? I am primarily concerned with development and local testing happening on a machine that doesn't support `subtensorapi`
Although, we don't officially support any OS that doesn't support `subtensorapi`, so maybe this isn't a valid concern.

shibshib:
> overall looks LGTM @camfairchild but we need to work with @eduardogr before merging this yet, since the AWS endpoints need to support this change as well (right now they're missing the ` --ws-max-out-buffer-capacity 1024 --rpc-max-payload 1000` flags

Done, in this commit: https://github.com/opentensor/subtensor/commit/f5f2b8523c703b88e0712d2da7f46206a43d5691

mrseeker:
What keeps this PR from keeping blocked? Can someone pull and merge master?

eduardogr:
> What keeps this PR from keeping blocked? Can someone pull and merge master?

Hi @mrseeker, a couple of things here.
- We have a scheduled release for the 12th of December and this will may be released then.
- The release process is not just merge into master.

We will keep the community informed about this release.

Thanks for your presence here.

camfairchild:
> What keeps this PR from keeping blocked? Can someone pull and merge master?

Also keeping this blocked is an update of the rest of our subtensor nodes/chain endpoints with the new flags.

camfairchild:
> overall looks LGTM @camfairchild but we need to work with @eduardogr before merging this yet, since the AWS endpoints need to support this change as well (right now they're missing the ` --ws-max-out-buffer-capacity 1024 --rpc-max-payload 1000` flags

This should be good to go on our infra. Need to work on the migration for users, but most should be able to just use the new compose file.

camfairchild:
Closing as planned with new impl for finney

PR #963, opened 2022-10-27
addons to receptor_pool

21bloc:
Following changes that is backwards compatible

- use executor.submit to asynchronously send and process requests. This allows you to create generators (next PR ;) ) and return the first N to succeed
- min_success parameter which indicates the number of successful (or fraction if it is 0<x<len(endpoints)). This means that as we submit the endpoints, we only wait for the first {min_success} to respond
- return_type: the return type allows users to return the final result as a tuple or a dictionary. This was due to the lack of readability of using tuples. 
- graph features are included into the return object, and the user can also specify which features they want to include with {graph features}

### DIFF
--- a/bittensor/_receptor/receptor_pool_impl.py
+++ b/bittensor/_receptor/receptor_pool_impl.py
@@ -280,11 +280,7 @@ class ReceptorPool ( torch.nn.Module ):
         # resolve output type as either a tuple or a dictionary
 
         if return_type in ['tuple', tuple]:
-            return_result =  tuple(results_dict['outputs'], 
-                    results_dict['codes'], 
-                    results_dict['times'])
-
-            return return_result
+            return results_dict['outputs'], results_dict['codes'], results_dict['times']
         elif return_type in ['dict', dict]:
             return results_dict
 
--- a/bittensor/_receptor/receptor_pool_impl.py
+++ b/bittensor/_receptor/receptor_pool_impl.py
@@ -18,7 +18,7 @@
 # DEALINGS IN THE SOFTWARE.
 
 import math
-from typing import Tuple, List, Union
+from typing import Tuple, List, Union, Dict
 from threading import Lock
 
 import torch
@@ -118,13 +118,13 @@ class ReceptorPool ( torch.nn.Module ):
             synapses: List[ 'bittensor.Synapse' ],
             inputs: Union[List [ torch.Tensor ], torch.Tensor],
             timeout: int,
-            min_success = None, 
-            return_success_only=False, 
-            return_type = tuple,
-            max_workers=None,
-            graph=None,
-            graph_features=['stake', 'ranks', 'trust', 'consensus', 'incentive', 'emission', 'dividends'],
-        ) -> Tuple[List[torch.Tensor], List[int], List[float]]:
+            min_success:int = None, 
+            return_success_only:bool=False, 
+            return_type:str = 'dict',
+            max_workers:int=None,
+            graph:'bittensor.Metagraph'=None,
+            graph_features:List[str]=['stake', 'ranks', 'trust', 'consensus', 'incentive', 'emission', 'dividends'],
+        ) -> Union[Tuple[List[torch.Tensor], List[int], List[float]], Dict[str, Union[torch.Tensor, int, float]]]:
         r""" Forward tensor inputs to endpoints.
             Args:
                 endpoints (:obj:`List[ bittensor.Endpoint ]` of shape :obj:`(num_endpoints)`, `required`):
@@ -152,10 +152,10 @@ class ReceptorPool ( torch.nn.Module ):
                     Note: If you include metagraph features this will be added in the order of the {graph_features} list
                 max_workers (int),
                     Maximum number of workers. If None, defaults to the number of endpoints.
-                graph (bittensor.graph):
-                    if the graph is passed, the state of the graph will be added to the uids for additional info.
+                graph (bittensor.graph): only if return_type == 'dict'
+                    if the graph is passed, the state of the graph will be added to the uids for additional info. 
                 
-                graph_features (:obj:`List[torch.Tensor]`)
+                graph_features (:obj:`List[torch.Tensor]`) only if return_type == 'dict'
                     The list of additional graph features per uid you want to include in the response.
                     defaults: ['stake', 'ranks', 'trust', 'consensus', 'incentive', 'emission', 'dividends'],
             Returns:
@@ -167,7 +167,7 @@ class ReceptorPool ( torch.nn.Module ):
                         dendrite backward call return ops.
                     forward_times (:obj:`List[ List [float] ]` of shape :obj:`(num_endpoints * ( num_synapses ))`, `required`):
                         dendrite backward call times
-                    **graph_features)
+
                 elif return_type in [dict]:
                     dictionary of keys {outputs, codes, times, **graph_features}
         """
@@ -285,10 +285,6 @@ class ReceptorPool ( torch.nn.Module ):
                     results_dict['times'],
                     results_dict['uids']]
 
-            # add graph features if they exists
-            for k in graph_features:
-                if k in results_dict:
-                    return_result.append(results_dict[k])
             return tuple(return_result)
         elif return_type in ['dict', dict]:
             return results_dict
--- a/bittensor/_receptor/receptor_pool_impl.py
+++ b/bittensor/_receptor/receptor_pool_impl.py
@@ -79,45 +79,103 @@ class ReceptorPool ( torch.nn.Module ):
         """
         return {hotkey: v.state() for hotkey, v in self.receptors.items()}
 
-    def forward (
+
+
+    def get_receptor_hotkeys(self):
+        """
+        Returns the receptor hotkeys.
+        Returns:
+            receptor_hotkeys: list of strings 
+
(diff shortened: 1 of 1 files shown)


bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -79,45 +79,103 @@ def get_receptors_state(self):
...
             synapses: List[ 'bittensor.Synapse' ],
-            inputs: List [ torch.Tensor ],
+            inputs: Union[List [ torch.Tensor ], torch.Tensor],
             timeout: int,
+            min_success = None, 
+            return_success_only=False, 
+            return_type = tuple,
+            max_workers=None,
+            graph=None,
+            graph_features=['stake', 'ranks', 'trust', 'consensus', 'incentive', 'emission', 'dividends'],
```
joeylegere, line 126: Can you add typer hinting?

joeylegere, line 126: Can you add typer hinting?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -79,45 +79,103 @@ def get_receptors_state(self):
         """
         return {hotkey: v.state() for hotkey, v in self.receptors.items()}
 
-    def forward (
+
+
+    def get_receptor_hotkeys(self):
```
joeylegere, line 84: same for these functions, can you add hints for arg & return types?

joeylegere, line 84: Can you add type hints for these new functions?

bittensor/_receptor/receptor_pool_impl.py
joeylegere: can you type hint these as well?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,109 @@ def forward (
...
         def call_forward( args ):
             return args['receptor'].forward( args['synapses'], args['inputs'], args['timeout'] )
+        
+        # Unpack responses
+        results_dict = dict(
+            outputs=[],
+            codes=[],
+            times=[],
+            uids=[]
+        )
```
joeylegere, line 201: Using a dict here is a huuuge improvement to legibility.

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,109 @@ def forward (
...
+            min_success = int(min_success*len(endpoints))
+        elif min_success >= 1:
+            min_success = int(min(min_success,len(endpoints)))
+        elif min_success <= 0:
+            raise Exception(' REQUIRED: 0<min_success<len(endpoints)')
+
+
+        # ensure max_workers
+        if not isinstance(max_workers, int):
+            max_workers = len(endpoints)
```
joeylegere, line 217: will we run into any performance issues if we try to query too many endpoints? would a factor of `multiprocessing.cpu_count()` be safer?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,109 @@ def forward (
...
+        if return_type in ['tuple', tuple]:
+            return_result =  [results_dict['outputs'], 
+                    results_dict['codes'], 
+                    results_dict['times'],
+                    results_dict['uids']]
+
+            # add graph features if they exists
+            for k in graph_features:
+                if k in results_dict:
+                    return_result.append(results_dict[k])
```
joeylegere: I don't think we should return graph features if `return_type == tuple`, and I think we should leave it as the default (legacy) setting and try to push the `dict` return type whenever we can (refactoring the code to do so now might be beyond the scope of this PR).

My reasoning is that.. well firstly, I don't like getting 3 different tensors back and having to figure out what they are. String labels are way better. 

Secondly and more importantly, what if we change the default value for `graph_features`? If the length changes, then we'll (at best) break any call to this function that expects an exact # of values returned even if they don't explicitly request it. We're actually seeing this in the unit tests because the `uids` are returned in the tuple now and the tests are expecting 3 values when you're handing them 4. At worst we'd raise the more pernicious

More perniciously, we could change the order of items in the default arg, get them in the wrong order, and truly ruin someones day as they try to debug why their previously-working code is throwing tantrums for no real reason.

the dict approach is just all-around better for legibility and safety.

joeylegere: Also, consider the return type hinting for this function. Currently it's `Tuple[List[torch.Tensor], List[int], List[float]]`. With the graph features it'd be something like `Tuple[List[torch.Tensor], List[int], List[float], List[int], , List[float], List[float], List[float], List[float], List[float], List[float], List[float]]`. Unless you change the number of metagraph features to return.

Compared to the `Dict[str, Union[torch.Tensor, int, float]]` which is just 🔥🔥🔥

bittensor/_receptor/receptor_pool_impl.py
21bloc: This is for threading only. I think we should assume 1 process per receptor pool. This is also the default to what is already in the receptor pool

bittensor/_receptor/receptor_pool_impl.py
21bloc: updated

bittensor/_receptor/receptor_pool_impl.py
21bloc: agreed

bittensor/_receptor/receptor_pool_impl.py
21bloc: so we can fix the metagraph features, but yah, i think we should include it in the new features haha.

bittensor/_receptor/receptor_pool_impl.py
joeylegere: Oh, I realize I submitted a fix for this myself a while ago. There's actually an unused arg in `init`, `self.max_worker_threads`. Could we use that instead and remove it as an arg from the forward call?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -118,13 +118,13 @@ def forward(
...
             inputs: Union[List [ torch.Tensor ], torch.Tensor],
             timeout: int,
-            min_success = None, 
-            return_success_only=False, 
-            return_type = tuple,
-            max_workers=None,
-            graph=None,
-            graph_features=['stake', 'ranks', 'trust', 'consensus', 'incentive', 'emission', 'dividends'],
-        ) -> Tuple[List[torch.Tensor], List[int], List[float]]:
+            min_success:int = None,
```
joeylegere, line 121: nit: if `None` is the default return type this should be `Optional[int]`

joeylegere (approved):

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -79,45 +79,103 @@ def get_receptors_state(self):
...
+        Returns:
+            key (str): The hotkey of the remove receptor
+        
+
+        '''
+
+        self.receptors[ key ].close()
+        del self.receptors[ key ]
+        return key
+    delete_receptor = del_receptor = rm_receptor
```
Eugene-hu, line 108: Is this chain assignment necessary? what is it used for?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,100 @@ def forward (
...
+        
+                # when the success_response_cnt > min_success
+                if success_response_cnt >= min_success:
+                    break
+
+        # cancel all the running futures and delete the future map
+        for future, future_call_args in future_map.items():
+            future.cancel()
+            future_call_args['receptor'].semaphore.release()
+        del future_map
```
Eugene-hu, line 274: Do we have to delete the future_map? Will it not be cleared automatically?

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,100 @@ def forward (
...
-        return forward_outputs, forward_codes, forward_times
+
+
+
+        # resolve output type as either a tuple or a dictionary
+
+        if return_type in ['tuple', tuple]:
+            return results_dict['outputs'], results_dict['codes'], results_dict['times']
+        elif return_type in ['dict', dict]:
+            return results_dict
```
Eugene-hu, line 285: I really like this design! Nice work!

bittensor/_receptor/receptor_pool_impl.py
```diff
@@ -133,28 +191,100 @@ def forward (
...
+
 
         # Submit calls to receptors.
-        with concurrent.futures.ThreadPoolExecutor( max_workers = len(endpoints) ) as executor:
-            responses = executor.map( call_forward, call_args, timeout=10*timeout)
-        
-        # Release semephore.
-        for receptor in receptors:
-            receptor.semaphore.release()
+        with concurrent.futures.ThreadPoolExecutor( max_workers = max_workers ) as executor:
```
Eugene-hu, line 221: This is actually a bug with the previous receptor pool implementation. Rather than creating a separate threadpool with each forward pass, it should be using the threadpool within the receptor pool already. Do you want to include this bug fix as well? :)

Eugene-hu (approved):
LGTM, Nice Work!

PR #964, opened 2022-10-31
Feature/receptor pool addons

21bloc:
Same as pull request involving master bug merging into nobunaga


unconst:
We can't merge this since we have a dendrite rewrite in asyncio already being built

PR #967, opened 2022-10-31
Dasyncio

unconst:


bittensor/_dendrite/dendrite_impl.py
```diff
@@ -1024,4 +1023,4 @@ def to_wandb( self ):
             return wandb_info
         except Exception as e:
             bittensor.logging.error( prefix='failed dendrite.to_wandb()', sufix = str(e))
-            return {}
```
shibshib, line 1027: Unrelated but I thought we removed wandb support? Is this function vestigial?

shibshib (approved):
Looks good!

bittensor/_dendrite/dendrite_impl.py
unconst: yeah just lingering

PR #971, opened 2022-11-02
[BIT-594] Decrease validator moving average window

opentaco:
### [BIT-594] Decrease validator moving average window

Decrease validator moving average window from 20 (alpha=0.05) to 10 (alpha=0.1) steps. This parameter could probably eventually be set to alpha=0.2.

The current 20-step window means that a server model change will take 20 steps * ~250 blocks/epoch * 12 sec = approx. 17 hours to reach full score in the validator neuron stats, because of the moving average slowly weighing in new model performance. 17 hours is probably too long, and it is also likely affecting registration immunity.

[BIT-594]: https://opentensor.atlassian.net/browse/BIT-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -178,7 +178,7 @@ class neuron:
         # === Neuron statistics variables ===
         self.neuron_stats = {}  # neuron statistics dict of dicts: [uid] -> {'stat1': val1, 'stat2': val2, ...}
         self.neuron_hotkeys = []  # keep neuron hotkeys to compare and check for changes after metagraph.sync()
-        self.alpha = 0.05  # EMA coefficient in [0, 1], higher alpha discounts older observations faster
+        self.alpha = 0.1  # EMA coefficient in [0, 1], higher alpha discounts older observations faster
 
         if self.config.neuron.validation_synapse == 'TextCausalLMNext':
             self.weight_key = 'shapley_values_nxt'  # stat key + ! to calculate neuron weights with


Eugene-hu (approved):
I agree, currently it takes too long

isabella618033 (approved):
LGTM

Question tho, have we also considered changing blocks/epoch? Do we still need that to be as high as 250 * 12 sec = 50 mins?

If we push it to topk = 100, that means we need 4096 / 100 (topk) * 15 sec = 10 mins to sample through all 4096 peers. And 250 blocks/epoch means that we can sample through the network 5 times in an epoch.
And 20 steps means that we can sample the network 5*20 = 100 times before setting weight. (If we push topk from 20 -> 100)

opentaco:
> LGTM
> 
> Question tho, have we also considered changing blocks/epoch? Do we still need that to be as high as 250 * 12 sec = 50 mins?
> 
> If we push it to topk = 100, that means we need 4096 / 100 (topk) * 15 sec = 10 mins to sample through all 4096 peers. And 250 blocks/epoch means that we can sample through the network 5 times in an epoch. And 20 steps means that we can sample the network 5*20 = 100 times before setting weight. (If we push topk from 20 -> 100)

Increasing `topk` and decreasing blocks/epoch should come later to speed up network cycling time, although it may likely have the effect of increasing server load, which will require more server memory/resources, so we'd need to gradually phase in larger `topk`.

unconst:
Is this ready to merge?

unconst (approved):

PR #976, opened 2022-11-06
Support arbitrary gRPC request metadata order

adriansmares:
#### Summary

This PR fixes the gRPC server interceptor logic such that it can handle any metadata order. gRPC request metadata is a set of key values - it has no order. So depending on the fact that the order of the sender is equivalent in the receiver is not correct.

gRPC metadata is sent over as HTTP headers, and the order of HTTP headers may be changed by intermediary proxies. 

#### Changes

- Format the `AuthInterceptor` using `black`.
- Use the invocation metadata as a dictionary, not a list of key-value pairs.
- Do not trust the user provided `request_type` and use the gRPC method instead.
- Fix `request_type` provided for backward calls.

#### Testing

Tested locally by proxying the axon traffic using Traefik.

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -339,39 +339,39 @@ class axon:
         forward_callback([sample_input], synapses, hotkey='')
 
 class AuthInterceptor(grpc.ServerInterceptor):
-    """ Creates a new server interceptor that authenticates incoming messages from passed arguments.
-    """
-    def __init__(self, key:str = 'Bittensor',blacklist:List = []):
-        r""" Creates a new server interceptor that authenticates incoming messages from passed arguments.
+    """Creates a new server interceptor that authenticates incoming messages from passed arguments."""
+
+    def __init__(self, key: str = "Bittensor", blacklist: List = []):
+        r"""Creates a new server interceptor that authenticates incoming messages from passed arguments.
         Args:
             key (str, `optional`):
                  key for authentication header in the metadata (default= Bittensor)
-            black_list (Fucntion, `optional`): 
+            black_list (Fucntion, `optional`):
                 black list function that prevents certain pubkeys from sending messages
         """
         super().__init__()
-        self._valid_metadata = ('rpc-auth-header', key)
+        self._valid_metadata = ("rpc-auth-header", key)
         self.nounce_dic = {}
-        self.message = 'Invalid key'
+        self.message = "Invalid key"
         self.blacklist = blacklist
+
         def deny(_, context):
             context.abort(grpc.StatusCode.UNAUTHENTICATED, self.message)
 
         self._deny = grpc.unary_unary_rpc_method_handler(deny)
 
     def intercept_service(self, continuation, handler_call_details):
-        r""" Authentication between bittensor nodes. Intercepts messages and checks them
-        """
+        r"""Authentication between bittensor nodes. Intercepts messages and checks them"""
         meta = handler_call_details.invocation_metadata
 
-        try: 
-            #version checking
+        try:
+            # version checking
             self.version_checking(meta)
 
-            #signature checking
+            # signature checking
             self.signature_checking(meta)
 
-            #blacklist checking
+            # blacklist checking
             self.black_list_checking(meta)
 
             return continuation(handler_call_details)
@@ -380,10 +380,9 @@ class AuthInterceptor(grpc.ServerInterceptor):
             self.message = str(e)
             return self._deny
 
-    def vertification(self,meta):
-        r"""vertification of signature in metadata. Uses the pubkey and nounce
-        """
-        variable_length_messages = meta[1].value.split('bitxx')
+    def vertification(self, meta):
+        r"""vertification of signature in metadata. Uses the pubkey and nounce"""
+        variable_length_messages = meta[1].value.split("bitxx")
         nounce = int(variable_length_messages[0])
         pubkey = variable_length_messages[1]
         message = variable_length_messages[2]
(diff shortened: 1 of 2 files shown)


bittensor/_axon/__init__.py
```diff
@@ -339,39 +339,40 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
             key (str, `optional`):
-                 key for authentication header in the metadata (default= Bittensor)
-            black_list (Fucntion, `optional`): 
+                 key for authentication header in the metadata (default = Bittensor)
+            black_list (Function, `optional`):
                 black list function that prevents certain pubkeys from sending messages
         """
         super().__init__()
-        self._valid_metadata = ('rpc-auth-header', key)
+        self._signature_separator = "bitxx"
```
adriansmares: For the record I am not sure that separating the signature using `bitxx` has been the best design decision.

The signature contains the SS58 public key, and SS58 is based on [base58](https://datatracker.ietf.org/doc/html/draft-msporny-base58-01). The base58 alphabet contains all characters of `bitxx`, so if someone is (un)lucky enough have a public key that contains `bitxx`, the split will be wrong.

I recommend considering some other character, like `.`, as a separator,.

unconst:
Will this change force users to update their endpoints or is it a soft change?

adriansmares:
> Will this change force users to update their endpoints or is it a soft change?

No, this is not a breaking change - it simply enables the traffic between a dendrite and an axon to go via a (reverse) proxy. I use a highly available setup in which the same axon is served from two machines and this change enables L3 load balancing.

bittensor/_axon/__init__.py
Eugene-hu: hmm interesting, yea I agree that the design is flawed for unlucky individuals. I believe the math works out so that the probability of it occurring is ~1e-9. However, we have to be careful about changing it, since it can destroy all backwards compatibility.

bittensor/_axon/__init__.py
```diff
@@ -339,51 +339,62 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
 
         self._deny = grpc.unary_unary_rpc_method_handler(deny)
 
     def intercept_service(self, continuation, handler_call_details):
-        r""" Authentication between bittensor nodes. Intercepts messages and checks them
-        """
-        meta = handler_call_details.invocation_metadata
+        r"""Authentication between bittensor nodes. Intercepts messages and checks them"""
+        method = handler_call_details.method
+        metadata = dict(handler_call_details.invocation_metadata)
```
Eugene-hu: Nice change! I like the dictionary design much better.

bittensor/_axon/__init__.py
```diff
@@ -339,51 +339,62 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
-        """
-        variable_length_messages = meta[1].value.split('bitxx')
+    def get_signature(self, meta):
+        r"""get_signature from the metadata. Raises exception when the signature is missing"""
+        signature = meta.get("bittensor-signature")
+        if signature is None:
+            raise Exception("Request signature missing")
+        return signature
+
+    def verification(self, metadata):
```
Eugene-hu: Can you put type hints in?

Eugene-hu (commented):
Nice PR!  I like the new design. I think the overall authentication system is due for a big cleanup. This is a good step in the right direction.

bittensor/_axon/__init__.py
```diff
@@ -339,101 +339,114 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
+        r"""Attempts to parse a signature using the legacy format, using `bitxx` as a separator"""
+        parts = signature.split("bitxx")
+        if len(parts) < 4:
+            return None
+        try:
+            nonce = int(parts[0])
+            parts = parts[1:]
+        except ValueError:
+            return None
+        receptor_uuid, parts = parts[-1], parts[:-1]
```
Eugene-hu, line 369: Any reason this is parsed backwards?

Eugene-hu (approved):
LGTM

bittensor/_axon/__init__.py
adriansmares: In this way if someone unfortunate enough to have `bitxx` in their public key, the split is still correct - you basically have a string that looks like `[ nonce, pubkey_fragment_1, pubkey_fragment_2, ..., signature, receptor_uid ]`, so you parse these 'safe' segments and the remainder is expected to be `[ pubkey_fragment_1, pubkey_fragment_2 ... ]`, which can be used to rebuild the pubkey.

bittensor/_axon/__init__.py
```diff
@@ -339,101 +339,114 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
+        endpoint_key = f"{pubkey}:{receptor_uuid}"
+
+        if endpoint_key in self.nonces.keys():
+            previous_nonce = self.nonces[endpoint_key]
+            # Nonces must be strictly monotonic over time.
+            if nonce - previous_nonce <= -10:
+                raise Exception("Nonce is too small")
+            if not keypair.verify(message, signature):
+                raise Exception("Signature mismatch")
+            self.nonces[endpoint_key] = nonce
```
shibshib, line 402: Any reason for the `10` nonce diff? i.e. is it an arbitrary value?

adriansmares: This is from the original source code - you can see it live in `nobunaga` and even master here: https://github.com/opentensor/bittensor/blob/68a8c0a247f99bdba9f1e17e3a8b3ca0ef8e5e6d/bittensor/_axon/__init__.py#L399

The nonce check is quite flawed and I've actually fixed it as part of the refactor being prepared in https://github.com/opentensor/bittensor/pull/983 ([here](https://github.com/opentensor/bittensor/pull/983/commits/edb73e4c85f61174e75cb0a861fef570af98344e)) , but I didn't want to delay this PR further. There are lots of details to review here, so I tried not to make it too complicated and I've separated the PRs.

shibshib: Understood

shibshib (approved):

unconst:
This can be closed at the expense of the newer PR with more features.

Eugene-hu:
@unconst Let's merge this version first. This PR is ready to merge, whereas the other PR still needs more time

PR #978, opened 2022-11-07
Feature/dataset impl addon

21bloc:
This is an asyncio implementation of the dataset loader. 
The following changes should be noted
- The producer thread is running in the background on another seperate asyncio loop
- The text hashes are saved locally, not the text


unconst:
Changing base to nobunaga

unconst:
Actually I cannot change the base, can you please

PR #979, opened 2022-11-07
dataset addon commit from general axon

21bloc:
This is an asyncio implementation of the dataset loader. 
The following changes should be noted
- The producer thread is running in the background on another seperate asyncio loop
- The text hashes are saved locally, not the text


bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+        self.shutdown()
+
+    def shutdown(self, wait=True):
+        if wait:
+            for t in self._threads:
+                try:
+                    t.join()
+                except Exception:
+                    pass
+
```
shibshib: Do we not know what kind of exception we're going to get? If not, can we at least print the message? Makes debugging much easier down the line. This will cause someone to grow gray hairs in the future :D

bittensor/_dataset/__init__.py
```diff
@@ -97,7 +98,8 @@ def __new__(
                 max_directories = config.dataset.max_directories
             )
         else:
-            return dataset_impl.GenesisTextDataset(
+
+            return dataset_impl.Dataset(
```
shibshib: Does it make sense to simply name this "Dataset"? It specifically encompasses text data, not all possible datasets.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+        self.max_threads = max_threads
+        self._idle_semaphore = threading.Semaphore(0)
+        self._threads = []
+        self._shutdown_lock = threading.Lock()
+        self._shutdown = False
+
+    def submit(self, fn, args=[],kwargs={}):
+        with self._shutdown_lock:
+            if self._shutdown:
+                raise RuntimeError('cannot schedule new futures after shutdown')
```
shibshib: Nicely done

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+        # if datasets is None then refer to all of the availabe datasets 
+        if len(self.dataset_name) == 0 or self.dataset_name == 'default':
+            self.dataset_name = self.available_datasets
+
         self.no_tokenizer = no_tokenizer
-        self.IPFS_fails = 0
-        self.backup_dataset_cap_size = 5e7 # set 50MB limit per folder
-        self.IPFS_fails_max = 10
+        if not  self.no_tokenizer:
+            self.set_tokenizer(tokenizer=tokenizer)
```
shibshib: This is a bit confusing. If there is not no tokenizer then set a tokenizer? Is this checking if there is indeed a tokenizer or that there is no tokenizer?

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
             
-        return text 
 
-    def get_dataset(self , file_meta):
-        r""" Either load a dataset, which is a list of hashes, from disk or download it from IPFS
-        Args:
-            file_meta (dict of str: int):
-                Specify the details of the dataset in the format of {'Name': , 'Hash':}.
+            if (len(new_text_hashes) + len(loaded_text_hashes)) >num_samples:
+                break
```
shibshib: Let's leave a message here too to explain why we broke out.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+            data = await self.api_post(f'{self.ipfs_url}/cat', params={'arg':file_meta['Hash']}, return_json=False, num_chunks=10)
+        except KeyError:
+            return []
+        decoded_hashes = []
+        hashes = ['['+h + '}]'for h in data.decode().split('},')]
+        for i in range(len(hashes)-1):
+            try:
+                decoded_hashes += [json.loads(hashes[i+1][1:-1])]
+            except json.JSONDecodeError:
+                pass
```
shibshib: error message here too

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
-""" Implementation for the dataset and GenesisTextDataset class, which handles dataloading from ipfs
-"""
-# The MIT License (MIT)
```
unconst, line 3: You should keep this and add an additional Copyright from Opentensor for 2022

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
-        ):
-        """ Creates a retriable session for request calls. This enables
-        automatic retries and back-off retries should any request calls fail.
+    ipfs_url = 'http://global.ipfs.opentensor.ai/api/v0'
+    dataset_dir = 'http://global.ipfs.opentensor.ai/api/v0/cat' 
+    text_dir = 'http://global.ipfs.opentensor.ai/api/v0/object/get'
+    mountain_hash = 'QmSdDg6V9dgpdAFtActs75Qfc36qJtm9y8a7yrQ1rHm7ZX'
+
+    def __init__(self, 
+                batch_size: int=8,
```
unconst: indent

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
-        except Exception as E:
-            logger.error(f"Failed to get from IPFS {file_meta['Name']} {E}")
-            return None
-
-        return response
-
-    def __len__(self):
-        """ Returns length of the dataset that the dataset is processing
-        """
+            loop (asyncio.loop):
```
unconst: Use standard naming here

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
-                pass
+            # gather dataset hashes async as their state is independent
+            for dataset in datasets:
+                tasks += [self.build_dataset(dataset=dataset, save=save, load=load, loop=loop)]
 
-            return text
-        
-        return None
+            dataset_hashes = self.async_run(asyncio.gather(*tasks), loop=loop)
+
```
unconst: Remove double empty lines, better to fill with comments if you need separation

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
-            self.IPFS_fails = 0
+
+        dir_path = os.path.dirname(path)
+
+        # ensure the json is the prefix
+        if path[-len('.json'):] != '.json':
+            path += '.json'
+
+        # ensure the directory exists, make otherwise
+        if not os.path.isdir(dir_path):
```
unconst, line 206: Should check if writable also see here: https://github.com/opentensor/bittensor/blob/521a367d66e68fcc5fb600f08ac6d5f72cfdcab8/bittensor/_keyfile/keyfile_impl.py#L369

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+
+    def save_json(self,loop:'asyncio.loop'=None, *args,**kwargs) -> str:
+        '''
+        Sync verson of async_save_json
+
+        Args
+            loop (asyncio.loop):
+                The asyncio loop to be past, otherwise self.loop
+
+        Returns (str)
```
unconst: See format standard for return arguments

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+        Returns: path (str)
+            Path of the saved JSON
+        """
+        
+        
+        if include_root:
+            path = os.path.join(self.data_dir, path)
+
+
+        # ensure extension
```
unconst: Capitals and periods at the end of comments

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,626 +1,800 @@
...
+                The asyncio loop to be past, otherwise self.loop
+
+        Returns (dict, list) 
+
+        '''
+        return self.async_run(job=self.async_load_json(*args,**kwargs), loop=loop)
+
+    
+    async def build_dataset(self, dataset=None, num_folders=10, num_samples=100, save=False, load=True, loop=None):
+
```
unconst, line 356: Requires argument docs

bittensor/_dataset/dataset_impl.py
```diff
@@ -633,46 +807,10 @@ def __len__(self):
...
-            
-            if response:
-                dataset_hashes = response.json()['Links']
-                if self.save_dataset:
-                    self.save_hash(mountain_meta, json.dumps(dataset_hashes) )
-            
-            if self.IPFS_fails > self.IPFS_fails_max and response == None:
-                dataset_hashes = json.loads(self.load_hash(mountain_meta))
-                break
+if __name__ == '__main__':
```
unconst: Remove this when done testing

bittensor/_dataset/dataset_impl.py
21bloc: i just removed it, it was from an existing thread_pool impl.

bittensor/_dataset/dataset_impl.py
21bloc: This is now depracted as i use a dictionary to better track new hashes.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-            elif action == 'post':
-                response = self.requests_retry_session(session=session).post(address, timeout=timeout)
-            logger.success("Loaded from IPFS:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
+        self.max_threads = max_threads
+        self._idle_semaphore = threading.Semaphore(0)
+        self._threads = []
+        self._shutdown_lock = threading.Lock()
+        self._shutdown = False
+
+    def submit(self, fn, args:list=[],kwargs:dict={}):
```
joeylegere: Default args shouldn't be mutable. Should be `Optional[list] = None` and instantiate it as `[]` within the function. (same for the kwargs)

joeylegere: Or just use the standard `*args` and `**kwargs`, should behave fine since they become a tuple and a dict within here.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-    """ One kind of dataset that caters for the data from ipfs 
+
+class GenesisTextDataset:
+    """ Implementation for the dataset class, which handles dataloading from ipfs
     """
+
+    ipfs_url = 'http://global.ipfs.opentensor.ai/api/v0'
+    dataset_dir = 'http://global.ipfs.opentensor.ai/api/v0/cat' 
+    text_dir = 'http://global.ipfs.opentensor.ai/api/v0/object/get'
+    mountain_hash = 'QmSdDg6V9dgpdAFtActs75Qfc36qJtm9y8a7yrQ1rHm7ZX'
```
joeylegere: These should be defined in a separate file, e.g. `bittensor.__init__` where a lot of other constants are defined.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-        max_directories
-    ):
-        super().__init__()
-        self.block_size = block_size
+            self, 
+            batch_size: int=8, 
+            sequence_length: int=256,
+            block_size: int = 10000,
+            num_workers: int = 1,
+            dataset_name: List[str]=['ArXiv'],
```
joeylegere: Same as the other comment, default args shouldn't be mutable.

joeylegere: Also, consider using `typing.Sequence` if it doesn't *have* to be a list (i.e. any reason i can't throw a tuple?)

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
         self.dataset_name = dataset_name
-        self.data_dir = data_dir
-        self.save_dataset = save_dataset
-        self.datafile_size_bound = 262158
+        self.set_event_loop(loop=loop)
+        # if datasets is None then refer to all of the availabe datasets 
         self.max_datasets = max_datasets
-        self.__infinite_dataset_iterator = None
+        if len(self.dataset_name) == 0 or self.dataset_name == 'default':
+            self.dataset_name = self.available_datasets
```
joeylegere: I would just raise an exception if `len(datasets) == 0`. And because you're explicitly specifying the type in the function call, i wouldn't allow a user to pass in "default" (unless this is important for legacy reasons in the rest of the codebase, in which case we should at least document this).

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
+        self.datasets = self.dataset_name = self.dataset_name[:self.max_datasets]
```
joeylegere: Do we need to have two variables with the same value, or does one get mutated later and mean something different? Tbh I'd prefer just keeping `self.datasets` as its a bit more descriptive. `self.dataset_name` sounds like it should just be a singular `str` imo, which it very much isn't

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
+import json
+import os
+import threading
+import asyncio
+import nest_asyncio
+from loguru import logger
+##########################
+##### Get args ###########
+##########################
+from typing import *
```
joeylegere: i've got a lot of unused imports flagged here, could you tidy them? (`pylint` should flag them easy and you can use `isort` to do PEP8 sorting super quick)

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
 
-from .thread_queue import ThreadQueue
+import streamlit as st
```
joeylegere: i know you're just using this for debugging but make sure to remove it as its not in requirements.txt and breaks tests.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
+        """
+        
+        # this is for starting a new thread
+        # the loop needs to be set within the new thread
+        if loop != None:
+            asyncio.set_event_loop(loop)
 
-        self.build_hash_table()
+        # chunk the text hashes into batch_sie chunks
+        text_hash_batch_list = self.chunk(self.all_text_file_metas,
```
joeylegere: I don't see this variable used in this function. do you mean to iterate over it below?

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-                pass
+        if len(dataset_hash_map) == 0:
+            tasks = []
 
-            return text
-        
-        return None
+            # Gather dataset hashes async as their state is independent.
+            for dataset in datasets:
+                tasks += [self.build_dataset(dataset=dataset, save=save, load=load, loop=loop)]
```
joeylegere: I think `self.build_dataset` should have a more descriptive name like `build_single_dataset`. It's too similar to this functions name (one character off) and I misunderstood why we were recursively calling this function in an asyncio loop lol

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-                with open(full_path, mode='r') as f:
-                    text = f.read()
+        all_text_file_metas = []
+        dataset_hash_map = {}
 
-                logger.success("Loaded from disk:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-            except Exception:
-                logger.success("Could not load from disk:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-                pass
+        if len(dataset_hash_map) == 0:
```
joeylegere: will this ever `!=0`? you define it in the previous line.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-                if self.save_dataset :
-                    self.save_hash(file_meta, json.dumps(response.json()) )
-                    
-            else:
-                self.IPFS_fails += 1
-                logger.warning("Failed to get dataset".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-                return None
+        # ensure the json is the prefix
+        if path[-len('.json'):] != '.json':
+            path += '.json'
```
joeylegere, line 203: fyi, https://docs.python.org/3/library/os.path.html#os.path.splitext is easier/more legibile for the future

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-                h['Folder'] = file_meta['Name']
-            return hashes 
+        # ensure the directory exists, make otherwise
+        if not os.path.isdir(dir_path):
+            os.makedirs(dir_path)
 
-    def get_hashes_from_dataset(self):
-        r""" Getting directories .
-        Where a directory could be leading to a data file or a directory file.
+        assert os.access( path , os.R_OK )
```
joeylegere: Maybe a warning is more appropriate here, rather than halting the entire program if we can't save the data

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
+        '''
+        Sync verson of async_save_json
+        Args
+            loop (asyncio.loop):
+                The asyncio loop to be past, otherwise self.loop
 
-                else:
-                    logger.error('Incorrect dataset name:'.ljust(20) + " <red>{}</red>.".format(key)+' Must be one of the following {}'.format(bittensor.__datasets__))
+        Returns 
+            output (dict)
```
joeylegere: does this return a str or a dict? comments contradict type hints

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
-""" Implementation for the dataset and GenesisTextDataset class, which handles dataloading from ipfs
+""" Implementation of Axon, services Forward and Backward requests from other neurons.
```
joeylegere: Artefact from the other branch? I see axons mentioned in some of the docstrings as well

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
+import argparse
 from torch.utils.data.dataloader import DataLoader
-
+from tqdm import tqdm
 import bittensor
+import glob
+from queue import Queue
+import numpy as np
+import asyncio
+import aiohttp
```
joeylegere, line 21: missing from requirements.txt

joeylegere (commented):
Not related to this PR, but there seems to be a pre-existing bug when I try to run this code at least on my mac. In line 178 of `bittensor/_dataset/__init__.py` is currently `defaults.dataset.data_dir = os.getenv('BT_DATASET_DATADIR') if os.getenv('BT_DATASET_DATADIR') != None else '~/.bittensor/data/'`.

This should use `os.path.expanduser()` as your default arguments do, otherwise it creates a local directory called `~` within the current working directory.

bittensor/_dataset/dataset_impl.py
joeylegere: This should check access on `dir_path`, and we're probably more interested in `os.W_OK` rather than if it's readable. Because `path` is the path to the file (and because it doesn't exist yet), it's going to raise an error here the first time, every time.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,536 +1,490 @@
...
-                    
-                    return self.get_root_text_hash(random_sub_directory)
-                else:
-                    logger.warning("Directory seems empty, ignoring directory:".ljust(20) + "<blue>{}</blue>". format(file_meta))
-        return None
+        # Load default if file does not exist.
+        try:
+            with open(path, 'r') as f:
+                obj = json.load(f)
+        except FileNotFoundError:
```
joeylegere, line 254: It's possible for this file to be empty, i.e. if someone executes this script and the file is partially created but not written to (as happened to me in my testing). Should add handling for `JSONDecodeError` as well.

bittensor/_dataset/dataset_impl.py
21bloc: I left the assert but changed to W_OK.

bittensor/_dataset/dataset_impl.py
21bloc: fair, must have been from an old code block

bittensor/_dataset/dataset_impl.py
21bloc: removed

bittensor/_dataset/dataset_impl.py
21bloc: cleaned it up by removing unused libraries

tests/integration_tests/test_dataset.py
```diff
@@ -25,6 +25,47 @@ def test_construct_text_corpus():
     dataset.construct_text_corpus()
     dataset.close()
 
+def test_change_data_size():
```
joeylegere, line 28: This is a good test. One thing I'd suggest is to also test your failure cases. That is, I see you have a local function called `check_valid` in there. You should try to trigger that as well. IMO you should just raise a `ValueError`, and use an `assertRaises`. Check https://stackoverflow.com/a/61017213

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,538 +1,494 @@
...
 import bittensor
-
-from .thread_queue import ThreadQueue
+from queue import Queue
+import numpy as np
+import asyncio
+import aiohttp
+import threading
+from loguru import logger
+from typing import *
```
joeylegere: Couple unused imports here still (numpy and tqdm). Can you run `isort` just to tidy it up?

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,538 +1,494 @@
...
+        self.datasets = datasets
+        self.num_folders = num_folders
+        self.set_event_loop(loop=loop)
+        # if datasets is None then refer to all of the availabe datasets 
         self.max_datasets = max_datasets
-        self.__infinite_dataset_iterator = None
+        assert len(self.datasets) > 0
+        if self.datasets == 'default':
+            self.datasets = self.available_datasets
+        self.datasets = self.datasets[:self.max_datasets]
```
joeylegere: `datasets` is of type `Optional[Union[List[str], str]]`. As an end-user I might understandably read this as if I can pass in just `"ArXiv"` (because passing in a list for a single dataset is counterintuitive). We should either support this behaviour, or not allow strings at all. (and have an empty list or `None` be the default value and behave as "default" does now).

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,538 +1,494 @@
...
+            sequence_length: int=256,
+            block_size: int = 10000,
+            num_workers: int = 1,
+            datasets: Optional[Union[List[str], str]]=['ArXiv'], 
+            loop:'asyncio.loop'=None, 
+            no_tokenizer: bool = False,
+            data_dir: str =  os.path.expanduser('~/./bittensor/data'),
+            save_dataset : bool = False,
+            load_dataset : bool = True,
+            run_generator:bool=False,
```
joeylegere: You should write a unit test where `run_generator is True`. Both as proof it works as intended and to ensure nobody else inadvertently breaks it in the future.

bittensor/_dataset/dataset_impl.py
joeylegere: The type hints in the code and the docstring here still contradict each other.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,538 +1,494 @@
...
+            num_samples (int):
+                The number of samples the user want so get from the dataset
+            load (bool):
+                Load the dataset hashes locally
+            save (bool):
+                Save the dataset hahses locally.
+            loop (asyncio.Loop):
+                Asyncio loop 
+
+        Returns: None
```
joeylegere: This docstring is also a lie.

Personally I'm in favour dropping type hints in docstrings (@unconst, thoughts?). It's redundant and not in a good way. They're easy to miss when refactoring, and tbh the type hints in the code should just tell the story themselves. This also facilitates static type checking w/ `mypy` or similar, which I'm strongly in favour of adding to our dev workflow.

bittensor/_dataset/dataset_impl.py
```diff
@@ -560,119 +516,471 @@ def check_valid(size):
...
+
+    def set_cache(self, cache_size=10, cache_calls_per_block=100) -> None:
+        self.cached_text_list = [] # Cache list for raw text.
+        self.cache_size = cache_size  # The maximum size of the cache (FIFO).
+        self.calls_for_current_block = 0 # The number of current calls on the set of blocks
+        self.cache_calls_per_block = cache_calls_per_block # The maximum calls ber block before replacing a block with a new block.
+
+
+
+    def __getitem__(self, idx: int= None, filler_token:str='FILLER_TEXT') -> Union[str, torch.tensor]:
```
joeylegere: nit. idx should be `Optional[int]` if its default is `None`.

bittensor/_dataset/dataset_impl.py
21bloc: I had left it to have default or a list, but i see what you mean. I can have default and None as the trigger for every dataset.

bittensor/_dataset/dataset_impl.py
21bloc: fixed the docstring

tests/integration_tests/test_dataset.py
21bloc: I dont really know what this mean, perhaps clarify so i can get a better understanding. I did add two tests that check the inputs that are tokenized and not tokenized under the generator and not the generator.

joeylegere: Yup.

So basically here: https://github.com/opentensor/bittensor/pull/979/files#diff-3ba712d25bdb72a19a0b805fa92e1f6167b76383a937684d92fee4017b59846fR493
you define a function called `check_valid`, but none of your test cases will hit it.

That is, you're not passing in a `size <= 0` or a size that isn't an `int`. You should add test cases for these edge cases as well.

My other point was this block:

```
        if check_valid(batch_size):
            self.batch_size = batch_size
        
        if check_valid(block_size):
            self.block_size = block_size
```

I would argue that if *either* your batch size or block size is invalid to throw an error, rather than silently only setting one of them. e.g. if I passed in (10, -100), i should raise an error, rather than setting the batch_size to 10 and persisting the same block_size (especially cause the logger call in there indicates "success", when one of them, at least imo, failed).

bittensor/_dataset/__init__.py
```diff
@@ -40,13 +40,21 @@ class dataset:
     def __new__(
             cls,
             config: 'bittensor.config' = None,
-            block_size: int = None,
+            block_size: int = 500,
```
joeylegere: This needs to be larger; i'm missing tokens with seq_length = 256. 5* sequence length is a good rule of thumb, try 1500?

bittensor/_dataset/__init__.py
```diff
@@ -164,7 +194,7 @@ def add_defaults(cls, defaults):
         defaults.dataset.block_size = os.getenv('BT_DATASET_BLOCK_SIZE') if os.getenv('BT_DATASET_BLOCK_SIZE') != None else 20
```
joeylegere, line 189: This should probably also get updated, 20 is the old seq_length on nobunaga. I'd actually prefer if we change this variable and leave it as `None` in the `new()` method, otherwise someone may change this line and accidentally not update it correctly.

bittensor/_dataset/dataset_impl.py
```diff
@@ -542,137 +537,494 @@ def set_data_size(self, batch_size, block_size):
...
+            raw_text =raw_text.split()
+            output = raw_text[:self.sequence_length]
+            remainder = self.sequence_length - len(output)
+            if remainder > 0:
+                output = output + [self.pad_token]*remainder
+            output = ' '.join(output)
+        else:
+            
+            tokenized_text =  self.tokenizer(raw_text, padding=True)['input_ids']
+            output = torch.tensor(tokenized_text)[:self.sequence_length]
```
joeylegere: Cast to `torch.long` here, otherwise empty strings produce tensors with type `torch.float32` and cause errors with tokenizers downstream.

bittensor/_dataset/dataset_impl.py
```diff
@@ -542,137 +537,494 @@ def set_data_size(self, batch_size, block_size):
...
+            sample_cat_params_list = random.sample(self.sample_cat_params_list, self.cache_size)
+            cache_sample_tasks = [self.cat(**sample_cat_params) for sample_cat_params in sample_cat_params_list]
+            self.sample_cache = [sample_bytes for sample_bytes in self.async_run(asyncio.gather(*cache_sample_tasks))]
 
-    def dataloader(self, epoch_length = 100):
-        r""" Creates a torch dataloader out of a subclass of this class.
+
+        return self.sample_cache
+
+    def __getitem__(self, idx: Optional[int]= None, filler_token:str='FILLER_TEXT') -> Union[List[str], torch.tensor]:
```
joeylegere: probably should remove `filler_token` since we default to `self.pad_token` to avoid ambiguity.

bittensor/_dataset/dataset_impl.py
```diff
@@ -542,137 +537,494 @@ def set_data_size(self, batch_size, block_size):
...
+            if remainder > 0:
+                output = output + [self.pad_token]*remainder
+            output = ' '.join(output)
+        else:
+            
+            tokenized_text =  self.tokenizer(raw_text, padding=True)['input_ids']
+            output = torch.tensor(tokenized_text)[:self.sequence_length]           
+            # Append the remainder with 0 (TODO use stop token)
+            seqeunce_length_remainder =  self.sequence_length - output.shape[0]
+            if seqeunce_length_remainder>0:
```
joeylegere: We should pad from the left-hand side as thats the tokenizers behaviour by default.

bittensor/_dataset/dataset_impl.py
```diff
@@ -542,137 +537,494 @@ def set_data_size(self, batch_size, block_size):
...
+                output = torch.nn.functional.pad(input=output, pad=(0,seqeunce_length_remainder), mode='constant', value=self.pad_token_idx  )
```
joeylegere: you can simplify this in one line:
`self.tokenizer("test", max_length=self.sequence_length, truncation=True, padding="max_length", return_tensors="pt")["input_ids"]`

bittensor/_dataset/dataset_impl.py
```diff
@@ -542,137 +537,494 @@ def set_data_size(self, batch_size, block_size):
...
+        try:
+            raw_text = raw_text_bytes.decode()
+        except UnicodeDecodeError as e:
+            raw_text = str(raw_text_bytes[2:-1])
+        if self.no_tokenizer:
+            raw_text =raw_text.split()
+            output = raw_text[:self.sequence_length]
+            remainder = self.sequence_length - len(output)
+            if remainder > 0:
+                output = output + [self.pad_token]*remainder
```
joeylegere: same here, convention is to pad from the left hand side

bittensor/_dataset/__init__.py
```diff
@@ -161,10 +191,10 @@ def add_defaults(cls, defaults):
         """
         defaults.dataset = bittensor.Config()
         defaults.dataset.batch_size = os.getenv('BT_DATASET_BATCH_SIZE') if os.getenv('BT_DATASET_BATCH_SIZE') != None else 10
-        defaults.dataset.block_size = os.getenv('BT_DATASET_BLOCK_SIZE') if os.getenv('BT_DATASET_BLOCK_SIZE') != None else 20
+        defaults.dataset.block_size = 1500
```
joeylegere: should still follow the convention of allowing an environment variable. Also, we need one for the new sequence length yeah?

21bloc:
image

bittensor/_dataset/utils.py
```diff
@@ -0,0 +1,143 @@
+import threading
+import ctypes
+import time
```
joeylegere: unused imports

bittensor/_dataset/utils.py
```diff
@@ -0,0 +1,143 @@
...
+from typing import *
+import asyncio
+
+class CustomThread(threading.Thread):
```
joeylegere: can you add a short docstring that explains exactly what functionality this adds?

joeylegere (approved):
After a few grueling weeks, I think this is looking pretty damn solid.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,750 @@
...
+        self.construct_text_corpus(datasets=self.datasets, load=self.load_dataset, save=self.save_dataset)
 
-        # Used to refresh corpus if we've exhausted the whole dataset
-        self.refresh_corpus = True
+    def construct_text_corpus(self, datasets:Optional[List[str]] = None, save:bool = False, load:bool = False) -> None :
+        """ Building all of the datasets specified by getting each of their 
+            text hashes from IPFS or local
+        Args:
+            datasets (List[str]):
+                Axon to serve.s
```
unconst: Attention to detail is required for this to be approved. Stopping my review here.

unconst (changes requested):

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,750 @@
...
+                              include_root:bool=True) -> str:
+        """ 
+        Async save of json for storing text hashes
 
-        Returns:
-            text (str): 
-                The text in the file.                
+        Args:
+            path (List[str]):
+                Axon to serve.
```
unconst: Again.

unconst (changes requested):

bittensor/_dataset/dataset_impl.py
21bloc: did a run through lingering syntax to make her pretty for the prom.

bittensor/_dataset/__init__.py
```diff
@@ -40,21 +40,32 @@ class dataset:
...
-            _mock:bool=None
+            sequence_length: Optional[int] = None,
+            block_size: Optional[int] = None,
+            block_size_bytes: Optional[int] = None,
+            max_hash_size: Optional[int] = None,
+            batch_size: Optional[int] = None,
+            num_workers: Optional[int] = None,
+            dataset_name: Optional[list] = None,
+            max_directories: Optional[int] = None,
+            max_datasets: Optional[int] = None,
```
unconst, line 51: There is no corresponding comment.

bittensor/_dataset/__init__.py
```diff
@@ -68,56 +79,67 @@ def __new__(
...
 
     @classmethod
     def config(cls) -> 'bittensor.Config':
-        """ Get config from the argument parser 
-            Return: bittensor.config object
+        """ 
+        Get config from the argument parser 
+            
+        Return: 
+            bittensor.config
```
unconst: typing

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
-import concurrent
-import json
 import os
 import random
-import time
-from multiprocessing import cpu_count
-from typing import Union
-
-import requests
+from copy import deepcopy
```
unconst: These can be reorganized.  Either alphabetical, sorted by from syntax and import syntax or by length.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
 
-        # Retrieve a random slice of the genesis dataset
-        self.data = []
-        self.data_reserved = []
+        # Build the text corpus by fetching the hashes of the textfiles (Current Heirarchy)
+        self.construct_text_corpus(datasets=self.datasets, load=self.load_dataset, save=self.save_dataset)
 
-        # Used to refresh corpus if we've exhausted the whole dataset
-        self.refresh_corpus = True
+    def construct_text_corpus(self, datasets:Optional[List[str]] = None, save:bool = False, load:bool = False) -> None :
```
unconst, line 136: Nice typing

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
-        Args:
-            file_meta (dict of str: int):
-                Specify the details of the dataset in the format of {'Name': , 'Hash':}.
+        self.dataset_hash_map = dataset_hash_map
 
-        Returns:
-            text (str): 
-                The text in the file.                
+        # Flatten the hashes to a list of hashes.
+        for  k,file_meta_list in dataset_hash_map.items():
```
unconst: Some strange spacing here. Generally, there should not be double spaces between variables nor extra lines. If you need an extra line it is a good opportunity to insert a comment.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
-                with open(full_path, mode='r') as f:
-                    text = f.read()
+        dir_path = os.path.dirname(path)
 
-                logger.success("Loaded from disk:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-            except Exception:
-                logger.success("Could not load from disk:".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-                pass
+        # Ensure the json is the prefix.
+        if path[-len('.json'):] != '.json':
```
unconst, line 202: clean

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+                Path of the loaded json
+            include_root (bool, optional):
+                Include self.data_dir as the prefix.
+                    - if True, ths meants shortens the batch and 
+                    specializes it to be with respect to the dataset's 
+                    root path which is in ./bittensor/dataset
+            default (dict, optional):
+                If there is a file not found, what is the default value.
+        Returns: 
+            obj (dict)
```
unconst, line 232: Does not match the typing on the function.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        return path
+
+    save_json = sync_wrapper(async_save_json)
 
-    def save_hash(self, file_meta, text):
-        r""" Save a hash to disk.
+    async def async_load_json(self, path:str, include_root:Optional[bool] = True, default:Optional[Union[list, dict]] = {}) -> Union[list, dict]:
+
+        """ 
+        Async save of json for storing text hashes
```
unconst, line 220: More description here could be helpful. Saved what? to where?

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        except FileNotFoundError:
+            obj = default
+        except json.JSONDecodeError:
+            obj = default
+        if isinstance(obj, str):
+            obj = json.loads(obj)
+
+        return obj
+
+    load_json = sync_wrapper(async_load_json)
```
unconst, line 263: These function extensions should include a comment. It is not obvious, at least to me, what you are accomplishing with this.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
-
-                # --- Save text if the save_dataset flag is on.
-                if self.save_dataset :
-                    self.save_hash(file_meta, json.dumps(response.json()) )
-                    
-            else:
-                self.IPFS_fails += 1
-                logger.warning("Failed to get dataset".ljust(20) + "<blue>{}</blue>".format(file_meta['Name']))
-                return None
+    def set_data_size(self, batch_size:Optional[int] = None, block_size:Optional[int] = None, sequence_length:Optional[int] = None,  block_size_bytes:Optional[int]= None, buffer_size:Optional[int]=None) -> None:
```
unconst, line 316: This line is too long. Generally, 80-120 characters no more. I find it easier to read if the arguments are indented on new lines.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        r""" 
+        Update the size of data (batch_size, sequence_length, block_size_bytes) that we need.
 
-        if hashes == None:
-            return None
-        else:
-            for h in hashes:
-                h['Folder'] = file_meta['Name']
-            return hashes 
+        Args:
```
unconst, line 320: Make sure there is a doc string here for each argument.

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        if check_valid(sequence_length):
+            self.sequence_length = sequence_length
+
+        if check_valid(block_size):
+            logger.warning('The block size represents the seqeunce length and will be depracted')
+            self.sequence_length = sequence_length
+    
+        if check_valid(block_size_bytes):
+            self.block_size_bytes = block_size_bytes
+
```
unconst: Extra lines like this should be removed. NIT

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+            buffer_size(int, optional):
+                The size of the buffer. 
         """
-        def get_hashes(dataset_meta):
-            if self.IPFS_fails > self.IPFS_fails_max:
-                sub_directories = json.loads(self.load_hash(dataset_meta))
-                for sub_directory in sub_directories:
-                    sub_directory['Folder'] = dataset_meta['Name']
-            else:
-                sub_directories = self.get_dataset(dataset_meta)
```
unconst, line 333: Extra line spaces like this not required. NIT

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+            buffer_size (int, required):
+                The size of the sample buffer.
 
-        Returns:
-            directory: Map{ Name: str, Hash: str, Size: int }: 
-                A random directory that lead to a datafile.
         """
-        # --- If the size of directory is small, it is leads to data file, return the data file.
-        if file_meta['Size'] <= self.datafile_size_bound:
-            return file_meta
```
unconst: Same here

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        '''
+        params = dict(arg=cid, offset=offset)
+        params['length'] = length
+        headers = {}
+        response = await self.api_post('cat', params=params, headers=headers, chunk_size=10000000, num_chunks=1)
+        return response
 
-            block_size(int, required):
-                The block_size of data that should be produced by dataloader. 
+    async def get_folder_text_hashes(self, file_meta:dict, dataset:str, max_chunks:int = 1, chunk_size:int = 100000000) -> List[Dict[str, Union[str, int]]]:
```
unconst: Line length

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
 
-        # empty the queue
-        while not self.data_queue.queue.empty():
-            self.data_queue.queue.get()
+            data = await self.cat(file_meta['Hash'], offset=chunk_i*chunk_size ,length=chunk_size)
 
-        # empty the dataset_iterator with the old sizing
-        self.__infinite_dataset_iterator = iter([])
+            hashes = ['['+h + '}]'for h in data.decode().split('},')]
+
```
unconst: Unneeded extra line breaks

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
+        Get Links from file_meta
+
+        Args
+            file_meta (dict): 
+                Dictionary containing hash and name of root link.
+
+        Returns 
+            response_links (List[dict])
+                The response links from the get call.
+
```
unconst: Unneeded

bittensor/_dataset/dataset_impl.py
```diff
@@ -1,678 +1,772 @@
...
 
-        logger.success(f"Updated data size: batch_size: {old_batch_size} --> {self.batch_size}, block_size: {old_block_size} --> {self.block_size}")
+        '''
+        response = await self.api_post( 'object/get',  params={'arg': file_meta['Hash']}, return_json= True)
+        response_links = response.get('Links', [])
+        return response_links
 
-    def dataloader(self, epoch_length = 100):
-        r""" Creates a torch dataloader out of a subclass of this class.
+    async def api_post(self, endpoint:str, params:Optional[Dict[str, Any]] = {}, headers:Optional[Dict[str, Any]] = {}, return_json:Optional[bool] = False,  content_type:Optional[str] = None, chunk_size:Optional[int] = 1024, num_chunks:Optional[int] = None, sock_connect:Optional[int]=20, sock_read:Optional[int]=20) -> Union[Dict, 'aiohttp.Response', bytes]:
```
unconst: Too long.

unconst (changes requested):

bittensor/_dataset/dataset_impl.py
21bloc: I get the double spaces, what do you mean by extra lines? Like double lines between code lines?

PR #982, opened 2022-11-08
Sign responses

unconst:
This PR includes a PR for response message signing. 

Response message signatures are generated like:

my_wallet.hotkey.keypair.sign(my_wallet.hotkey.keypair.ss58_address + str(TensorMessage.tensors[0]) + str(TensorMessage.tensors[1]) + ...  )

Receivers check like:
keypair( TensorMessage.hotkey ).verify( TensorMessage.hotkey + str(TensorMessage.tensors[0]) + str(TensorMessage.tensors[1]) + ... , TensorMessage.signature )

The approach solves man-in-the-middle attacks. However there are potential repeat attacks here if the endpoint is not creating unique response tensors (since the adversary can remember the signatures) however this is likely intractable given the number of unique batch combinations and inputs to the endpoint.


camfairchild (changes requested):
Add a test for this and then LGTM

robertalanm (approved):
✅

bittensor/_axon/axon_impl.py
```diff
@@ -127,6 +127,21 @@ def __str__(self) -> str:
...
+                response_tensors (:obj:`List[bittensor.proto.Tensor`, `required`): 
+                    response_tensors to be signed            
+            Returns:
+                signature (str, required): 
+                    The string signature generated by signing the response tensors.
+        """
+        signature_messages = self.wallet.hotkey.ss58_address
+        if response_tensors != None:
+            for resp_tensor in response_tensors:
+                signature_messages += resp_tensor.serializeToString()
```
Eugene-hu, line 142: This will lead to a significant increase in network bandwidth, no? It will almost double the amount of bandwidth per message.

bittensor/_proto/bittensor.proto
```diff
@@ -92,6 +92,10 @@ message TensorMessage {
 	// Synapses hold function information this tells the axon how to use the tensor inputs.
 	// i.e. where to send them
  	repeated Synapse synapses = 9;
+
+	// Message Signature generated as: wallet.hotkey.keypair.sign( wallet.hotkey.keypair.ss58_address + str(TensorMessage.tensors[0]) + str(TensorMessage.tensors[1]) + ...  )
+	// Receivers check vis a via: keypair( TensorMessage.hotkey ).verify( TensorMessage.hotkey + str(TensorMessage.tensors[0]) + str(TensorMessage.tensors[1]) + ... , TensorMessage.signature )
+ 	string signature = 10;
```
Eugene-hu, line 98: Is this built yet? Also, will this allow for backward compatibility with previous versions?

bittensor/_axon/axon_impl.py
Eugene-hu: Would it be better if the signature message were the hotkey of the recipient?

Eugene-hu (commented):
Left you a few questions

unconst:
bunk

PR #983, opened 2022-11-09
Add signature v2 format

adriansmares:
#### Summary

References https://github.com/opentensor/bittensor/pull/976

This PR introduces a new signature format for requests, in order to avoid situations in which validator requests are fulfilled by nodes which are not targeted by the RPC.

This PR is based on #976 in order to avoid merge conflicts - only the last two commits are relevant here.

#### Changes

- Remove the static header check. Receptors will still keep adding it, but it is ignored from now on.
- Add `v2` signature format, which also takes into account the target `axon` hotkey. The `v2` signature ensures that the signature cannot be faked by an intermediary that is not a validator.
- Ensure that nonces cannot be replayed by disallowing equality. Allowing nonce equality renders the nonce moot.
- On chain parameters of an axon are now always updated. Previously they would be updated only on IP/port changes.

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -431,10 +431,6 @@ class AuthInterceptor(grpc.ServerInterceptor):
             # Nonces must be strictly monotonic over time.
             if nonce <= previous_nonce:
                 raise Exception("Nonce is too small")
-            if not keypair.verify(message, signature):
-                raise Exception("Signature mismatch")
-            self.nonces[endpoint_key] = nonce
-            return
 
         if not keypair.verify(message, signature):
             raise Exception("Signature mismatch")
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -429,7 +429,7 @@ class AuthInterceptor(grpc.ServerInterceptor):
         if endpoint_key in self.nonces.keys():
             previous_nonce = self.nonces[endpoint_key]
             # Nonces must be strictly monotonic over time.
-            if nonce - previous_nonce <= -10:
+            if nonce <= previous_nonce:
                 raise Exception("Nonce is too small")
             if not keypair.verify(message, signature):
                 raise Exception("Signature mismatch")
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -156,7 +156,7 @@ class axon:
             thread_pool = futures.ThreadPoolExecutor( max_workers = config.axon.max_workers )
         if server == None:
             server = grpc.server( thread_pool,
-                                  interceptors=(AuthInterceptor(blacklist=blacklist),),
+                                  interceptors=(AuthInterceptor(wallet=wallet, blacklist=blacklist),),
                                   maximum_concurrent_rpcs = config.axon.maximum_concurrent_rpcs,
                                   options = [('grpc.keepalive_time_ms', 100000),
                                              ('grpc.keepalive_timeout_ms', 500000)]
@@ -341,19 +341,26 @@ class axon:
 class AuthInterceptor(grpc.ServerInterceptor):
     """Creates a new server interceptor that authenticates incoming messages from passed arguments."""
 
-    def __init__(self, blacklist: List = []):
+    def __init__(
+        self,
+        wallet: "bittensor.Wallet",
+        blacklist: Callable = None,
+    ):
         r"""Creates a new server interceptor that authenticates incoming messages from passed arguments.
         Args:
+            wallet(bittensor.Wallet):
+                the wallet of the receiver which runs the interceptor
             black_list (Function, `optional`):
                 black list function that prevents certain pubkeys from sending messages
         """
         super().__init__()
         self.nonces = {}
         self.blacklist = blacklist
+        self.wallet = wallet
 
     def parse_legacy_signature(
         self, signature: str
-    ) -> Union[Tuple[int, str, str, str], None]:
+    ) -> Union[Tuple[int, str, str, str, int], None]:
         r"""Attempts to parse a signature using the legacy format, using `bitxx` as a separator"""
         parts = signature.split("bitxx")
         if len(parts) < 4:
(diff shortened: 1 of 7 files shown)


camfairchild:
why not just up the network version itself?

edit: then use the version to decide on the parse method

adriansmares:
> why not just up the network version itself?
> 
> edit: then use the version to decide on the parse method

That wouldn't be backward compatible - new dendrites cannot talk to old axons (since old axons expect the old format). In this way there can be a transition period in which new dendrites can speak with old axons, and old axons allow traffic from old dendrites. In some future version, the `v1` signature format can be completely removed and only `v2` would remain.

camfairchild:
> > why not just up the network version itself?
> > 
> > edit: then use the version to decide on the parse method
> 
> That wouldn't be backward compatible - new dendrites cannot talk to old axons (since old axons expect the old format). In this way there can be a transition period in which new dendrites can speak with old axons, and old axons allow traffic from old dendrites. In some future version, the `v1` signature format can be completely removed and only `v2` would remain.

I'm not suggesting you get rid of this. I'm suggesting you use the wire protocol version to do this instead of adding metadata

adriansmares:
> I'm not suggesting you get rid of this. I'm suggesting you use the wire protocol version to do this instead of adding metadata

So I've ended up doing this based on the advertised version on chain. The dendrite/receptor decides based on the endpoint version which signature to use. 

Note that this is a bit awkward right now - we cannot use version `3.4.2` as a version cutoff point, since `3.4.2` axons cannot parse this signature, but also  `nobunaga` is still on version `3.4.2`, so in order to test this, one must manually bump his `__version_as_int__`. Unit tests do try `3.4.3` as a receiving version however.

unconst (approved):
LGTM

unconst:
I think we can safely begin the review process on this PR. @eduardogr I would like your take as well.

bittensor/_axon/__init__.py
```diff
@@ -339,101 +339,147 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
+        else:
+            raise Exception("Invalid signature version")
+        # Build the key which uniquely identifies the endpoint that has signed
+        # the message.
+        endpoint_key = f"{sender_hotkey}:{receptor_uuid}"
+
+        if endpoint_key in self.nonces.keys():
+            previous_nonce = self.nonces[endpoint_key]
+            # Nonces must be strictly monotonic over time.
+            if nonce <= previous_nonce:
```
Eugene-hu, line 432: The previous nounce was to allow for asynchronous multi-calls to arrive within a small error. I am not against this change, however it should be more strict. The nounce should always increase in this case

bittensor/_axon/__init__.py
```diff
@@ -339,101 +339,147 @@ def check_forward_callback( forward_callback:Callable, synapses:list = []):
...
-        self._valid_metadata = ('rpc-auth-header', key)
-        self.nounce_dic = {}
-        self.message = 'Invalid key'
+        self.nonces = {}
         self.blacklist = blacklist
-        def deny(_, context):
-            context.abort(grpc.StatusCode.UNAUTHENTICATED, self.message)
-
-        self._deny = grpc.unary_unary_rpc_method_handler(deny)
+        self.wallet = wallet
```
Eugene-hu: I am against putting the whole wallet into server interceptor. The base wallet class contains far too many functionalities and connections for a lightweight interceptor. Instead, this should be an ss58 key.

bittensor/_receptor/receptor_impl.py
```diff
@@ -123,20 +123,32 @@ def __del__ ( self ):
...
+    def sign_v2(self):
+        nonce = f"{self.nonce()}"
+        sender_hotkey = self.wallet.hotkey.ss58_address
+        receiver_hotkey = self.endpoint.hotkey
+        message = f"{nonce}.{sender_hotkey}.{receiver_hotkey}.{self.receptor_uid}"
+        signature = f"0x{self.wallet.hotkey.sign(message).hex()}"
+        return ".".join([nonce, sender_hotkey, signature, self.receptor_uid])
+
+    def sign(self):
+        if self.endpoint.version >= bittensor.__new_signature_version__:
```
Eugene-hu, line 144: I like this design, really nice use of the versioning from the endpoint information. However, we have to make sure that the version information is updated properly on the chain.

Eugene-hu: @camfairchild Do you know if the version information from the chain is updated based on every extrinsic or if it is updated based on the time of registration?

Eugene-hu (changes requested):
Nice job! I like the new signature function. There are a few things that we need to fix before merging

bittensor/_axon/__init__.py
adriansmares: But this is what this check does, right ? If the nonce is smaller or equal to the last nonce, the validation fails (an exception is raised).

Regarding the parallel calls - this is why I've moved nonce generation to use the nanosecond monotonic counters. That should suffice even for high parallelism.

Also please note that it is never correct to allow these checks to have these 'margins' without some memory of the past nonces. The previous form of the check allowed the last request (or more generally, all of the requests in the window) to be replayed indefinitely.

bittensor/_receptor/receptor_impl.py
adriansmares: The on-chain/metagraph information is updated on serve, which miners do usually do after starting their axon:
https://github.com/opentensor/bittensor/blob/f5c5e1de4bd1827c639942f702e683d8c9bfb791/bittensor/_subtensor/subtensor_impl.py#L684-L691

Edit: but maybe then the axon serving check should be relaxed a bit above:
https://github.com/opentensor/bittensor/blob/f5c5e1de4bd1827c639942f702e683d8c9bfb791/bittensor/_subtensor/subtensor_impl.py#L674-L678
Otherwise the version won't be updated unless the IP and port change.

unconst: It is updated when calling the serve function which is only done at startup

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -671,24 +671,29 @@ def serve (
...
-        with bittensor.__console__.status(":satellite: Checking Axon..."):
-            neuron = self.neuron_for_pubkey( wallet.hotkey.ss58_address )
-            if not neuron.is_null and neuron.ip == net.ip_to_int(ip) and neuron.port == port:
-                bittensor.__console__.print(":white_heavy_check_mark: [green]Already Served[/green]\n  [bold white]ip: {}\n  port: {}\n  modality: {}\n  hotkey: {}\n  coldkey: {}[/bold white]".format(ip, port, modality, wallet.hotkey.ss58_address, wallet.coldkeypub.ss58_address))
-                return True
-
-        ip_as_int  = net.ip_to_int(ip)
-        ip_version = net.ip_version(ip)
-
-        # TODO(const): subscribe with version too.
```
adriansmares, line 683: Is this TODO still relevant ?

bittensor/_axon/__init__.py
Eugene-hu: Yea, I agree; the previous implementation was problematic. The issue is that message routing differences between parallel calls resulted in incorrect ordering as the messages arrived.  This meant that some calls were rejected due to an incorrect nounce. 

Anyways, my issue is that the nounce shouldn't be equal to the last nounce. It should only increase from message to message. I don't see a situation where the nanosecond nounce would be identical to a previous message.

adriansmares: But again, this is what the check does ? I don't see where you are drawing the conclusion that equal nonces are allowed by this check.

bittensor/_receptor/receptor_impl.py
Eugene-hu: Yea, the issue is that axon serving extrinsic is only submitted if the neuron IP or port has been changed. For older miners, the version information can be quite outdated. Instead, servers should submit a new extrinsic if the version information has also been changed

bittensor/_axon/__init__.py
Eugene-hu: Ahh, my bad there, thats the criteria for rejection. Makes much more sense now.  Sorry about that confusion haha

bittensor/_receptor/receptor_impl.py
adriansmares: https://github.com/opentensor/bittensor/pull/983/commits/0a91336fc8390570a10fc2ca03af5fd68df4825f should make the serve updates apply on every parameter change.

bittensor/_subtensor/subtensor_impl.py
Eugene-hu: Remove it, this was already done

bittensor/_receptor/receptor_impl.py
Eugene-hu: I see, you are checking all of the neuron params. Nice change!

Eugene-hu (approved):
LGTM, Nice job!

bittensor/_axon/__init__.py
```diff
@@ -367,52 +372,71 @@ def parse_legacy_signature(
...
+        parts = signature.split(".")
+        if len(parts) != 4:
+            return None
+        try:
+            nonce = int(parts[0])
+        except ValueError:
+            return None
+        sender_hotkey = parts[1]
+        signature = parts[2]
+        receptor_uuid = parts[3]
```
eduardogr, line 392: Regarding maintainability, 
what about having a separated and well tested function that give us an object with nonce, sender_hotkey, signature and receptor_uuid?

This also will help us to localize when some error occurs because of a signature not having the 4 parts (len(parts) != 4) and not mixing all the logic in the same place.

What do you think about that?

bittensor/_axon/__init__.py
```diff
@@ -367,52 +372,71 @@ def parse_legacy_signature(
...
+        return (nonce, sender_hotkey, signature, receptor_uuid, 2)
```
eduardogr, line 393: Continue of the previous comment... since you finally send this tuple, I identify this as a smell that a parser/helper functions/ or some kind of collaborator piece of code could generate this object

bittensor/_axon/__init__.py
eduardogr: I just realized that I was proposing something that already exists haha Sorry for that

I resolved this

bittensor/_axon/__init__.py
eduardogr: As i said... 
I just realized that I was proposing something that already exists haha Sorry for that

I resolved this

bittensor/_receptor/receptor_impl.py
```diff
@@ -123,20 +123,32 @@ def __del__ ( self ):
...
+
+    def sign_v2(self):
+        nonce = f"{self.nonce()}"
+        sender_hotkey = self.wallet.hotkey.ss58_address
+        receiver_hotkey = self.endpoint.hotkey
+        message = f"{nonce}.{sender_hotkey}.{receiver_hotkey}.{self.receptor_uid}"
+        signature = f"0x{self.wallet.hotkey.sign(message).hex()}"
+        return ".".join([nonce, sender_hotkey, signature, self.receptor_uid])
+
+    def sign(self):
```
eduardogr, line 143: I am missing unit tests for this function, can you add them?

eduardogr (approved):
LGTM,
(commented already, i miss a couple of unit tests but everything good)

Thanks for this!

bittensor/_receptor/receptor_impl.py
adriansmares: So this should be covered by `test_axon_receptor_connection_forward_works` - it varies between the old endpoint version (which uses `sign_v1`), new endpoint version (which uses `sign_v2`) and the current version (just future proofing).

It looks like I accidentally broke the test name while adding this matrix test - should be fixed now.

For that matter - do you have a line coverage overview somewhere ? I couldn't find it in the CircleCI runs.

eduardogr: commented offline,

- no coverage available right now.
- a dedicated unit test for this will help us localize the failure if that happens and ensure that we have the code covered by a tests. Unit tests coverage should always be specific pieces of code that cover the code we added and not a side effect of a bigger test

Eugene-hu:
@adriansmares Merging this PR today, it will be released in the next version

PR #985, opened 2022-11-09
Docker Action: dont run tests before docker image push

camfairchild:
This PR makes the docker image push action not run tests first, as these should be completed anyway before merge to master occurs

### DIFF
--- a/.github/workflows/docker_image_push.yml
+++ b/.github/workflows/docker_image_push.yml
@@ -1,7 +1,6 @@
-# This workflow will install Python dependencies, run tests and lint with a single version of Python
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+# This workflow with build and push a Docker image dockerhub on every push to the master branch
 
-name: Pushing Image to Docker
+name: Pushing Latest Release Image to Docker
 
 on: 
   push:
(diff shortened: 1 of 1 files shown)


shibshib:
Hm, not sure if we should take this out. The point of testing one more time is basically for sanity -- what if two PRs were pushed at the same time? Just because tests pass before merging to master doesn't mean master is in the clear just yet.

camfairchild:
> Hm, not sure if we should take this out. The point of testing one more time is basically for sanity -- what if two PRs were pushed at the same time? Just because tests pass before merging to master doesn't mean master is in the clear just yet. 

my main issue is that the pytest tests don't run properly*.

Does circleci run them after merge too?

eduardogr:
> > Hm, not sure if we should take this out. The point of testing one more time is basically for sanity -- what if two PRs were pushed at the same time? Just because tests pass before merging to master doesn't mean master is in the clear just yet.
> 
> my main issue is that the pytest tests don't run properly*.
> 
> Does circleci run them after merge too?

@shibshib @camfairchild Since we are moving the release of the docker image to a CircleCI workflow, I am proposing removing this file, in the PR https://github.com/opentensor/bittensor/pull/1011.

Having said that this CircleCI release workflow will run the tests aswell before releasing.

Are we align with these actions?

- closing this PR, since we are removing this Github action
- CircleCI workflow to release and running the needed checks before release, of course

PR #995, opened 2022-11-15
[Feature] mention balance if not no prompt

camfairchild:
This PR adds a balance check and display during a transfer (without `--no_prompt`)  
image  
This should improve UX as a user can know how much TAO they have before sending


bittensor/_cli/__init__.py
```diff
@@ -742,15 +742,27 @@ def check_transfer_config( config: 'bittensor.Config'):
...
-                console.print(":cross_mark:[red] Invalid Tao amount[/red] [bold white]{}[/bold white]".format(amount))
-                sys.exit()
+            if not config.no_prompt:
+                amount = Prompt.ask("Enter TAO amount to transfer")
+                try:
+                    config.amount = float(amount)
+                except ValueError:
+                    console.print(":cross_mark:[red] Invalid TAO amount[/red] [bold white]{}[/bold white]".format(amount))
+                    sys.exit()
+            else:
```
Eugene-hu, line 763: What is this `else` for? if config.no_prompt == True?

camfairchild: It's nested. It would be for `config.get('amount') is None and config.no_prompt == True`

Eugene-hu: I see, so it errors out if the user doesn't specify an amount and picks no_prompt?

camfairchild: exactly

Eugene-hu (approved):
LGTM

PR #996, opened 2022-11-15
[Fix] fix max stake for single

camfairchild:
This PR addresses a bug with max stake that effected staking with only a single hotkey `--wallet.hotkey hk0` using the `--max_stake <#>` flag.  

e.g. `btcli stake --wallet.hotkey hk0 --max_stake 1.0`


bittensor/_cli/cli_impl.py
```diff
@@ -347,9 +347,7 @@ def stake( self ):
                 bittensor.wallet( config = self.config, hotkey = hotkey ) for hotkey in self.config.wallet.get('hotkeys')
             ]
         else:
-            # Do regular stake
-            subtensor.add_stake( wallet, amount = None if self.config.get('stake_all') else self.config.get('amount'), wait_for_inclusion = True, prompt = not self.config.no_prompt )
-            return None
+            wallets_to_stake_to = [ bittensor.wallet( config = self.config ) ]
```
Eugene-hu, line 353: I am confused about the previous implementation. So the `else` only covers the case in which there is a single wallet to stake too?  If so, we should be more specific.

bittensor/_cli/cli_impl.py
```diff
@@ -384,6 +383,11 @@ def stake( self ):
             final_amounts.append(stake_amount_tao)
             final_wallets.append(wallet)
 
+        if len(final_wallets) == 0:
+            # No wallets to stake to.
+            bittensor.__console__.print("Not enough balance to stake to any hotkeys or max_stake is less than current stake.")
```
Eugene-hu, line 391: The console error should be updated I think

bittensor/_cli/cli_impl.py
```diff
@@ -374,7 +372,8 @@ def stake( self ):
             stake_amount_tao: float = self.config.get('amount')
             if self.config.get('max_stake'):
                 wallet_stake: Balance = wallet.get_stake()
-                stake_amount_tao: float =  self.config.get('max_stake') - wallet_stake.tao
+                stake_amount_tao: float = self.config.get('max_stake') - wallet_stake.tao
```
Eugene-hu, line 378: I see, so previously, `max_stake` never worked if there was only a single wallet

Eugene-hu (approved):
Nice job!

bittensor/_cli/cli_impl.py
camfairchild: Hotkeys (wallets) are removed when we don't have enough balance to stake the amount to them.

PR #997, opened 2022-11-16
Add Mask to TextLastHiddenState

unconst:
Adds masking to the TextLastHiddenState Synapse.

For a return tensor of size `(batch_size, sequence_length, hidden_state)`, we can now optionally pass a mask as: `bittensor.synapse.TextLastHiddenState(mask: Optional[List[int]])`. The mask will apply across the `sequence_length` dimension, i.e. if `mask == [0]`, only the hidden state for the 0th element will be filled. Explicitly, `return_tensor[:, 1:, :]` would `==0`.

Alternatively a list of non-consecutive integers can be passed, e.g. `bittensor.synapse.TextLastHiddenState(mask = [0, 5, 3])` would only return non-empty tensors for sequence indices `0`, `5`, and `3`.

This change can drastically decrease the network bandwidth.

Alternatively, the user can specify a random mask
d.text_last_hidden_state ( ... mask = random.choice( seq_len, k), ...)
This random masking can be used for adversarial resistance.  Or Jepa style training.


test.ipynb
```diff
@@ -0,0 +1,292 @@
+{
```
Eugene-hu: Please remove

bittensor/_synapse/__init__.py
```diff
@@ -224,5 +226,7 @@ def deserialize( synapse_wire_proto: bittensor.proto.Synapse ) -> Synapse:
             return TextCausalLMNext.deserialize_from_wire_proto(synapse_wire_proto)
         elif synapse_wire_proto.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_SEQ_2_SEQ:
             return TextSeq2Seq.deserialize_from_wire_proto( synapse_wire_proto )
+        elif synapse_wire_proto.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_MASKED_LAST_HIDDEN_STATE:
+            return TextMaskedLastHiddenState.deserialize_from_wire_proto( synapse_wire_proto )
```
Eugene-hu: Is TextMaskedLastHiddenState class defined anywhere? I can't seem to find it

bittensor/_synapse/__init__.py
```diff
@@ -224,5 +226,7 @@ def deserialize( synapse_wire_proto: bittensor.proto.Synapse ) -> Synapse:
             return TextCausalLMNext.deserialize_from_wire_proto(synapse_wire_proto)
         elif synapse_wire_proto.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_SEQ_2_SEQ:
             return TextSeq2Seq.deserialize_from_wire_proto( synapse_wire_proto )
+        elif synapse_wire_proto.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_MASKED_LAST_HIDDEN_STATE:
```
Eugene-hu: When do we call TEXT_MASKED_LAST_HIDDEN_STATE? Is this a seperate synapse or a special case of the TEXT_LAST_HIDDEN_STATE?

bittensor/_synapse/text_lasthiddenstate_impl.py
```diff
@@ -111,27 +143,88 @@ def check_forward_response_tensor    ( self, forward_request_tensor, forward_res
...
-   
+        if len( forward_response_tensor.shape ) == 3:
+            if ( forward_response_tensor.size(0) != forward_request_tensor.size(0) or
+                 forward_response_tensor.size(1) != forward_request_tensor.size(1) or
+                 forward_response_tensor.size(2) != bittensor.__network_dim__ ):
+                raise ValueError( "forward_response_tensor.shape must be in [{}, {}, {}], got: {} for synapse: {}".format( forward_request_tensor.size(0) , forward_request_tensor.size(1), bittensor.__network_dim__, list(forward_response_tensor.shape), self ) ) 
+        
+        elif len( forward_response_tensor.shape ) == 2:
+            mask_len = forward_request_tensor.size(1) if self.mask == [] else len(self.mask)
+            if ( forward_response_tensor.size(0) != forward_request_tensor.size(0) * mask_len or
```
Eugene-hu, line 167: Is the response tensor (batch*mask_len, network_dim)? Any reason why we want to push it down to 2 ranks instead of the (batch, mask_len, network_dim)?

Eugene-hu: On a similar note, I think there should a check if the mask > than the size of the request tensor here. For both the check_forward_request_tensor function and the check_forward_response_tensor functions
.

bittensor/_synapse/__init__.py
joeylegere: No, I think that was from a previous design and it's just an optional `mask` parameter in `TEXT_LAST_HIDDEN_STATE` now. will remove traces of this

bittensor/_synapse/__init__.py
joeylegere: removed this.

bittensor/_synapse/text_lasthiddenstate_impl.py
joeylegere: > Is the response tensor (batch*mask_len, network_dim)? Any reason why we want to push it down to 2 ranks instead of the (batch, mask_len, network_dim)?

I was on the fence about this too, and I can go either way on it. I like two dimensions here because it forces us to not make assumptions about the second dimension of the tensor. If we aren't careful and we miss a check with the mask, then we could return garbage--e.g. assuming that the `seq_len==mask_len` and not unpacking it correctly.

> On a similar note, there should a check if the mask > than the size of the request tensor here

Good idea. I've added that in the `shift_mask_based_on_shape` method.

Eugene-hu (approved):

Eugene-hu (approved):
LGTM

PR #998, opened 2022-11-16
Mark registration threads as daemons

adriansmares:
#### Summary

This PR marks solver (registration) processes as daemons. This allows the main process to exit when an exception occurs (generally this happens when the `subtensor` is unreachable), instead of hanging indefinitely.

#### Changes

- Mark solver processes as daemons, in order to avoid having them get joined at the end of the program.
  - They are already joined on the 'happy' path, but they should not block program termination if an exception occurs on the main process.


camfairchild (approved):
LGTM. Thanks for the contribution

PR #999, opened 2022-11-18
[BIT-598] Validator debug response table

opentaco:
### [BIT-598] Validator debug response table

Adds an additional debug table to `core_validator... --logging.debug` that samples query tasks and server responses, formats it compactly, and displays the topk predicted phrases with probabilities.

The intended use of this table is to monitor and confirm proper query handling by displaying actual response content. This can also help to detect response anomalies that can help identify various new code/system/network errors that may have been introduced.

Apart from a task selection the table also includes overall step statistics for responsive servers.

Rich MarkupError exceptions are infrequently caught and handled when special characters in the task representation corrupt added markup.

Screenshot 2022-11-18 at 12 32 39

Screenshot 2022-11-18 at 12 39 22

[BIT-598]: https://opentensor.atlassian.net/browse/BIT-598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -1444,9 +1444,14 @@ def response_table(batch_predictions: List, stats: Dict, sort_col: str, console_
         return
     batch_perm = torch.randperm(batch_size)  # avoid restricting observation to predictable subsets
 
-    columns = [column for column in neuron_stats_columns if column[1] in ['uid', 'loss_nxt', 'synergy_nxt']]
+    columns = [c[:] for c in neuron_stats_columns if c[1] in ['uid', 'shapley_values_nxt', 'loss_nxt', 'synergy_nxt']]
+    # === Sort rows ===
     sort = sorted([(uid, s[sort_col]) for uid, s in stats.items() if sort_col in s],
                   reverse=True, key=lambda _row: _row[1])
+    col_keys = [c[1] for c in columns]
+    if sort_col in col_keys:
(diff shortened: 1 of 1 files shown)


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -1232,6 +1233,11 @@ def _synergy(first, second, target, ext):
     logger.info(f'{str(synapse)} \t| Shapley synergy values <dim>[{time.time() - synergy_start_time:.3g}s]</dim>')
 
     if logging:
+        # === Response table ===
+        # Prints the query response table: top prediction probabilities and texts for batch tasks
+        batch_predictions = format_predictions(uids, query_responses, return_ops, inputs, validation_len, index_s)
+        response_table(batch_predictions, stats, sort_col='shapley_values_nxt', console_width=console_width)
```
Eugene-hu, line 1239: Would this be too much for the average user? Would it be beneficial to put another level of debugging?

Eugene-hu (approved):
LGTM

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Potentially yes, we could add a flag to make this optional later. Right now the log space is dominated by IPFS dataset loaded and receptor create/destroy lines which far exceed space taken up by these neuron stat overviews, so I'd say this PR adds a relatively acceptable amount of log bloat.

PR #1000, opened 2022-11-20
[BIT-599] Validator weight setting improvements

opentaco:
### [BIT-599] Validator weight setting improvements
---
### Remove responsive prioritization from validator weight calculation
Weight setting limitation based on responsive prioritization is no longer needed, so has been removed. Part of the original intent of the limitation was a chain storage concern of setting too many weights, but since the network has gained very high > 90% response rate the concern is moot.

The downside of applying the limitation is that validators with longer step size artificially set fewer weights simply because they could not query the network in a single epoch. This introduced a counterproductive weight setting variability across validators.

Responsiveness is scored in any case via a Shapley EMA zero-push penalty, so setting weights on non-responsive nodes still relay an accurate scoring.

---
### Move metagraph_sync just before weight setting
The metagraph is one epoch (~250 blocks) outdated by the time weight setting is done. This means that newly registered keys will have weights set based on the stats of the key just previously registered on the same UID.

Weights will now be set more accurately when the metagraph is updated just before weight setting, which will reset stats of a UID that has changed ownership.

---
### Add metagraph registration tracker to validator
Tracks new hotkey registrations on UIDs, and records statistics of previous hotkey.

[BIT-599]: https://opentensor.atlassian.net/browse/BIT-599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -1428,10 +1428,7 @@ def format_predictions(uids: torch.Tensor, query_responses: List[List[torch.Floa
                     phrase = topk_tokens[i]
                     phrase = phrase[phrase >= 0]  # strip negative ignore_index = -100
                     phrase_str = repr(std_tokenizer.decode(phrase))[:15]  # decode, escape and truncate
-
-                    prob = f'{topk_probs[i]:.3f}'
-                    prob = prob[1:] if prob[0] == '0' else prob[:-1]  # remove obvious leading 0
-
+                    prob = f'{topk_probs[i]:.3f}'.lstrip('0')
                     topk_predictions += f"[green]{prob}[/green]: {phrase_str} "
 
                 predictions[uid] = topk_predictions[:-1]  # strip trailing space
@@ -1511,12 +1508,12 @@ def synergy_table(stats, syn_loss_diff, sort_col, console_width):
     columns = [uid_col] + [[f'{s[0]}', '', '{:.2f}', ''] for s in sort]
     rows = [[uid_col[2].format(s[0])] +
             [('[bright_cyan]{:.2f}[/bright_cyan]' if t == s else
-              '[magenta]{:.2f}[/magenta]' if syn_loss_diff[s[0]][t[0]] > 0 else
+              '[magenta]{:.2f}[/magenta]'.lstrip('0') if syn_loss_diff[s[0]][t[0]] > 0 else
               '[dim]{:.0f}[/dim]').format(syn_loss_diff[s[0]][t[0]]) for t in sort] for s in sort]
 
     # === Synergy table ===
     table = Table(width=console_width, box=None)
-    table.title = f'[white] Synergy [/white]'
+    table.title = f'[white] Synergy table [/white] | Pairwise synergy'
     table.caption = f'loss decrease'
 
     for col, _, _, stl in columns:  # [Column_name, key_name, format_string, rich_style]
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -1508,8 +1508,8 @@ def synergy_table(stats, syn_loss_diff, sort_col, console_width):
     columns = [uid_col] + [[f'{s[0]}', '', '{:.2f}', ''] for s in sort]
     rows = [[uid_col[2].format(s[0])] +
             [('[bright_cyan]{:.2f}[/bright_cyan]' if t == s else
-              '[magenta]{:.3f}[/magenta]'.lstrip('0') if syn_loss_diff[s[0]][t[0]] > 0 else
-              '[dim]{:.0f}[/dim]').format(syn_loss_diff[s[0]][t[0]]) for t in sort] for s in sort]
+              '[magenta]{:.3f}[/magenta]' if syn_loss_diff[s[0]][t[0]] > 0 else
+              '[dim]{:.0f}[/dim]').format(syn_loss_diff[s[0]][t[0]]).replace('0.', '.') for t in sort] for s in sort]
 
     # === Synergy table ===
     table = Table(width=console_width, box=None)
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -627,11 +627,11 @@ class neuron:
         # === Reset neuron stats if uid got replaced
         for uid, old_hotkey in enumerate(old_hotkeys):
             if old_hotkey != self.neuron_hotkeys[uid]:
+                block = self.subtensor.block
                 self.neuron_register.setdefault(uid, {})  # [uid] -> dict() of blocks
-                self.neuron_register[uid][self.subtensor.block] = {'new_hotkey': self.neuron_hotkeys[uid],
-                                                                   'old_hotkey': old_hotkey}
+                self.neuron_register[uid][block] = {'new_hotkey': self.neuron_hotkeys[uid], 'old_hotkey': old_hotkey}
                 if uid in self.neuron_stats:
-                    self.neuron_register[uid][self.subtensor.block]['old_stats'] = self.neuron_stats[uid]
+                    self.neuron_register[uid][block]['old_stats'] = self.neuron_stats[uid]
                     del self.neuron_stats[uid]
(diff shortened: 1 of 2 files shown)


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -430,13 +430,19 @@ def run_epoch( self ):
...
         # normal epoch duration is blocks_per_epoch if all UIDs have been queried
         # try to query each UID at least once - assumes nucleus samples without replacement
-        # but keep maximum epoch duration at 2 * blocks_per_epoch
+        # but keep minimum epoch duration at blocks_per_epoch * 12 sec block_period
+        # in case of subtensor outage causing invalid block readings to prevent fast repeated weight setting
+        start_block = self.subtensor.block
         while (self.subtensor.block < start_block + blocks_per_epoch or
-               (self.subtensor.block < start_block + 2 * blocks_per_epoch and
-                len(epoch_queried_uids) < self.metagraph.n)):
+               time.time() - epoch_start_time < blocks_per_epoch * 12):
```
isabella618033: can change 12 into `bittensor.__blocktime__`

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -430,13 +430,19 @@ def run_epoch( self ):
...
+
+            logger.info(f'Run epoch {self.epoch} (step {epoch_steps}) while '
+                        f'({self.subtensor.block} < {start_block + blocks_per_epoch} '
+                        f'= {start_block} + {blocks_per_epoch}) or '
+                        f'({time.time() - epoch_start_time:.2f} < {blocks_per_epoch * 12})')
```
isabella618033: Also can change 12 into `bittensor.__blocktime__`

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -606,24 +610,34 @@ def run_epoch( self ):
...
     def metagraph_sync(self):
         r""" Syncing metagraph together with other metagraph-size related objects
         """
-        old_hotkeys = self.neuron_hotkeys if self.neuron_hotkeys else self.metagraph.hotkeys
-        self.metagraph.sync()
+        old_hotkeys = self.neuron_hotkeys + [] if self.neuron_hotkeys else self.metagraph.hotkeys
+        self.metagraph.sync(cached=self.config.neuron.metagraph_cached or self.subtensor.network != 'local')
         self.neuron_hotkeys = self.metagraph.hotkeys
 
+        assert len(old_hotkeys) == len(self.neuron_hotkeys), 'metagraph hotkeys length do not match'
```
isabella618033: This may break the validators when one day we decide to increase the number of peers from 4096 but we forget we have this constrain here xd

isabella618033 (commented):
Looks good!

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -606,24 +610,31 @@ def run_epoch( self ):
...
 
+        # Iterate epochs.
+        self.epoch += 1
+
     def metagraph_sync(self):
         r""" Syncing metagraph together with other metagraph-size related objects
         """
-        old_hotkeys = self.neuron_hotkeys if self.neuron_hotkeys else self.metagraph.hotkeys
-        self.metagraph.sync()
+        old_hotkeys = self.neuron_hotkeys + [] if self.neuron_hotkeys else self.metagraph.hotkeys
```
Eugene-hu, line 622: what does adding an empty list to the back of the old hotkeys do?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -606,24 +610,31 @@ def run_epoch( self ):
...
+        self.metagraph.sync(cached=self.config.neuron.metagraph_cached or self.subtensor.network != 'local')
```
Eugene-hu: So if the network isn't local, the metagraph will always be synced using the cached?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -606,24 +610,31 @@ def run_epoch( self ):
...
         self.neuron_hotkeys = self.metagraph.hotkeys
 
         changed_hotkeys = []
         # === Reset neuron stats if uid got replaced
         for uid, old_hotkey in enumerate(old_hotkeys):
             if old_hotkey != self.neuron_hotkeys[uid]:
+                block = self.subtensor.block
+                self.neuron_register.setdefault(uid, {})  # [uid] -> dict() of blocks
+                self.neuron_register[uid][block] = {'new_hotkey': self.neuron_hotkeys[uid], 'old_hotkey': old_hotkey}
```
Eugene-hu: This will continuously add more and more memory as the validator runs. I think it's beneficial, primarily to diagnose the network. Should we add a flag, so only validators that we run will store all of this information?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -669,44 +680,25 @@ def neuron_stats_update(self, neuron_stats: Dict[int, Dict[str, Any]]):
...
         # === Filter to non-zero weights ===
-        neuron_weights = neuron_weights[preferred_uids]  # rearrange neuron_weights to match preferred_uids order
-        preferred_uids = preferred_uids[neuron_weights > 0]  # filter to non-zero weights
-        neuron_weights = neuron_weights[neuron_weights > 0]  # filter to non-zero weights
-
-        # === Slice weights_to_set UIDs ===
-        weights_to_set = max([min_allowed_weights, len(responsive_uids)])
-        sample_uids = preferred_uids[:weights_to_set]  # slice to weights_to_set
-        sample_weights = neuron_weights[:weights_to_set]  # slice to weights_to_set
+        sample_uids = torch.argwhere(neuron_weights > 0).squeeze(dim=1)  # find uids with non-zero weight
```
Eugene-hu, line 706: Now that we have removed the responsive requirement, do you think we should still limit the validator to set weights once it has fully queried all of the network?

Eugene-hu (approved):
Looks good! Let you a few questions

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Just a sanity measure to ensure `old_hotkeys` is not a reference to a list that will be modified outside, but is rather a copy of the list, which is what `+ []` does.

bittensor/_neuron/text/core_validator/__init__.py
shibshib: Can't recall, does using cached mean it'll be using the metagraph copy on IPFS?

shibshib (approved):

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, that is the current validator setting to always use metagraph cache in any case, and definitely want to use cache if not local because it can take an hour to update from nakamoto otherwise. Local update without cache takes about 10 minutes, and gives a more recent state than IPFS cache which is updated ~20 mins.

isabella618033 (approved):

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Roughly seeing 100KB/day disk space added at current higher rate of deregistrations, so it is not that excessive, but refactored and added a flag `--neuron.track_hotkey_changes` default=False that makes this opt-in.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: So the epoch conditions have been changed to set weights exactly every 250 blocks, regardless of whether the network was fully queried. Validators that are unable to query the network fully in 250 blocks will at least have a partial update ready for setting weights, so the on-chain incentive calculations have a steady influx of updated weights from validators.

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: But do we have to store the hotkeys for each of the blocks? I think we can regularly prune the dictionary for blocks that is like 2 days ago

PR #1001, opened 2022-11-20
Preview the inference data and save each response to a file.

divineopera:
Displays the first and last 65 characters of the original text from requests sent to servers from validators, `text_batch`, and saves the entire last request in original text to a file, `inference_data_last.txt`, in the `core_server/` directory.

The `stdout` for the core server's `run.py` will now include this data after picking up a request from the validator and before the forward pass, resulting in the following logs from `pm2`:
image

### DIFF
--- a/bittensor/_neuron/text/core_server/nucleus_impl.py
+++ b/bittensor/_neuron/text/core_server/nucleus_impl.py
@@ -205,6 +205,15 @@ class server(torch.nn.Module):
             std_tokenizer = self.std_tokenizer
 
         text_batch = std_tokenizer.batch_decode(token_batch)  # decode tokens to original text
+
+        # displaying a preview of the inference data
+        print(('Inference Data: ' + str(text_batch)[2:65]) + '......' + (str(text_batch)[-65:-2]))
+        # saving the inference data to a file (inference_data.txt in core_server/)
+        f = open('inference_data.txt', 'a')
+        f.write(str(text_batch)[2:-2])
+        f.write('\n')
+        f.close()
+
         result = translate_special_token_text(text_batch, std_tokenizer, self.tokenizer)  # translate special tokens
         to_text_batch, from_offsets_batch, to_offsets_batch, pad_offsets_batch = result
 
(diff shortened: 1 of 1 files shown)


eduardogr:
Hi! Please, fill the PR description following the guideline that proceeds:
- https://github.com/opentensor/bittensor/blob/master/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md

bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -205,6 +205,15 @@ def remapping_token(self, token_batch, std_tokenizer=None, return_offsets_mappin
...
 
         text_batch = std_tokenizer.batch_decode(token_batch)  # decode tokens to original text
+
+        # displaying a preview of the inference data
+        print(('Inference Data: ' + str(text_batch)[2:65]) + '......' + (str(text_batch)[-65:-2]))
+        # saving the last inference data query to a file (inference_data_last.txt in core_server/)
+        f = open('inference_data_last.txt', 'w')
+        f.write(str(text_batch)[2:-2])
+        f.write('\n')
+        f.close()
```
camfairchild, line 215: We really shouldn't be writing to a file on every request. 

We should also be catching any IO errors that occur during this as otherwise it will disrupt normal function.

bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -205,6 +205,15 @@ def remapping_token(self, token_batch, std_tokenizer=None, return_offsets_mappin
             std_tokenizer = self.std_tokenizer
 
         text_batch = std_tokenizer.batch_decode(token_batch)  # decode tokens to original text
+
+        # displaying a preview of the inference data
+        print(('Inference Data: ' + str(text_batch)[2:65]) + '......' + (str(text_batch)[-65:-2]))
```
camfairchild, line 210: This whole feature should probably be a flag, and disabled by default, rather than enabled always.

camfairchild (changes requested):

camfairchild:
Closing this as I'm not sure it's a good fit for integration into `btcli`. 

Feel free to discuss this further by replying here.

PR #1002, opened 2022-11-21
No Serve

CreativeBuilds:
adds a flag `--neuron.no_serve` that stops an axon from being served for a validator. 

useful for those running validators and servers from the same hotkey and need the axon port to be used by their miner.

### DIFF
--- a/bittensor/_neuron/text/core_validator/__init__.py
+++ b/bittensor/_neuron/text/core_validator/__init__.py
@@ -233,6 +233,7 @@ class neuron:
         parser.add_argument('--neuron.print_neuron_stats', action='store_true', help='If True, print neuron_stats and exit.', default=False)
         parser.add_argument('--neuron.restart', action='store_true', help='If True, reset neuron_stats and validate anew.', default=False)
         parser.add_argument('--neuron.restart_on_failure',  action='store_true', help='''Restart neuron on unknown error.''', default=True )
+        parser.add_argument('--neuron.no_serve',  action='store_true', help='If True, neuron wont serve axon (disables prometheus)', default=True )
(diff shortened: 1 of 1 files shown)


camfairchild:
Please rebase this on staging

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -234,6 +234,7 @@ def add_args( cls, parser ):
         parser.add_argument('--neuron.print_neuron_stats', action='store_true', help='If True, print neuron_stats and exit.', default=False)
         parser.add_argument('--neuron.restart', action='store_true', help='If True, reset neuron_stats and validate anew.', default=False)
         parser.add_argument('--neuron.restart_on_failure',  action='store_true', help='''Restart neuron on unknown error.''', default=True )
+        parser.add_argument('--neuron.no_serve',  action='store_true', help='If True, neuron wont serve axon (disables prometheus)', default=True )
```
Eugene-hu, line 237: This needs to default to `False`, or else, the store_action would not do anything

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -299,6 +300,11 @@ def __enter__(self):
                 root_dir = self.config.neuron.full_path
             )
 
+        # Check if no_serve is set.
+        if self.config.neuron.no_serve:
+            print('--neuron.no_serve flag set, not serving to network. Prometheus metrics will not be available.')
```
Eugene-hu, line 305: Rather than printing a simple string, this should be a warning issued by the logger, `logger.warning`

shibshib:
Hey @CreativeBuilds can you please address comments in here or close the PR?

PR #1003, opened 2022-11-22
Improving the way we manage requirements

eduardogr:
## What are you going to find?

- Now all versions are pinned, no >= or > without an upper bound is used
- Extracting dev requirements into a new file
- Extracting cubit requirements into a new file
- Moving requirements file to a new directory 'requirements'
- New Makefile provided for development commands
  - creating a virtual env
  - removing packages from within the virtualenv
  - installing bittensor, bittensor[dev], bittensor[cubit]

### Why pinned versions?

We should know the quality of what we release so we need to know if the released version works or not.
If we let dependencies open and some update breaks the version, we are going to end up with a bittensor broken scenario. Potentially miners are going to have a different set of packages<=>version and we do not want that, we want to know as much as possible about what people is running.

### What about updating packages?

Within the proposal of this change, packages updates will be related with a bittensor new version release and, of course, a dedicated branch and time to check if the new package version integrated is working along our basecode.

## What i need form you reviewing this PR?

I need a double check that the pinned versions are the ones we want.

Thanks!

### Installing bittensor for a development environment
```
python3 -m pip install '.[dev]'
```

### Installing bittensor with cubit
```
python3 -m pip install '.[cubit]'
```


Eugene-hu (approved):
LGTM, but the testing change might have showcased some new flaky tests. We should fix them asap.

camfairchild (approved):
Looks good to me

requirements/prod.txt
```diff
@@ -0,0 +1,46 @@
+argparse==1.4.0
```
eduardogr, line 1: Are we good with these versions @Eugene-hu  @opentaco  @isabella618033 @unconst  ??

I would like to review them well so we agree this versions are expected versions for bittensor

opentaco: I checked some existing bittensor installations and most of their packages have newer versions than what is pinned here. This just means we'll have to do some more work in future to advance the pinned version safely to the latest, which may offer improved package benefits, while we ensure bittensor doesn't break.

opentaco (approved):

requirements/prod.txt
eduardogr: Great great.
I don't especially want these versions, we can change them to the ones that best fit us

PR #1007, opened 2022-11-23
Check no version checking

unconst:
Two minor changes:
-- checks for no_version_checking in config creation.
-- If not command is passed to the cli we print the help

### DIFF
--- a/bittensor/_cli/__init__.py
+++ b/bittensor/_cli/__init__.py
@@ -669,7 +669,7 @@ class cli:
 
     @staticmethod   
     def check_config (config: 'bittensor.Config'):
-        """ Check if the essential condig exist under different command
+        """ Check if the essential config exist under different command
         """
         if config.command == "run":
             cli.check_run_config( config )
--- a/bittensor/_cli/__init__.py
+++ b/bittensor/_cli/__init__.py
@@ -661,6 +661,9 @@ class cli:
             type=str, 
             required=False
         )
+        if len(sys.argv) == 1:
+            parser.print_help()
+            sys.exit()
 
         return bittensor.config( parser, args=args )
 
@@ -704,6 +707,8 @@ class cli:
             cli.check_help_config(config)
         elif config.command == "update":
             cli.check_update_config(config)
+        else:
+            config.he
 
     def check_metagraph_config( config: 'bittensor.Config'):
         if config.subtensor.get('network') == bittensor.defaults.subtensor.network and not config.no_prompt:
--- a/bittensor/_cli/__init__.py
+++ b/bittensor/_cli/__init__.py
@@ -707,8 +707,6 @@ class cli:
             cli.check_help_config(config)
         elif config.command == "update":
             cli.check_update_config(config)
-        else:
-            config.he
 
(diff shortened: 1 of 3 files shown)


bittensor/_cli/__init__.py
```diff
@@ -704,6 +707,8 @@ def check_config (config: 'bittensor.Config'):
             cli.check_help_config(config)
         elif config.command == "update":
             cli.check_update_config(config)
+        else:
```
camfairchild: ?

bittensor/_cli/cli_impl.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, config: 'bittensor.Config' ):
                 config (:obj:`bittensor.Config`, `required`): 
                     bittensor.cli.config()
         """
-        if not config.no_version_checking:
+        if "no_version_checking" in config and config.no_version_checking:
```
camfairchild: `if not config.get("no_version_checking", True):`

would be better.

camfairchild (changes requested):

bittensor/_cli/__init__.py
unconst: lmao

bittensor/_cli/cli_impl.py
unconst: Kinda, although you need to read the docs on the get function to understand the boolean

camfairchild:
Rebased this onto finney and resolved the changes.  

`btcli` with no other args will print the help text and `btcli [command] --no_version_checking` will skip version checks

camfairchild (approved):

bittensor/utils/__init__.py
```diff
@@ -66,7 +66,7 @@ def version_checking():
     latest_version_as_int = (100 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
 
     if latest_version_as_int > bittensor.__version_as_int__:
-        print('\u001b[31m Current Bittensor Version: {}, Latest Bittensor Version {} \n Please update to the latest version'.format(bittensor.__version__,latest_version))
+        print('\u001b[31mBittensor Version: Current {}/Latest {}\nPlease update to the latest version\u001b[0m'.format(bittensor.__version__,latest_version))
```
eduardogr: Should we let them know that they don't need to do it urgently?

bittensor/_cli/__init__.py
```diff
@@ -662,11 +662,16 @@ def config(args: List[str]) -> 'bittensor.config':
             required=False
         )
 
+        # If no arguments are passed, print help text.
+        if len(args) == 0:
+            parser.print_help()
+            sys.exit()
```
eduardogr, line 668: Thanks for this!

bittensor/utils/__init__.py
camfairchild: I think if we make it yellow maybe.
And yes we can add like `"Please update your client when convenient"`

camfairchild: @eduardogr how's this?

eduardogr (approved):

PR #1011, opened 2022-11-24
Removing GitHub workflow push docker

eduardogr:
- Removing github workflow for pushing docker image


camfairchild (approved):
I agree. Should be on circleCI instead

Eugene-hu (approved):
Yea, I agree. I think this consistently fails on github actions. Do we need another script to push to docker or is that depreciated?

eduardogr:
> Yea, I agree. I think this consistently fails on github actions. Do we need another script to push to docker or is that depreciated?

Great! 🚀 
@Eugene-hu In fact, i've created [this script](https://github.com/opentensor/bittensor/blob/master/scripts/release/release_docker.sh) to release the docker image

PR #1012, opened 2022-11-24
[hotfix] pin scalecodec lower

camfairchild:


Eugene-hu (approved):
Approved, do we want to pin a specific version? Or are all versions between 1.0.35 and 1.1.0 okay?

PR #1014, opened 2022-11-24
[MERGEBACK][HOTIFX] Fix/pin scalecodec

eduardogr:
3.5.1 hotfix merged back into to nobunaga

Adding upper bound for scalecodecs dependency


eduardogr:
not applies anymore

PR #1016, opened 2022-11-25
Modifying Dockerfile to build bittensor from repository version and not from github

eduardogr:
- Modifying Dockerfile to build pip package from local repository and not from github.
- Version to 3.5.1 in VERSION file and code.
- Adding circleci check for version update check in release and hotfix branches

After this PR, no version 3.5.2 will generated but a rebuilt of the 3.5.1 docker image will be done.


unconst (approved):

unconst:
LGTM

.circleci/config.yml
```diff
@@ -123,3 +135,11 @@ workflows:
...
           #requires:
             #- build-and-test
+
+  check-version-updated:
+    jobs:
+      - check-version-updated:
+          filters:
+            branches:
+              only:
+                - /^(release|hotfix)/.*/
```
Eugene-hu, line 145: So all branches with /release/ or /hotfix/ in its name will have to update their version?

Eugene-hu (commented):
LGTM

Eugene-hu (approved):

.circleci/config.yml
eduardogr: yes, that's the idea.
branches like:
  - release/*
  - hotfix/*
  
will add all we need to release, so when branches arrive the default branch we can trigger the release

eduardogr: this is to ensure that we are going to have the version update in files when release without need to push to master (we do not want that) or a new PR

camfairchild (approved):
LGTM

PR #1022, opened 2022-12-01
[BIT-601] Scaling law on EMA loss

opentaco:
### [BIT-601] Scaling law on EMA loss

The neural language model scaling law [1] is typically meant to be computed on a loss averaged over the entire training data. Currently it is computed within-batch only, which frequently sees losses below 1.69 (the natural entropy of text).

Here we now compute the scaling law and the resultant effective number of model parameters on the exponentially moving average loss for a server, which should greatly improve the definition of the result.

[1] (OpenAI scaling laws) Kaplan, Jared, et al. "Scaling laws for neural language models." arXiv:2001.08361 (2020)

[BIT-601]: https://opentensor.atlassian.net/browse/BIT-601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -739,21 +770,25 @@ def weights_table(self, sample_uids, sample_weights, include_uids=None, num_rows
...
                 _neuron_stats[uid]['weight'] = weight
+                uid_weights += [(uid, weight)]
             else:
                 unvalidated += [uid]
 
-        avail_include_uids = None
         if include_uids is not None and num_rows is not None:
-            avail_include_uids = list(set(_neuron_stats.keys()) & set(include_uids))  # exclude include_uids with no stats
+            sorted_uids = sorted(uid_weights, key=lambda tup: tup[1])
+            top_bottom_uids = [_uid for _uid, _ in sorted_uids[:5] + sorted_uids[-10:]]
```
Eugene-hu, line 789: Is there a reason we want to include the bottom uids in the weights table?

Eugene-hu (approved):
LGTM

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Viewing the bottom 5 weights in the validator debug weight table is useful to confirm the exclusion quantile and associated Shapley values, losses, base parameters, and synergy values, which miners can use to choose models for newly registered servers.

opentaco:
### Comparative analysis of BIT-601

We compare the neuron stats of two types of nakamoto validators, one on current master branch, and the other on the BIT-601 branch. The change from master is that BIT-601 now applies the scaling law on the average loss across multiple batches, instead of on each batch_loss separately like master.

**master**: `avg(scaling_law(batch_loss))`
```python
# estimate the effective number of model parameters from the batch_loss
_num_params = scaling_law_loss_to_params(_loss)
```

**BIT-601**: `scaling_law(avg(batch_losses))`
```python
# estimate the effective number of model parameters from EMA loss
_num_params = scaling_law_loss_to_params(torch.tensor(stats['loss_nxt']))
```
We expect a change in `base_params_nxt` and `shapley_values_nxt` from master to BIT-601. We measure the probability distribution shift and see that overall BIT-601 has smaller values.
1. Generally `avg(batch_losses) > 1.69`, because over a lot of batches we start to be limited by the natural entropy of text, which means that BIT-601 will rarely clamp in the scaling law.
2. Frequently `batch_loss < 1.69`, because small individual batches on fine-tuned data can often get the next token prediction correct, which means that master often clamps the scaling law.
3. Generally when `batch_loss < avg(batch_losses)`, then despite clamping on `batch_loss` this inequality still seems to be the case in the result obtained where `base_params_nxt` is then larger in master compared to BIT-601.
4. However, since less clamping happens in BIT-601, we expect a more accurate scoring and separation of model sizes because the noise due to clamping is removed, which means that larger models will be able to separate themselves better.


Screenshot 2022-12-08 at 19 36 47

5. With BIT-601 we observe improved penetration into higher base parameter counts, and we also see less noise between the `loss_nxt` and `base_params_nxt` relationship, compared to master.
6. A small minority of BIT-601 parameter counts are clamped internally at a loss of 1.69, which is relatively acceptable to prevent outliers from getting extremely large parameter counts occasionally when solving tasks they've been fine-tuned on.
7. The EMA window is not really wide enough, and the current task base not novel enough, to naturally achieve the 1.69 limit, so it has to be enforced.

Screenshot 2022-12-09 at 11 54 14

PR #1023, opened 2022-12-06
Release 3.6.0

eduardogr:
Release branch for bittensor version 3.6.0

Here we will be adding squashed commits to be released. They will be added next to their PR

Changes between this branch and master: https://github.com/opentensor/bittensor/compare/master..release/3.6.0

Since the last release I merged the release branch squashing the commits, the three-dot diff seen here is not the same than the 2-dot diff. So, we can review the 2-dot diff i've shared before this line to review the release.

PRs included:
- https://github.com/opentensor/bittensor/pull/1020
- https://github.com/opentensor/bittensor/pull/1011
- https://github.com/opentensor/bittensor/pull/995
- https://github.com/opentensor/bittensor/pull/996
- https://github.com/opentensor/bittensor/pull/983
- https://github.com/opentensor/bittensor/pull/1003


Eugene-hu (approved):

setup.py
```diff
@@ -48,7 +60,12 @@
     include_package_data=True,
     author_email='',
     license='MIT',
-    install_requires=install_requires,
+    python_requires='>=3.7',
+    install_requires=requirements,
+    extras_requires={
+        'cubit': extra_requirements_cubit,
```
camfairchild: Does this work? Usually you can't use git packages as extra
iirc pypi will complain

camfairchild: @eduardogr

camfairchild (changes requested):
See comment on setup.py @eduardogr

setup.py
eduardogr: Thanks for pointing that, I'm checking that...

eduardogr: removed till we know how we want to add this extra

camfairchild (approved):

PR #1026, opened 2022-12-09
neuron.ip_version fix

Eugene-hu:
subtensor.serve was incorrectly calling neuron.ip_version instead of neuron.ip_type


camfairchild (approved):
LGTM

eduardogr (approved):

PR #1027, opened 2022-12-10
[BIT-602] Update scaling power from subtensor

opentaco:
### [BIT-602] Update scaling power from subtensor

Since `self.config.nucleus.scaling_law_power` is updated from default -1 at nucleus init, the condition here at epoch start needs to be removed and has to update from subtensor always.

[BIT-602]: https://opentensor.atlassian.net/browse/BIT-602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Eugene-hu (approved):
LGTM

PR #1029, opened 2022-12-13
Integration dendrite test fixes

Eugene-hu:
Parallel testing in circle ci leads to a flaky test in the dendrite. This PR fixes the issue along with an issue in the speed. The test was improperly pulling data when it was not supposed to.


eduardogr (approved):
Nice catch @Eugene-hu  !! Thanks for this!

PR #1030, opened 2022-12-13
notify user that hotkey is not registered

camfairchild:
This PR notifies the user if the hotkey is not registered during `btcli stake` and `btcli unstake` calls


camfairchild:
Closing to properly rebase on finney

Eugene-hu (approved):
LGTM

PR #1031, opened 2022-12-13
Fixes for get_external_ip

Eugene-hu:
External API was causing longer than expected unit tests. This is the first part of many to fix the issues

### DIFF
--- a/bittensor/utils/networking.py
+++ b/bittensor/utils/networking.py
@@ -19,7 +19,7 @@
 
 import os
 import urllib
-
+import json
 import miniupnpc
 import netaddr
 import requests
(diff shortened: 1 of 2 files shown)


bittensor/utils/networking.py
```diff
@@ -92,27 +92,29 @@ def get_external_ip() -> str:
...
 
-    # --- Try ipify
+    # --- Try AWS
     try:
-        external_ip = requests.get('https://api.ipify.org').text
+        external_ip = requests.get('https://checkip.amazonaws.com').text.strip()
         assert isinstance(ip_to_int(external_ip), int)
         return str(external_ip)
     except Exception:
         pass
```
eduardogr: is this already ordered? Or should we have AWS as the first option?

Eugene-hu: Not yet, it should be AWS -> ifconfig -> ipinfo. AWS should have the most consistency.

eduardogr:
correct me if i'm wrong, but i think this branch probably has to be merged in the new staging branch

eduardogr (approved):

PR #1033, opened 2022-12-14
add explorer urls to init and check before print

camfairchild:
This PR changes the explorer url output to only output when the network explorer is known in our map of explorer urls

### DIFF
--- a/bittensor/_subtensor/subtensor_impl.py
+++ b/bittensor/_subtensor/subtensor_impl.py
@@ -1129,8 +1129,13 @@ To run a local node (See: docs/running_a_validator.md) \n
                     bittensor.__console__.print(":white_heavy_check_mark: [green]Finalized[/green]")
                     block_hash = response.block_hash
                     bittensor.__console__.print("[green]Block Hash: {}[/green]".format( block_hash ))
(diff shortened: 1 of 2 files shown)


bittensor/_subtensor/subtensor_impl.py
```diff
@@ -1129,8 +1129,13 @@ def transfer(
                     bittensor.__console__.print(":white_heavy_check_mark: [green]Finalized[/green]")
                     block_hash = response.block_hash
                     bittensor.__console__.print("[green]Block Hash: {}[/green]".format( block_hash ))
-                    explorer_url = "https://explorer.nakamoto.opentensor.ai/#/explorer/query/{block_hash}".format( block_hash = block_hash )
-                    bittensor.__console__.print("[green]Explorer Link: {}[/green]".format( explorer_url ))
+
+                    if self.network in bittensor.__network_explorer_map__:
+                        # Get explorer link if we are on a known network.
+                        explorer_root_url = bittensor.__network_explorer_map__[self.network]
+                        explorer_url = "{root_url}/#/explorer/query/{block_hash}".format( root_url=explorer_root_url, block_hash = block_hash )
```
eduardogr, line 1136: what about testing this using different network_explorer maps? So we test what we're doing here

camfairchild:
Close in favour of #1037

PR #1034, opened 2022-12-21
[BIT-612] Validator robustness improvements

opentaco:
### [BIT-612] Validator robustness improvements

Adds logits divergence statistics and context pruning to `core_validator` to improve validation robustness by evaluating model generalization ability to a higher degree.

Logits divergence statistics measures the statistical deviation of a neuron's response compared to the population average logits over the batch task. These statistics can be used to detect anomalous logit distributions for monitoring purposes.

Context pruning deletes random tokens, similar to BERT-style masked training, except that full token deletion is done since GPT-style tokenizers are not trained with [MASK] tokens so that substitute is unavailable. This is an initial inclusion of context noising added to the validator at a very low noise intensity to reward models that have not simply overfit to the known training data. Validator testing at the intended noise level has shown a strong ranking correlation to the noiseless case.

[BIT-612]: https://opentensor.atlassian.net/browse/BIT-612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -1362,6 +1373,105 @@ def shapley_base(uids: torch.Tensor, query_responses: List[List[torch.FloatTenso
...
+        # For each batch task, calculate excess deviation above a single stddev, in terms of stddev,
+        # and apply power to increase score above two stddev, and decrease between one and two stddev.
+        # This will effectively allow zero excess below one stddev, and minimal excess below two stddev,
+        # but amplify any excess above two stddev (only 2.1% of population for normal dist).
+        for _uid, _stats in stats.items():
+            if 'logits_divergences' + ext in _stats:
+                try:
+                    excess = torch.clamp(_stats['logits_divergences' + ext] - (avg + std), 0)  # divergence > avg + std
+                    excess /= std + 1e-9  # stddev multiples above 1 stddev
+                    excess = torch.pow(excess, 3)  # reduce < 2std, increase > 2std
```
Eugene-hu, line 1460: Would we want this to be parameter on subtensor?

Eugene-hu: Not for this PR, but I think it would be something similar to the synergy/shapley power laws

Eugene-hu (approved):
LGTM

bittensor/utils/tokenizer_utils.py
```diff
@@ -1014,6 +1014,54 @@ def phrase_cross_entropy(target_phrases: Union[List[List[int]], torch.Tensor],
...
+    topk_tokens = topk_tensor[:, :-1, 1].round().to(torch.int64)  # [batch_size, topk] first tokens
+    topk_probs = topk_tensor[:, :-1, 0]  # [batch_size, topk] Probabilities for each phrase in topk
+    floor_probs = topk_tensor[:, -1, 0]  # [batch_size] Floor probabilities as mean probability for non-topk tokens
+
+    topk_probs = torch.clamp(topk_probs, 0, 1)  # [batch_size, topk] ensure probabilities within [0, 1]
+    floor_probs = torch.clamp(floor_probs, 0, 1)  # [batch_size] ensure floor probabilities within [0, 1]
+
+    # === Ensure total probability is 1 ===
+    total_probs = topk_probs.sum(dim=-1) + max(0, vocab_size_min - topk) * floor_probs  # [batch_size] total probs
+    n_topk_probs = topk_probs / total_probs[:, None]  # [batch_size, topk] normalized topk_probs
```
isabella618033, line 1056: `topk_probs` exclude `floor_probs` while `total_probs` include `floor_probs`, 
so i think `n_topk_probs` may not sum up to 1 as the comments describe.

bittensor/_neuron/text/core_validator/__init__.py
opentaco: Yes, together with the prune_len, although we'll do this in a future PR.

bittensor/utils/tokenizer_utils.py
opentaco: Correct, `n_topk_probs` would not necessarily sum to 1, but at least the normalization ensures that with `floor_probs` everything sums to 1, which is the main concern.

isabella618033 (approved):
LGTM!

unconst (approved):

PR #1036, opened 2023-01-02
fix btcli list with wallet.path

camfairchild:
This PR fixes a bug with `btcli list` when `wallet.path` ends in a '/'  
The expected behaviour is that the output is the same as when `wallet.path` does not end in `/`  
This is fixed by using `os.path.join` instead of a manual joining.   

See:
Screenshot 2023-01-02 at 11 59 45 AM


joeylegere (approved):
LGTM

PR #1038, opened 2023-01-04
Bit 614 gptneox compatibility check

isabella618033:
Some models (eg. gptneox / pythia / finetuned opt) are not compatible with half precision, and outputing hidden state / logits with nan causing the error message mentioned [here](https://discord.com/channels/799672011265015819/801875196801450045/1059181423388524664)
Which is not an internal bug.

Updated warning message to suggest users to turn off `--neuron.autocast` in case of using half precision  non compatible models
image

### DIFF
--- a/bittensor/_neuron/text/core_server/nucleus_impl.py
+++ b/bittensor/_neuron/text/core_server/nucleus_impl.py
@@ -476,7 +476,7 @@ class server(torch.nn.Module):
         """
         if hasattr(model_output, 'hidden_states') and model_output.hidden_states[-1].isnan().sum() > 0:
             raise ValueError("Got nan value from model last hidden state. If you are using cuda, try setting --neuron.autocast to False.")
-        
+
         if model_output.logits.isnan().sum() > 0:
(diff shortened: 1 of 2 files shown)


bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -459,6 +462,26 @@ def _forward(_model_output=model_output):
...
+
+            Returns:
+                check_status (:type: `bool`):
+                    True if the model_output is valid.
+        """
+        if hasattr(model_output, 'hidden_states') and model_output.hidden_states[-1].isnan().sum() > 0:
+            raise ValueError("Got nan value from model last hidden state. If you are using cuda, try setting --neuron.autocast to False.")
+
+        if model_output.logits.isnan().sum() > 0:
+            raise ValueError("Got nan value from model logit. If you are using cuda, try setting --neuron.autocast to False.")
```
opentaco: > Got nan value from model logit.

change to

> Got nan value from model logits.

opentaco (approved):

PR #1039, opened 2023-01-04
Circle CI fixes

Eugene-hu:
The mock subtensor was giving us all sorts of issues. We should instead mock all of the calls that are needed for the test.


camfairchild (approved):
LGTM 🔥

PR #1042, opened 2023-01-04
Reformat of the Server code 

Eugene-hu:
The server now follows the same format as the core validator and is a single class.

List of changes:
- Removed run.py from the core_server and moved it into the main class
- All functions are now part of the core_server.neuron class
- Removed the default backwards from axon, now instead expects user to the attach their own backward function
- Added additional documentation


bittensor/_neuron/text/core_server/nucleus_impl.py
```diff
@@ -428,6 +426,7 @@ def encode_forward_causallmnext(self, token_batch, std_tokenizer=None, topk: int
         """
         transformers.set_seed(0)
         transformers.enable_full_determinism(0)
+        print('forward called checkpoint')
```
isabella618033: Is this a debug logging or just temporary?

isabella618033 (approved):
LGTM!

PR #1049, opened 2023-01-10
support SubnetInfo RPC

camfairchild:
This PR adds the impl for using the custom subnet info rpc methods instead of a manual pull from the chain.

See: https://github.com/opentensor/paratensor/pull/33


unconst (approved):
LGTM

Eugene-hu (approved):

PR #1050, opened 2023-01-11
Fix stake is none

camfairchild:
Stake on the chain is represented as a map of coldkey to stake value for each Neuron (hotkey).  

This PR modifies the neuron stake to match this representation, as expected.

## TODO:
- [ ] ~~Convert stake map to tensor~~
- [ ] ~~Convert stake map to dataframe for state dict~~
Decided not to save stake map per @unconst


camfairchild:
Closing until tests pass

PR #1054, opened 2023-01-12
[GH-1052] - Fix prompts when default same as flag

### LINKED ISSUE #1052: Prompted to select network on Finney even when --subtensor.network finney argument is passed
https://github.com/opentensor/bittensor/blob/017952f7c28f4370bc6d1c9add0c4d728072218e/bittensor/_cli/commands/run.py#L87

When using 'btcli run' it prompts me to select the network even though I pass in the --subtensor.network finney argument.

Changing the referenced line above to:
if **not** config.subtensor.get('network') == bittensor.defaults.subtensor.network and not config.no_prompt:

Resolved the issue.

camfairchild:
This PR adds a map `Config.__is_set` to the `Config` class.  
This map is from the flattened parameter name (e.g. `config.subtensor.network` -> `config.__is_set["subtensor.network"]`) to a boolean value indicating if the value has been set in a flag (or not).  

This map is created from the defaults set in `bittensor.defaults` on creation of the config, from the parser. Any args that use `default=argparse.SUPPRESS` will be set to `False` in the map unless they are set as a flag. 

Then, the config will be filled with the default values for these flags, avoiding a breaking change to the config class.  

You can then check if any of these args are set using `config.is_set('subtensor.network')`.  

The result, is we don't require prompting the user for values that are set, even if the arg is equal to the default. i.e. `btcli --subtensor.network finney` will not prompt for network again. 

Closes #1052

### DIFF
--- a/bittensor/_cli/commands/delegates.py
+++ b/bittensor/_cli/commands/delegates.py
@@ -101,10 +101,10 @@ class DelegateStakeCommand:
 
     @staticmethod   
     def check_config( config: 'bittensor.Config' ):
-        if config.subtensor.get('network') == bittensor.defaults.subtensor.network and not config.no_prompt:
+        if not config.is_set('subtensor.network') and not config.no_prompt:
             config.subtensor.network = Prompt.ask("Enter subtensor network", choices=bittensor.__networks__, default = bittensor.defaults.subtensor.network)
 
-        if config.wallet.get('name') == bittensor.defaults.wallet.name and not config.no_prompt:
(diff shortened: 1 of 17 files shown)


S0AndS0 (approved):
Other than `bittensor/_config/__init__.py` everything looks easy to understand!
> I.E. I'd have to spend some quality time with the code to form an educated opinion

I feel like `bittensor/_subtensor/__init__.py` argument additions/string-concatenation probably be a good candidate for future format string(s) refactoring

camfairchild:
Closed to re-open as #1242

PR #1057, opened 2023-01-17
Hotfix/3.6.2/validator logit parameters

Eugene-hu:
Hotfix Bittensor v3.6.2

***Motivation***
In v3.6.1, we introduce anomaly detector, which measures the divergence of server responses to the overall network response. However we saw that anomaly detection was negatively impacting the server scores that deviate too much from the consensus. As such, we are adding two parameters that control the power of the anomaly detection.

***Major Changes***
- added subtensor parameters logit_divergence and prune_len 

***Minor Changes***
- Small fixes for the integration tests and config


bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -400,13 +400,15 @@ def run_epoch( self ):
         batch_size = self.subtensor.validator_batch_size 
         sequence_length = self.subtensor.validator_sequence_length
         validation_len = self.config.neuron.validation_len  # Number of tokens to holdout for phrase validation beyond sequence context
-        prune_len = self.config.neuron.prune_len  # Number of tokens to holdout for phrase validation beyond sequence context
+        # Number of tokens to prune for phrase validation beyond sequence context
+        prune_len = self.subtensor.prune_len if self.config.neuron.prune_len == -1 else self.config.neuron.prune_len
```
isabella618033: Should we allow user's prune_len override the subtensor prune_len?

isabella618033 (approved):

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -424,6 +424,32 @@ def make_substrate_call_with_retry():
...
+        Returns:
+            logits_divergence (int):
+                the divergence value for logit distances, a measure for anomaly detection 
+        """
+        @retry(delay=2, tries=3, backoff=2, max_delay=4)
+        def make_substrate_call_with_retry():
+            with self.substrate as substrate:
+                U64MAX = 18446744073709551615
+                return substrate.query( module='SubtensorModule', storage_function = 'ValidatorLogitsDivergence' ).value/U64MAX
+        return make_substrate_call_with_retry()
```
eduardogr, line 452: I miss unit tests to cover this functions.

From my point of view, If we are hotfixing, we should define what we are covering creating the missing tests that was not existing and lead us to have the scenario we are fixing.


What do you think about this?

opentaco (approved):

bittensor/_subtensor/subtensor_impl.py
Eugene-hu: Yes, I agree. In this case, the master branch was failing tests, so I pulled a few updates that fixed the tests. Those in of itself should have been a hotfix. I will update the description

PR #1065, opened 2023-01-21
[hotfix][3.6.3] Merging back hotfix into finney

eduardogr:


bittensor/_cli/cli_impl.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, config: 'bittensor.Config' ):
                 config (:obj:`bittensor.Config`, `required`): 
                     bittensor.cli.config()
         """
-        if not config.no_version_checking:
+        if config.get('no_version_checking') != None and not config.no_version_checking:
```
camfairchild: Could be `if not config.get('no_version_checking'):`   

because None is falsey and indicates False here also

camfairchild (commented):
LGTM. Just a nit

bittensor/_cli/cli_impl.py
eduardogr: please, add the code suggestion and we add it right away :)

camfairchild (approved):
lgtm

Eugene-hu (approved):

PR #1067, opened 2023-01-23
Remove Py 3.7

camfairchild:
This PR ups the python requirements to py3.8+, therefore removing support for python3.7.  

This comes up as per: https://devguide.python.org/versions/  python 3.7 is being deprecated shortly this year.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,7 @@ workflows:
       - build-and-test:
           matrix:
             parameters:
-              python-version: ["3.7.14", "3.8.14", "3.9.13", "3.10.6"]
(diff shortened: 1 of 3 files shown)


eduardogr:
Have you checked if all packages in our requirements are aligned with python3.7. deprecation ? Maybe we have to update some packages.

Don't rely on tests here because we do not have fully coverage there.

TL;DR: Check if every python package we work with is compatible with python >=3.8

eduardogr (approved):
If everything is OK with our dependencies I'm ok with this

thanks for the update!

Eugene-hu (approved):

PR #1072, opened 2023-01-26
[fix] instantiate subtensor before check for netuids

camfairchild:
This PR fixes an error with the core server.   
The subtensor was instantiated to the wrong variable before a call was tried to it.

### DIFF
--- a/bittensor/_neuron/text/core_server/__init__.py
+++ b/bittensor/_neuron/text/core_server/__init__.py
@@ -79,6 +79,17 @@ class neuron:
         if config == None: config = server.config()
         config = config; 
 
+        config.netuid = netuid if netuid != None else config.netuid
+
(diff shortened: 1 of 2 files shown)


unconst (approved):

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -79,6 +79,17 @@ def __init__(
         if config == None: config = server.config()
         config = config; 
 
+        config.netuid = netuid if netuid != None else config.netuid
+
+        subtensor = bittensor.subtensor ( config = config ) if subtensor == None else subtensor
```
Eugene-hu, line 84: This is changed for #1053, lets merge your PR first @camfairchild, I will add this to that pr

camfairchild: @Eugene-hu merged

PR #1075, opened 2023-01-26
[fix] [ci] modify cache key

camfairchild:
This PR adds `requirements/dev.txt` to the venv cache key
This should make the venv re-run when there are changes in **either** `prod.txt` or `dev.txt`


eduardogr (approved):
LGTM

PR #1078, opened 2023-01-28
Fix Val logging

camfairchild:
This PR fixes validator log out


camfairchild:
closing, fixed in #1053

PR #1079, opened 2023-01-28
change back to ip str

camfairchild:
This converts the IP back to a string for logging the `Already Served` axon output.


camfairchild:
Close as implemented by #1053

PR #1087, opened 2023-02-09
update substrate interface version

camfairchild:


eduardogr (approved):
super 🚀

eduardogr:
Do we have tests to cover our scenarios that involved these libraries?

camfairchild:
> Do we have tests to cover our scenarios that involved these libraries?

Writing them now for this branch.   

This version needs to be updated for chain compatibility, though, so we must update.

PR #1088, opened 2023-02-09
fix bittensor dir path

camfairchild:


eduardogr (approved):
OMG, thanks for this catch

LGTM!

PR #1091, opened 2023-02-09
Server blacklist and Bittensor version fix

Eugene-hu:
- Correctly returns unauthenticated errors due to blacklisting
- updates bittensor versioning such that higher versions such (4.11.0) are permitted by design.
- Remove legacy signature


bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -634,7 +637,7 @@ def time_check():
             registration_check()
             time_check()
             stake_check()            
-            return False
+            return False, ''
```
camfairchild:
```suggestion
            return False, None
```

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -634,7 +637,7 @@ def time_check():
...
         except Exception as e:
             self.prometheus_counters.labels("blacklisted").inc()
-            return True
+            return True, e
```
camfairchild: Also add type hint to function 
```python
-> Tuple[bool, Optional[Exception]]
```

camfairchild (changes requested):
See: https://github.com/opentensor/bittensor/pull/1091/files#r1102231464

bittensor/_neuron/text/core_server/__init__.py
Eugene-hu: thanks! Nice catch, let me fix that

Eugene-hu:
closes the pr in favour of #1240

PR #1093, opened 2023-02-13
Uint serializers

unconst:


Eugene-hu:
Closing in favor of #1100 which merges these update into finney branch

PR #1103, opened 2023-02-28
Security update 3.7.0

Eugene-hu:
- Additional protection against misformed requests
- Allows the ability to blacklist against specific hotkeys (`--neuron.blacklist.hotkeys`)


bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -92,7 +92,8 @@ def __init__(
         causallmnext = None,
         seq2seq = None,
         synapse_list = None,
-        netuid = None
+        netuid = None,
+        blacklist_hotkeys = [],
```
joeylegere: Should use `None` here rather than a mutable arg: https://stackoverflow.com/questions/1132941

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -427,18 +432,25 @@ def synapse_check(self, synapse, hotkey):
                 return False
             
         elif synapse.synapse_type == bittensor.proto.Synapse.SynapseType.TEXT_CAUSAL_LM:
-
-            if self.metagraph.S[incoming_uid] < self.config.neuron.causallm_stake:
+            batch_size, sequence_len  =  inputs_x[0].size()
+            if (self.metagraph.S[incoming_uid] < self.config.neuron.causallm_stake) \
+                and (batch_size > self.config.neuron.max_batch_size) \
+                and (sequence_len > self.config.neuron.max_sequence_len):
```
joeylegere: Should these be `or`? Otherwise I could still ddos someone with a large batch size and smaller sequence length, or vice versa.

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -406,7 +411,7 @@ def run(
                     except Exception as e:
                         logger.error('Failure setting weights on chain with error: {}', e)
 
-    def synapse_check(self, synapse, hotkey):
+    def synapse_check(self, synapse, hotkey, inputs_x=None):
```
joeylegere, line 414: We should expand the checks on `TEXT_LAST_HIDDEN_STATE` as well; it currently only checks stake, but it requires a significant amount of bandwidth to send the response. (e.g. `float32 * 32 * 256 * 1024` is 32MB to upload)

bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -427,18 +432,25 @@ def synapse_check(self, synapse, hotkey):
...
-
-            if (self.metagraph.S[incoming_uid] < self.config.neuron.seq2seq_stake) and (self.metagraph.S[incoming_uid,  self.uid]):
+            batch_size, sequence_len  =  inputs_x[0].size()
+            if (self.metagraph.S[incoming_uid] < self.config.neuron.seq2seq_stake) \
+                and (batch_size > self.config.neuron.max_batch_size) \
+                and (sequence_len > self.config.neuron.max_sequence_len) \
+                and (self.metagraph.W[incoming_uid,  self.uid]):
                 return False     
         else:
             return False
```
joeylegere: If the synapse isn't in the known synapse types, maybe we should raise an error or warning instead? It could indicate someone's server is out of date, for instance.

bittensor/_neuron/text/core_server/__init__.py
Eugene-hu: Nice catch! Fixed

bittensor/_neuron/text/core_server/__init__.py
Eugene-hu: Fixed as well

bittensor/_neuron/text/core_server/__init__.py
Eugene-hu: Good call; just added this

bittensor/_neuron/text/core_server/__init__.py
Eugene-hu: Fixed as well

joeylegere (approved):

PR #1105, opened 2023-03-01
Release 3.7 subtensor integration

isabella618033:
- Based from finney, as well supporting backward compatibility with nakamoto.
- Bug fixes on both nakamoto cli and finney cli.
- Classes that was left keeping both nakamoto and finney implementation: cli, metagraph, subtensor'
    - Checked by subtensor.network or config.subtensor.network

### DIFF
--- a/bittensor/_neuron/text/core_server/__init__.py
+++ b/bittensor/_neuron/text/core_server/__init__.py
@@ -101,15 +101,6 @@ class neuron:
 
         config.netuid = netuid if netuid != None else config.netuid
 
-        subtensor = bittensor.subtensor ( config = config ) if subtensor == None else subtensor
-        if config.netuid == None:
-            config.netuid = subtensor.get_subnets()[0]
-        
-        # Verify subnet exists
-        if not subtensor.subnet_exists( netuid = config.netuid ):
-            bittensor.__console__.print(f"[red]Subnet {config.netuid} does not exist[/red]")
-            sys.exit(1)
-
         if synapse_list != None:
             config.neuron.lasthidden = False
             config.neuron.causallm = False
@@ -155,7 +146,7 @@ class neuron:
         self.prometheus_info = Info('neuron_info', "Info sumamries for the running server-miner.", registry=registry)
         self.config.to_prometheus()
 
-        if self.config.netuid == None:
+        if self.config.netuid == None and self.config.subtensor.network == 'finney':
             subtensor = bittensor.subtensor(config = config) if subtensor == None else subtensor
             self.config.netuid = subtensor.get_subnets()[0]
 
@@ -182,7 +173,7 @@ class neuron:
         self.query_data = {}
 
         # Verify subnet exists
-        if not self.subtensor.subnet_exists( netuid = self.config.netuid ):
+        if self.config.subtensor.network == 'finney' and not self.subtensor.subnet_exists( netuid = self.config.netuid ):
             bittensor.__console__.print(f"[red]Subnet {self.config.netuid} does not exist[/red]")
             sys.exit(1)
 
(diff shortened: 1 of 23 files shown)


bittensor/_neuron/text/core_server/__init__.py
```diff
@@ -638,3 +629,27 @@ def time_check():
...
             return True
+
+    def get_neuron(self):
+        if self.subtensor.network == 'finney':
+            nn = self.subtensor.get_neuron_for_pubkey_and_subnet(self.wallet.hotkey.ss58_address, netuid = self.config.netuid)
+        elif self.subtensor.network == 'nakamoto':
+            nn = self.subtensor.neuron_for_pubkey(self.wallet.hotkey.ss58_address)
+        return nn
+
+    def get_n(self):
```
eduardogr: @isabella618033 what do you think about having more semantic in this function name??

From my point of view `get_n` should expose something more specific

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -553,22 +577,22 @@ def run_epoch( self ):
             weights=sample_weights.detach().to('cpu'),
             netuid = self.config.netuid,
             wallet=self.wallet,
-            version_key=bittensor.__version_as_int__, # TODO: correct?
+            version_key=1,
```
eduardogr, line 574: what is this about?

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -126,6 +130,7 @@ def __init__(
             self.config.axon._mock = True
         print ( self.config )
 
+        copy.deepcopy(self.config)
```
eduardogr: why are executing deepcopy three times? Do we need that? copy.deepcopy returns the result as a value right? So, why are not receiving the result?

eduardogr (commented):
Nice Done @isabella618033 !

I'd approve but first i'd like to resolve some comment I left. After that everything LGTM. Nicely done!

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: It was for some debugging purpose, and it is redundant now.

bittensor/_neuron/text/core_validator/__init__.py
Eugene-hu: The version key should be the bittensor version. Basically, the chain will check if the version key matches the latest bittensor version. However, that is not working right now, so it is set to 1

bittensor/_neuron/text/core_server/__init__.py
isabella618033: Renamed get_n to get_neuron_num.

bittensor/_neuron/text/core_validator/__init__.py
isabella618033: Then are we keeping it as 1?

Eugene-hu: For now, yes. I do not believe that feature is turned on yet.

isabella618033: Okay! Just reverted it to 1:)

bittensor/_neuron/text/core_validator/__init__.py
eduardogr: can we remove it now??

Eugene-hu (approved):

bittensor/_neuron/text/core_validator/__init__.py
```diff
@@ -48,6 +48,7 @@
 from loguru import logger
 from threading import Lock
 from prometheus_client import Counter, Gauge, Histogram, Summary, Info
+import copy
```
eduardogr:
```suggestion
```

eduardogr: since we are not using it anymore

eduardogr (approved):
LGTM Thanks @isabella618033  !

PR #1106, opened 2023-03-02
Dataset API cleanup

joeylegere:
* Change the name to reflect that `dataset_names` is plural.
* Add checks and warn if a requested dataset doesn't exist.
* Allow strings for selecting a single dataset from the mountain.


Eugene-hu (approved):
LGTM

PR #1107, opened 2023-03-02
Add hyperparams to metagraph

camfairchild:
This PR adds an `info: SubnetInfo` attribute to `Metagraph` and allows it to be saved and loaded from/to the metagraph file.


camfairchild:
closing in-favour of #1121

PR #1123, opened 2023-03-09
btcli burned_register

camfairchild:
This PR adds the `btcli burned_register` command and a test

Also:
- fixes default netuid to 1 
- fixes inspect when not registered
- switches to subtensor.get_balance for inspect


bittensor/_subtensor/extrinsics/registration.py
```diff
@@ -178,4 +178,97 @@ def register_extrinsic (
...
+            # create extrinsic call
+            call = substrate.compose_call( 
+                call_module='SubtensorModule',  
+                call_function='burned_register', 
+                call_params={ 
+                    'netuid': netuid,
+                    'hotkey': wallet.hotkey.ss58_address
+                } 
+            )
+            extrinsic = substrate.create_signed_extrinsic( call = call, keypair = wallet.hotkey )
```
isabella618033: should we use coldkey to sign instead? cause burn reg burns tao (stake)

camfairchild: Oops. fixed

bittensor/_subtensor/subtensor_impl.py
```diff
@@ -364,6 +382,11 @@ def kappa (self, netuid: int, block: Optional[int] = None ) -> Optional[int]:
     def difficulty (self, netuid: int, block: Optional[int] = None ) -> Optional[int]:
         if not self.subnet_exists( netuid ): return None
         return self.query_subtensor( "Difficulty", block, [netuid] ).value
+    
+    """ Returns network Burn hyper parameter """
+    def burn (self, netuid: int, block: Optional[int] = None ) -> Optional[bittensor.Balance]:
```
Eugene-hu, line 387: Can we change this naming with something more specific? We might have more parameters later that also require burning

Eugene-hu (approved):
LGTM, left you a comment

PR #1124, opened 2023-03-10
Add prometheus extrinsic

isabella618033:


bittensor/_subtensor/extrinsics/prometheus.py
```diff
@@ -0,0 +1,100 @@
...
+
+def prometheus_extrinsic(
+    subtensor: 'bittensor.Subtensor',
+    wallet: 'bittensor.wallet',
+    port: int, 
+    netuid: int,
+    ip: int = None,
+    wait_for_inclusion: bool = False,
+    wait_for_finalization = True,
+    prompt = False,
```
Eugene-hu: I think we can remove prompt from the args. I don't see it being used anywhere

Eugene-hu (approved):
LGTM; just one comment

PR #1126, opened 2023-03-10
3.7 server fix

Eugene-hu:
- Fixes for the core-server on finney.
- Changes neuron.stake back to a balance
- Added neuron.stake_dict which contains a dict of balances


bittensor/_subtensor/chain_data.py
```diff
@@ -40,7 +40,8 @@ class NeuronInfo:
     netuid: int
     active: int    
     # mapping of coldkey to amount staked to this Neuron
```
camfairchild: Move comment above `stake_dict`

camfairchild (approved):
Just move the comment and remove total_stake

PR #1131, opened 2023-03-13
Use new rpc format

camfairchild:
This PR uses the new format for custom RPC responses as changed here: https://github.com/opentensor/subtensorv3/pull/31

PR #1132, opened 2023-03-15
Revert "Use new rpc format"

camfairchild:
Reverts opentensor/bittensor#1131


camfairchild:
Reverting until chain is upgraded to support changes

PR #1134, opened 2023-03-16
Use scalecodec for custom RPCs

camfairchild:
This PR changes the custom RPC encoding to use scalecodec
See: https://github.com/opentensor/subtensorv3/pull/42


bittensor/_subtensor/chain_data.py
```diff
@@ -16,28 +16,134 @@
...
+                ["ip_type", "u8"],
+            ],
+        },
+    }   
+}
+
+class ChainDataType(Enum):
+    NeuronInfo = 1
+    SubnetInfo = 2
+    DelegateInfo = 3
```
Eugene-hu, line 121: Does this class need to be extended to the axon and prometheus info as well?

Eugene-hu (approved):
LGTM, left you a comment

bittensor/_subtensor/chain_data.py
camfairchild: those don't get sent by themselves on rpc so they don't need the enum for decoding

shibshib (approved):

PR #1139, opened 2023-03-17
Change ValidatorLogitsDivergence to u16

opentaco:
Change ValidatorLogitsDivergence to u16 from u64


camfairchild (approved):
Looks good to me :fire:

PR #1148, opened 2023-03-18
version key fixes

Eugene-hu:
- updates version key to be the bittensor version
- Remove version key from set weight calls in the server and validator
- Please update version key on subtensor before the merging this pr


camfairchild (approved):
LGTM. Do we know the size of the int?

unconst (approved):

PR #1164, opened 2023-03-19
recycle register 2

quac88:
This PR changes burned_register to recycle_register. Please not that the function and class calls in bittensor have been changed but Subtensor remains the same.


Eugene-hu (approved):
LGTM, nice work

PR #1184, opened 2023-03-21
changed explorer_url for transaction on finney chain

softIntelligent:
Updated the explorer_url after made a transaction.


shibshib (approved):

camfairchild (changes requested):
This change is only for nakamoto. So the URL should not change here.
See #1200

camfairchild:
Closed in favour of #1200

PR #1186, opened 2023-03-21
Delegate index system

theeblardo:
coldkey 5CezfqBQx7gy9YL69mepHDPHUC4unV1HYcCtYa5KdXEerBUN
hotkey 5D9wuLzoWSbucNx6NQRo7wdNfvW2poMy3NumSiggsEStPc9X

TAO ALMANACH

TAO ALMANACH is the validator of bittensor mod blardo and polkadot mod mister_cole, and we'd love your delegation!

okay, can't figure out the signature ... got this instead:
honestly i think this is the best way to prove it is blardo and mr_cole

(well, i tried ... blardo)

### DIFF
--- a/bittensor/_cli/commands/delegates.py
+++ b/bittensor/_cli/commands/delegates.py
@@ -24,9 +24,13 @@ from rich.prompt import Prompt
 from rich.prompt import Confirm
 from rich.console import Text
 from dataclasses import dataclass
+from scalecodec import ss58_decode
 import requests
 
 console = bittensor.__console__
+
+delegate_index_length: int = 9 # Must be ODD. Number of characters in the delegate index.
+
 @dataclass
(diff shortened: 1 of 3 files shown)


mrseeker:
@theeblardo Have your tried running it on the same server as the validator, using "python3" instead of "python"?

theeblardo:
yep.  i did using python 3 and the screenshot is the result.

blardo

(ps ... such a smoooooth brain.  this should be low priority for all
y'all.  i'd like to get on the list, but have to get back to my real life
for a while now)

camfairchild:
@theeblardo Please open a new PR with a signature or a screenshot of an on-chain remark

theeblardo:
what qualifies as an on-chain remark? would btcli overview be sufficient?

blardo

PR #1195, opened 2023-03-21
3.7 version checking

Eugene-hu:
Blacklist missed previous version checking
- Adds back version checking for older signatures
- Anything older than 3.7.0 will not be able to query the servers


isabella618033 (approved):

camfairchild (approved):
🔥 🚒

bittensor/_axon/__init__.py
```diff
@@ -412,8 +412,12 @@ def parse_signature(
     ) -> Tuple[int, str, str, str, int]:
         r"""Attempts to parse a signature from the metadata"""
         signature = metadata.get("bittensor-signature")
+        version = metadata.get('bittensor-version')
         if signature is None:
             raise Exception("Request signature missing")
+        if version < 370:
+            raise Exception("Incorrect Version")
+        
         for parser in [self.parse_signature_v2, self.parse_legacy_signature]:
```
adriansmares, line 421: You can just drop the `self.parse_legacy_signature` from below. Clients older than 3.6.0 will send such signatures and will be rejected.

bittensor/_axon/__init__.py
```diff
@@ -412,8 +412,12 @@ def parse_signature(
     ) -> Tuple[int, str, str, str, int]:
         r"""Attempts to parse a signature from the metadata"""
         signature = metadata.get("bittensor-signature")
+        version = metadata.get('bittensor-version')
```
adriansmares, line 415: Metadata is a `Dict[str, str]`. You need to cast to integer if you want to compare it with an integer.

Eugene-hu: Thanks Adrian!

PR #1196, opened 2023-03-21
Update delegates.json

LucDeville:
Verified True
	Validator information: {"5Gpt8XWFTXmKrRF1qaxcBQLvnPLpKi6Pt2XC4vVQR7gqNKtU": {"name": "StakeBoost Validator", "url": "https://taostats.io/validators/stakeboost-validator/", "description": "We refund 50% of the 18% take that validators get from delegated stake emissions. Payback is automatic after a minimum holding period of 30 days to the users coldkey."}}
 	Validator signature: 9c437cff7a9008b39e741431e15454e8b877ff1f8c0786ec0e33f771920ea1623d28bf19005e7150c805496db6a53c0c62014802d0847a9ff533a77e0d94a48a

### DIFF
--- a/delegates.json
+++ b/delegates.json
@@ -63,5 +63,10 @@
         "name": "Tensor.Exchange", 
         "url": "www.tensor.exchange", 
         "description": "Bittensor's first community OTC exchange"
+    },
+    "5Gpt8XWFTXmKrRF1qaxcBQLvnPLpKi6Pt2XC4vVQR7gqNKtU": {
+        "name": "StakeBoost Validator", 
+        "url": "https://taostats.io/validators/stakeboost-validator/", 
+        "description": "We refund 50% of the 18% take that validators get from delegated stake emissions. Payback is automatic after a minimum holding period of 30 days to the users coldkey."
     }
 }


delegates.json
```diff
@@ -63,5 +63,10 @@
         "name": "Tensor.Exchange", 
         "url": "www.tensor.exchange", 
         "description": "Bittensor's first community OTC exchange"
+    },
+    "5Gpt8XWFTXmKrRF1qaxcBQLvnPLpKi6Pt2XC4vVQR7gqNKtU": {
+        "name": "StakeBoost Validator", 
+        "url": "https://taostats.io/validators/stakeboost-validator/", 
+        "description": "We refund 50% of the 18% take that validators get from delegated stake emissions. Payback is automatic after a minimum holding period of 30 days to the users coldkey."
```
camfairchild, line 70: Hi, please don't include any "refunds", etc in the description

camfairchild (changes requested):

delegates.json
LucDeville: why not? this is how we want to make it attractive for now.  we have a mission beyond that in graphics AI. but for now we need to have a USP, because we are on the edge with our stake and if we get out of top128 we cant contribute anything

camfairchild: > why not  

We have a strict rule not to include these kinds of statements in our codebase. Apologies

camfairchild:
Closes in-favour of #1197

PR #1212, opened 2023-03-22
Update delegates.json

cisterciansis:


camfairchild:
@Dalber0630 please use the PR script to generate a signature for this change, using your hotkey

camfairchild:
I will re-open when you add the signature

PR #1240, opened 2023-03-28
Blacklist fixes + depreciation of old signatures

Eugene-hu:
- removes the old signatures from the code
- adds in fixes for the blacklist error 
- adds additional checks for synapse when allowing no_registration


bittensor/_receptor/receptor_impl.py
```diff
@@ -141,9 +132,7 @@ def sign_v2(self):
         return ".".join([nonce, sender_hotkey, signature, self.receptor_uid])
 
     def sign(self):
```
isabella618033: maybe combine sing and sign_v2 ?

isabella618033 (approved):

PR #1241, opened 2023-03-29
[BIT-636] Change u16 weight normalization to max-upscaling

opentaco:
### [BIT-636] Change u16 weight normalization to max-upscaling

Subtensor has been upgraded with [[BIT-104] Max-upscale weights](https://github.com/opentensor/subtensor/pull/99), which now allows for validator weights to be stored with greater precision than before, yet still using `u16` 16-bit unsigned ints.

Bittensor `core_validator` can now use full `u16` bitwidth so that `max_weight=U16_MAX`, then rely on subtensor epoch to properly normalize weights in `I32F32`. This means that weights submission extrinsic to subtensor does not have to be pre-normalized, and weights are transmitted and stored with higher precision.

Weight-setting should now be more precise and server rewards more accurate than before.

[BIT-636]: https://opentensor.atlassian.net/browse/BIT-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


isabella618033 (approved):
LGTM!

Eugene-hu (approved):
LGTM!

PR #1243, opened 2023-03-29
Add raw spec for local test and new bins

camfairchild:
Updates the mock subtensor bins and adds a spec file

Also fixes a few tests and minor bugs


bittensor/_cli/commands/overview.py
```diff
@@ -338,5 +338,8 @@ def check_config( config: 'bittensor.Config' ):
             config.wallet.name = str(wallet_name)
 
         if config.netuid != []:
-            config.netuid = [int(netuid) for netuid in config.netuid]
+            if not isinstance(config.netuid, list):
```
Eugene-hu, line 341: Is this change only for the overview or for all of the cli commands?

bittensor/_subtensor/__init__.py
```diff
@@ -37,7 +37,7 @@ class subtensor:
     """Factory Class for both bittensor.Subtensor and Mock_Subtensor Classes
 
     The Subtensor class handles interactions with the substrate subtensor chain.
-    By default, the Subtensor class connects to the Nakamoto which serves as the main bittensor network.
+    By default, the Subtensor class connects to the Finney which serves as the main bittensor network.
```
Eugene-hu, line 40: We should update a lot of the nakamoto mentions in the docstrings

bittensor/_subtensor/extrinsics/staking.py
```diff
@@ -68,7 +68,7 @@ def add_stake_extrinsic(
         hotkey_ss58 = wallet.hotkey.ss58_address 
 
     # Flag to indicate if we are using the wallet's own hotkey.
-    own_hotkey: bool = (wallet.hotkey.ss58_address == hotkey_ss58)
+    own_hotkey: bool = False
```
Eugene-hu: Why is this always false?

Eugene-hu (approved):
LGTM, let you a few questions

bittensor/_cli/commands/overview.py
camfairchild: This is only present for overview to allow restricting to specific netuids, instead of querying the metagraph for all. 
The issue was the API used an int for `config.netuid` and I had changed the arg to a `List[int]`, so this check fixes the broken expectation.

bittensor/_subtensor/extrinsics/staking.py
camfairchild: Lower down. This first is set to default False before the check happens. I guess it could also be
`own_hotkey: bool` with no instantiation 
https://github.com/opentensor/bittensor/pull/1243/files/87284b3dc683e59677eb30e0165ab7a5ebf58300#diff-ed3e1bbf0b6155e43a0fd1870524f320c8c8b0ed9eed4d2d35e7aaa47eee2cf8L75-L78
```python
        # Get hotkey owner
        hotkey_owner = subtensor.get_hotkey_owner( hotkey_ss58 )
        own_hotkey = (wallet.coldkeypub.ss58_address == hotkey_owner)
        if not own_hotkey:
```

Eugene-hu (approved):

PR #1245, opened 2023-03-29
Grab delegates details from GitHub

camfairchild:
This PR grabs the extra delegate details from the new github repo, instead of from `delegates.json`  

See: https://github.com/opentensor/bittensor-delegates


joeylegere (approved):
Solid, LGTM

eduardogr (approved):
very nice!

missing some testing but it looks nice

PR #1247, opened 2023-03-31
Nonce security update

isabella618033:
This PR would address the following from the weekly security status report (28 Mar). 

Point 3
- removed uncessary keyfile data print

Point 5 
- nonce should be bounded by a +/- block_time
- nonce key (ss58_address) should be validated by suvvessful decoding before being written to memory
- added pruning when the nonces dictionary grow to big so that the memory for it would not grow indefinitely

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -22,10 +22,11 @@ import argparse
 import os
 import copy
 import inspect
-import time
+import time as clock
 from concurrent import futures
 from typing import Dict, List, Callable, Optional, Tuple, Union
 from bittensor._threadpool import prioritythreadpool
+from substrateinterface.utils.ss58 import ss58_decode
 
 import torch
 import grpc
@@ -373,6 +374,7 @@ class AuthInterceptor(grpc.ServerInterceptor):
         self.nonces = {}
         self.blacklist = blacklist
         self.receiver_hotkey = receiver_hotkey
+        self.nonce_len = 4096
 
     def parse_legacy_signature(
         self, signature: str
@@ -424,6 +426,15 @@ class AuthInterceptor(grpc.ServerInterceptor):
                 return parts
         raise Exception("Unknown signature format")
 
+    def prune_nonce(self):
+        r"""Remove stored nonce when the size of nonce is too big.""" 
+        if len(self.nonces) > self.nonce_len:
+            nonces = self.nonces.copy()
+            lower_bound = clock.monotonic_ns() - bittensor.__blocktime__*1e9
+            for key in nonces:
+                if self.nonces[key] < lower_bound:
+                    del self.nonces[key]
+
     def check_signature(
         self,
         nonce: int,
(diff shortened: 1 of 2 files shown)


bittensor/_axon/__init__.py
```diff
@@ -441,19 +453,21 @@ def check_signature(
...
-        # Build the key which uniquely identifies the endpoint that has signed
-        # the message.
-        endpoint_key = f"{sender_hotkey}:{receptor_uuid}"
 
         if endpoint_key in self.nonces.keys():
             previous_nonce = self.nonces[endpoint_key]
             # Nonces must be strictly monotonic over time.
-            if nonce <= previous_nonce:
-                raise Exception("Nonce is too small")
+            lower_bound = clock.monotonic_ns() - bittensor.__blocktime__*1e9 # nonce should not be too much earlier then block_time
```
adriansmares, line 460: The reference point of the monotonic clock functions (both `monotonic` and `monotonic_ns`) is undefined - the monotonic clocks are not global, such that nonce bounds cannot be established in this way. Generally, monotonic clocks are implemented based on system uptime timestamps, which obviously are not global.

You can easily test this by simply checking the return value of `monotonic_ns` on two different machines - you will see that the values are wildly different, and as such the bounds computed on the server cannot be used to bound the values generated by the clients.

Time boxing nonces is not useful nor necessary. If the staleness of the request is really a concern, another timestamp should be added to the signature composition, but it won't significantly improve the security of the signature.

bittensor/_axon/__init__.py
```diff
@@ -441,19 +453,21 @@ def check_signature(
...
+            upper_bound = clock.monotonic_ns() + bittensor.__blocktime__*1e9 # nonce should not be too much older then block_time
+            if nonce < lower_bound or nonce > upper_bound or nonce <= previous_nonce:
+                raise Exception(f"Nonce {nonce} is out of range {lower_bound} to {upper_bound} OR nonce is earlier then previous nonce {previous_nonce}.")
 
         if not keypair.verify(message, signature):
             raise Exception("Signature mismatch")
-        self.nonces[endpoint_key] = nonce
+
+        ss58_decode(sender_hotkey) # verify it is a valid hotkey to be written, otherwise raise value error
```
adriansmares, line 468: Why is this needed ? The initializer for `Keypair`, on line 448, already does this:
https://github.com/polkascan/py-substrate-interface/blob/1f558e6a5fa9a1ea71ae25211989ebf858a14075/substrateinterface/keypair.py#L108-L109

bittensor/_axon/__init__.py
```diff
@@ -484,13 +498,19 @@ def intercept_service(self, continuation, handler_call_details):
...
             # signature checking
-            self.check_signature(
-                nonce, sender_hotkey, signature, receptor_uuid, signature_format
-            )
+            self.check_signature(nonce, sender_hotkey, signature, receptor_uuid, signature_format, endpoint_key)
 
             # blacklist checking
             self.black_list_checking(sender_hotkey, method)
+            
+            # update nonce when all checking pass
```
adriansmares, line 510: If you write the nonce only after the blacklist check, the nonce may be reused if the blacklist check fails.

bittensor/_axon/__init__.py
isabella618033: I see, thank you for noting that, I think we have to figure out another solution then

PR #1252, opened 2023-04-06
Bit 151 http securities

isabella618033:
#10 Make sure all of http requests have a timeout.
#9 Remove insecure method of getting external ip through curl ifconfig.me  /myip.dnsomatic.com


bittensor/utils/networking.py
```diff
@@ -82,8 +81,8 @@ def ip__str__(ip_type:int, ip_str:str, port:int):
 class ExternalIPNotFound(Exception):
     """ Raised if we cannot attain your external ip from CURL/URLLIB/IPIFY/AWS """
 
-def get_external_ip() -> str:
-    r""" Checks CURL/URLLIB/IPIFY/AWS for your external ip.
+def get_external_ip(timeout:int = 12) -> str:
```
camfairchild:
```suggestion
def get_external_ip(timeout:int = bittensor.__blocktime__) -> str:
```
Use blocktime here too?

bittensor/utils/networking.py
```diff
@@ -92,59 +91,39 @@ def get_external_ip() -> str:
...
+        get_external_ip_from_ipv6, 
+        get_external_ip_from_wiki
+    ]:
+        try:
+            external_ip = get_ip()
+            if external_ip != None: 
+                assert isinstance(ip_to_int(external_ip), int)
+                return str(external_ip)
+        except Exception:
+            logger.info(f'Failed to get ecternal IP from {get_ip}')
```
camfairchild:
```suggestion
            logger.info(f'Failed to get external IP from {get_ip.__name__}')
```
Typo, and maybe use function name here

bittensor/utils/networking.py
```diff
@@ -92,59 +91,39 @@ def get_external_ip() -> str:
...
+    for get_ip in [
+        get_external_ip_from_aws, 
+        get_external_ip_from_ipinfo, 
+        get_external_ip_from_ipv6, 
+        get_external_ip_from_wiki
+    ]:
+        try:
+            external_ip = get_ip()
+            if external_ip != None: 
+                assert isinstance(ip_to_int(external_ip), int)
```
camfairchild: In what case is the return value not an `int`?

camfairchild (approved):
Looks good, just some typos

bittensor/utils/networking.py
isabella618033: it would be int, but i think it is just for making sure the IP could be converted into int successfully.

bittensor/utils/networking.py
```diff
@@ -23,8 +23,8 @@
 import miniupnpc
 import netaddr
 import requests
-
 from loguru import logger
+import bittensor
```
eduardogr: What do you think about import just the value instead of full module here?? (using relative path)

From my point of view, inside of a module we should use relatives paths since it's the module itself where we are

tests/unit_tests/bittensor_tests/utils/test_network_utils.py
```diff
@@ -57,7 +58,17 @@ def test_int_to_ip6_underflow():
...
+        assert utils.networking.get_external_ip()
+
+    with patch('bittensor.utils.networking.ip_to_int', side_effect = [Exception, 0, Exception, Exception]):
+        assert utils.networking.get_external_ip()
+
+    with patch('bittensor.utils.networking.ip_to_int', side_effect = [Exception, Exception, 0, Exception]):
+        assert utils.networking.get_external_ip()
+
+    with patch('bittensor.utils.networking.ip_to_int', side_effect = [Exception, Exception, Exception, 0]):
+        assert utils.networking.get_external_ip()
```
eduardogr, line 71: great! very nice!

bittensor/utils/networking.py
```diff
@@ -92,59 +92,39 @@ def get_external_ip() -> str:
...
+        get_external_ip_from_ipv6, 
+        get_external_ip_from_wiki
+    ]:
+        try:
+            external_ip = get_ip()
+            if external_ip != None: 
+                if not isinstance(ip_to_int(external_ip), int): raise(ValueError('IP could not be converted to int.'))
+                return str(external_ip)
+        except Exception:
+            logger.info(f'Failed to get external IP from {get_ip.__name__}')
```
eduardogr: very nice!

bittensor/utils/networking.py
```diff
@@ -92,59 +92,39 @@ def get_external_ip() -> str:
...
+            
+
+            pass
```
eduardogr, line 124: do we need this statement??

eduardogr (approved):
After solving comments, LGTM!

Eugene-hu (approved):

PR #1253, opened 2023-04-06
Bit 102 fix unbounded nonce dictionary

isabella618033:
#7 Saving and loading nonce dictionary upon restart of server 
#5 Nonce should always be larger then previous stored nonce and smaller then first nonce + 10 years of time
#3 Removing insecure prints from keyfile error logging


bittensor/_axon/__init__.py
```diff
@@ -423,15 +447,26 @@ def check_signature(
...
         if endpoint_key in self.nonces.keys():
-            previous_nonce = self.nonces[endpoint_key]
+            first_nonce = self.nonces[endpoint_key][0]
+            previous_nonce = self.nonces[endpoint_key][1]
             # Nonces must be strictly monotonic over time.
-            if nonce <= previous_nonce:
-                raise Exception("Nonce is too small")
-
+            if nonce <= previous_nonce or nonce > first_nonce + self.upper_bound:
+                raise Exception(f"Nonce {nonce} is smaller or equal to previous nonce {previous_nonce} or larger then {first_nonce + 3.154e+17}.")
```
camfairchild:
```suggestion
                raise Exception(f"Nonce {nonce} is smaller or equal to previous nonce {previous_nonce} or larger than {first_nonce + 3.154e+17}.")
```

camfairchild (approved):
LGTM

bittensor/_axon/__init__.py
isabella618033: oh nooooooo

bittensor/_axon/__init__.py
```diff
@@ -361,18 +364,39 @@ def __init__(
...
         super().__init__()
+        self.path = f"{path}/nonces_dict.txt" if path != None else None
         self.nonces = {}
         self.blacklist = blacklist
         self.receiver_hotkey = receiver_hotkey
+        self.load()
+        self.upper_bound = int(3.154e+17) # 10 years in nano second
+
+    def load(self):
+        if self.path != None and os.path.exists(self.path):
```
eduardogr: As far as i know `is` is more encouraged that `==` or `!=` for this cases. 
Take a look at the references, I think it's a good scenario to compare using `is` instead of `==` 

I leave you here some references to that:
- https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or
- https://codeql.github.com/codeql-query-help/python/py-test-equals-none/

bittensor/_axon/__init__.py
```diff
@@ -361,18 +364,39 @@ def __init__(
...
+            with open(self.path) as nonces_file:
+                self.nonces = json.load(nonces_file)
+        else:
+            self.nonces = {}
+
+    def save(self):
+        if self.path != None:
```
eduardogr: same here

bittensor/_axon/__init__.py
```diff
@@ -31,6 +30,7 @@
 import grpc
 from substrateinterface import Keypair
 
+import json
 import bittensor
```
eduardogr, line 34: I know that you did not change this in this PR, but as I mentioned in other PR,
What do you think about changing this absolute import to a relative one since we are inside the `bittensor` module

tests/unit_tests/bittensor_tests/test_axon.py
```diff
@@ -1263,6 +1265,105 @@ def test_external_ip_port_set_full_address_internal(self):
         assert f'{internal_ip}:{internal_port}' == full_address1, f'{internal_ip}:{internal_port} is not eq to {full_address1}'
         assert f'{external_ip}:{external_port}' != full_address1, f'{external_ip}:{external_port} is eq to {full_address1}'
 
+class TestInterceptor(unittest.TestCase):
+
+    def setUp(self):
+        try:
+            os.remove('./nonces_dict.txt')
+        except:
+            Exception('No nonce_dict to be removed.')
```
eduardogr: I know that we are not doing this in so many different tests but...I am going to share my point of view with this kind of practices in terms of unit testing.

In order to test this  piece of code it is pointed out as good practices to avoid to manage exceptions and also that we should not allow to have side effects generally, like the creation of the file by the code being tested.
In order to do it so, we should mock that piece of code so the file is not created, so we do not have to remove it. Tests should be run in a way that no side effects are produced.

What do you think about this?
If you agree and like this, feel free to reach me out to address this together in case you do not know how to proceed there.

eduardogr:
left some comments, thanks for this 👍🏽

bittensor/_axon/__init__.py
isabella618033: Hey, @eduardogr cause the same practise have been widely used across the project, so i think it might be better if we can change them in a separate PR.

Cause axon is a rather big submodule and it is using a lots of other bittensor submodules like prometheus /  wallet / proto / threadpool / config / grpc.

tests/unit_tests/bittensor_tests/test_axon.py
isabella618033: The file was created from the [test_saving](https://github.com/opentensor/bittensor/pull/1253/files/68326b438dcecf317a9f1d1ec916e7245e6a3c61#diff-9bb2374a00ee2f68a526e31be4b02d7c663f17bad6e67b2693194fa7f4852a08R1287) and [test_loading](https://github.com/opentensor/bittensor/pull/1253/files/68326b438dcecf317a9f1d1ec916e7245e6a3c61#diff-9bb2374a00ee2f68a526e31be4b02d7c663f17bad6e67b2693194fa7f4852a08R1296) function that is targeting at the [saving](https://github.com/opentensor/bittensor/pull/1253/files#diff-369f767fecee3a55495f39d9f75d496a6a61cac45c0ea2491760d190d38695d3R393) and [loading](https://github.com/opentensor/bittensor/pull/1253/files/68326b438dcecf317a9f1d1ec916e7245e6a3c61#diff-369f767fecee3a55495f39d9f75d496a6a61cac45c0ea2491760d190d38695d3R386) function here which are some basic file reading/writing function. So I am not sure how can we test it with mock, or is it actually not necessary to get these 2 function tested?

bittensor/_axon/__init__.py
eduardogr: ok, i see, fair enough

tests/unit_tests/bittensor_tests/test_axon.py
eduardogr: yeah I know it's beign created. In terms of logic this is Ok. For the sake of simplicity and merge this, if you want, just proceed as it is, it's fine.

From the testing point of view, when testing that code that creates the file, we should mock that so no file is created. And we just intereact with the Mock. If you want just merge as it is and let's create a tech-deb task for that

eduardogr (approved):

tests/unit_tests/bittensor_tests/test_axon.py
isabella618033: Thanks edu, i understand, let me change the test to mock before merging it:)

PR #1259, opened 2023-04-13
Rename requiremenets.txt to requirements.txt

KMFODA:
Fixing a typo in the `requirements.txt` file

PR #1261, opened 2023-04-13
Adding support for Pythia-Chat-Base-7B, GPT-NeoXT-Chat-Base-20B

ViktorThink:
Adding support for using Pythia-Chat-Base-7B and GPT-NeoXT-Chat-Base-20B for mining.

The previous pythia example doesn't work.

Fixes typo mentioned in: #1259 

Tested on netuid 91


ViktorThink:
Now tested on netuid 1 and working as intended.

PR #1272, opened 2023-04-17
Bit 90 authenicate grpc request

Eugene-hu:
-Adds request authentication to bittensor messaging
-Now, the sum of input is checked as part of the signature. This ensures that request tensors has not been tampered with during transport.


bittensor/_axon/__init__.py
```diff
@@ -427,34 +432,65 @@ def check_signature(
...
 
+    def request_deserialization(
+            self,
+            request
+    ):
+        r"""deserialization of request in payload. Used for authenication"""
+        
+        try: 
+            synapse = bittensor.synapse.deserialize(request.synapses[0])
+            request_matrix = synapse.deserialize_forward_request_tensor(request.tensors[0])
```
adriansmares, line 449: Shouldn't take this into consideration if this is a `Forward` or a `Backward` ? You have that in the `method_name` at the call site.

PR #1286, opened 2023-04-20
BIT-261 add new versions in files and adding execution permissions fo…

eduardogr:
## Description
- Adding - missed - new version to the expected files
- Adding - expected - execution permisions for the file `scripts/release/github_release.sh`


eduardogr:
seems like it was already done.... my bad

PR #1293, opened 2023-04-21
Fix Bug with Miner recieving messages in list of strings instead of list of dicts

### LINKED ISSUE #1292: Miner error on netuid 1 - string indicies must be integers
When running a miner on netuid 1, we get this output with current text_prompting branch:

image

It should cause an error for any miner, since what happens is that before the bittensor.BasePromptingMiner (in this case in neurons/text/prompting/miners/pythia/neuron.py) previously recieved the "messages" argument in the form of a list of dictionaries, but now gets it in the form of a list with strings.

To clarify,efore, we got something like this:

[{'role': 'system', 'content': '\nYou are designed to assist with a wide range of tasks, from answering simple questions to providing in-depth explanations and discussions on a wide range of topics.\n'}, {'role': 'user', 'content': '\nAsk me a random question about anything. Make the question very domain specific. Do not include the answer in the question.\n'}]

Now we get something like this:

['{"role": "system", "content": "\\nYou are designed to assist with a wide range of tasks, from answering simple questions to providing in-depth explanations and discussions on a wide range of topics.\\n"}', '{"role": "user", "content": "What is the function of the Golgi apparatus in a eukaryotic cell?"}']

While making a more permanent fix to this issue, I can confirm that this quickfix works:

Adding this to the start of the  forward() function in the miner, in this case in this case in neurons/text/prompting/miners/pythia/neuron.py:

```
            import json
            messages = [json.loads(item) for item in messages]
```

It takes all the strings in the messages variable and turns them into dictionaries.

ViktorThink:
This PR fixes #1292 

On the previous version of bittensor/_synapse/text_prompting/synapse.py (https://github.com/opentensor/bittensor/blob/eff272409704d801ce337a2ce850ba49368848d5/bittensor/_synapse/text_prompting/synapse.py)
On line 57 it is:

`formatted_messages = [ json.loads(message) for message in forward_call.messages ]`

The most recent update to the file made yesterday, had the json.loads function removed, causing the error:

`self.formatted_messages = [ message for message in self.messages ]`

So I simply added import of json and json loading the messages.

I will test the solution today to make doubly sure it solves the issue.

### DIFF
--- a/bittensor/_synapse/text_prompting/synapse.py
+++ b/bittensor/_synapse/text_prompting/synapse.py
@@ -21,6 +21,7 @@ import bittensor
 
 from typing import List, Dict, Union, Callable
 from abc import ABC, abstractmethod
+import json
 
 class SynapseForward( bittensor.SynapseCall ):
     name: str = "text_prompting_forward"
@@ -35,7 +36,7 @@ class SynapseForward( bittensor.SynapseCall ):
         ):
         super().__init__( synapse = synapse, request_proto = request_proto )
         self.messages = request_proto.messages
-        self.formatted_messages = [ message for message in self.messages ]
+        self.formatted_messages = [ json.loads(message) for message in self.messages ]
         self.forward_callback = forward_callback
 
     def apply( self ):


ViktorThink:
I have tested now and it works correctly:

`messages: [{'role': 'system', 'content': '\nYou are designed to assist with a wide range of tasks, from answering simple questions to providing in-depth explanations and discussions on a wide range of topics.\n'}, {'role': 'user', 'content': '\nAsk me a random question about anything. Make the question very domain specific. Do not include the answer in the question.\n'}]`

joeylegere (approved):

PR #1300, opened 2023-04-22
Authinterceptor signature typehints

steffencruz:
Small bug fix which corrects the return type hint

### DIFF
--- a/bittensor/_axon/__init__.py
+++ b/bittensor/_axon/__init__.py
@@ -284,7 +284,7 @@ class AuthInterceptor(grpc.ServerInterceptor):
         self.receiver_hotkey = receiver_hotkey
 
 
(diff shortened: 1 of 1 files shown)


steffencruz:
Also `def intercept_service(self, continuation, handler_call_details):` is missing type hints

bittensor/_axon/__init__.py
```diff
@@ -284,7 +284,7 @@ def __init__(
         self.receiver_hotkey = receiver_hotkey
 
 
-    def parse_signature_v2(self, signature: str) -> Union[Tuple[int, str, str, str, int], None]:
+    def parse_signature_v2(self, signature: str) -> Union[Tuple[int, str, str, str], None]:
```
camfairchild:
```suggestion
    def parse_signature_v2(self, signature: str) -> Optional[Tuple[int, str, str, str]]:
```

PR #1304, opened 2023-04-25
validator update

Eugene-hu:
- added loading weights
- fixes moving average
- adds additional features for logging
- reduces the memory of the history

### DIFF
--- a/bittensor/_metagraph/__init__.py
+++ b/bittensor/_metagraph/__init__.py
@@ -124,7 +124,6 @@ class metagraph( torch.nn.Module ):
         self.version = torch.nn.Parameter( torch.tensor( [bittensor.__version_as_int__], dtype=torch.int64 ), requires_grad=False )
         self.block = torch.nn.Parameter( torch.tensor( subtensor.block, dtype=torch.int64 ), requires_grad=False )
         self.uids = torch.nn.Parameter( torch.tensor( [ neuron.uid for neuron in self.neurons ], dtype=torch.int64 ), requires_grad=False )
-        self.emission = torch.nn.Parameter( torch.tensor( [ neuron.emission for neuron in self.neurons ], dtype=torch.float32 ), requires_grad=False )
         self.trust = torch.nn.Parameter( torch.tensor( [ neuron.trust for neuron in self.neurons ], dtype=torch.float32 ), requires_grad=False )
         self.consensus = torch.nn.Parameter( torch.tensor( [ neuron.consensus for neuron in self.neurons ], dtype=torch.float32 ), requires_grad=False )
         self.incentive = torch.nn.Parameter( torch.tensor( [ neuron.incentive for neuron in self.neurons ], dtype=torch.float32 ), requires_grad=False )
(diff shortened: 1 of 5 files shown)


bittensor/_metagraph/__init__.py
```diff
@@ -124,7 +124,6 @@ def sync ( self, block: Optional[int] = None, lite: bool = True ) -> 'metagraph'
         self.version = torch.nn.Parameter( torch.tensor( [bittensor.__version_as_int__], dtype=torch.int64 ), requires_grad=False )
         self.block = torch.nn.Parameter( torch.tensor( subtensor.block, dtype=torch.int64 ), requires_grad=False )
         self.uids = torch.nn.Parameter( torch.tensor( [ neuron.uid for neuron in self.neurons ], dtype=torch.int64 ), requires_grad=False )
-        self.emission = torch.nn.Parameter( torch.tensor( [ neuron.emission for neuron in self.neurons ], dtype=torch.float32 ), requires_grad=False )
```
opentaco, line 127: Why is the emission removed from the metagraph?

neurons/text/prompting/validators/core/gating.py
```diff
@@ -113,7 +113,8 @@ def forward( self, message: str ) -> 'torch.FloatTensor':
                     Scores for each uids as output by the gating model.
         """
         inputs = self.tokenizer( message, return_tensors="pt" ).to( self.device )
-        hidden_states = self.model( **inputs ).last_hidden_state[0, -1, :]
+        with torch.no_grad():
```
opentaco, line 116: Would the gating model not need gradients here in order to be trained?

opentaco (approved):

bittensor/_metagraph/__init__.py
Eugene-hu: This was duplicated a bit lower in the metagraph

neurons/text/prompting/validators/core/gating.py
Eugene-hu: The idea is that we don't have enough data to fully fine-tune the transformer block. Instead, we should only train the last linear layer and freeze the current state of the transformer.

neurons/text/prompting/validators/core/neuron.py
```diff
@@ -135,6 +134,9 @@ def __init__( self ):
         self.uid = self.wallet.get_uid( subtensor = self.subtensor, netuid = self.config.netuid )
         self.tokenizer = AutoTokenizer.from_pretrained( 'EleutherAI/gpt-j-6b' )
 
+        self.moving_averaged_scores = torch.zeros((self.metagraph.n)).to( self.device )
+        self.alpha = 0.99
```
isabella618033, line 138: alpha = 0.99 seems a bit high, think we used to have alpha = 0.1 (0.9)

neurons/text/prompting/validators/core/neuron.py
```diff
@@ -416,9 +434,12 @@ def train( self ):
                 )
 
                 # Resync metagraph before returning. (sync every 15 min or ~75 blocks)
-                if last_epoch_block % 10 == 0:
+                if self.subtensor.block % 10 == 0:
```
isabella618033, line 437: I think %10 would be syncing every 2 mins, wouldnt that be a bit frequent?

isabella618033 (approved):
LGTM!

shibshib (approved):

neurons/text/prompting/validators/core/neuron.py
Eugene-hu: hmm good point, this might need to change as we increase the number of uids. Lets put in a ticket and do deeper analysis once the network is stable

PR #1305, opened 2023-04-26
Text prompting asyncio fix

isabella618033:
Fixed the error below by avoiding running dendrite_pool forward for both inference and training at the same time that shares the same grpc resourses. 
-> Added  priority threadpool with max_worker = 1 inside dendrite_pool so that only 1 forward/backward would be processed at a time. 

```
Exception in callback PollerCompletionQueue._handle_events(<_UnixSelecto...e debug=False>)()
handle: <Handle PollerCompletionQueue._handle_events(<_UnixSelecto...e debug=False>)()>
Traceback (most recent call last):
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/completion_queue.pyx.pxi", line 147, in grpc._cython.cygrpc.PollerCompletionQueue._handle_events 
blockingIOError: [Errno 11] Resource temporarily unavailable 
```


shibshib (approved):

neurons/text/prompting/validators/core/neuron.py
```diff
@@ -348,7 +348,8 @@ def inference(
             roles = roles, 
             messages = contents, 
             uids = uids, 
-            timeout = timeout
+            timeout = timeout,
+            priority = 2 # the priority of inference is always higher then training
```
Eugene-hu: Should inference take priority over training? Wouldnt this create issues where the model is always in inference?

Eugene-hu (approved):

PR #1317, opened 2023-05-09
Additional fix for rewards and growth of network

Eugene-hu:
- Adds fix for reward system by specifying about question and answer
- Adds in preliminary fix for the gating model

### DIFF
--- a/neurons/text/prompting/validators/core/neuron.py
+++ b/neurons/text/prompting/validators/core/neuron.py
@@ -541,6 +541,7 @@ class neuron:
                 # Resync metagraph before returning. (sync every 15 min or ~75 blocks)
                 if self.subtensor.block - self.last_sync > 100:
                     self.metagraph.sync()
+                    self.last_sync = self.subtensor.block
                     self.save()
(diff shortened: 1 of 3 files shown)


neurons/text/prompting/validators/core/neuron.py
```diff
@@ -550,6 +552,9 @@ def train( self ):
                     self.my_nominators = { nomin[0]: nomin[1] for nomin in delegates[0][0].nominators } if len(delegates) else {}
                     self.check_weights()
 
+                    if self.metagraph.n > self.gating_model.num_uids:
```
isabella618033, line 555: is this just a place holder? cause i think it is not good if we have to wipe out the weight every time there is a new guy joining xd 

Or we can initiate a over sized linear layer at the beginning.

isabella618033 (approved):

neurons/text/prompting/validators/core/neuron.py
Eugene-hu: yeap, but @dl-guru is looking for a better solution to this problem, this is simple patch for the issue. We shouldn't wipe out the weights everytime

PR #1328, opened 2023-05-14
metagraph sync warning update. avoid potential crash

ifrit98:
For some reason, if on full sync and there are no neurons. the `torch.stack()` calls would result in an uncaught exception. 

Figured we should log/warn the user this is happening or we can raise an exception if desired. Just ran into this while writing the unittest for the metagraph.


Eugene-hu (approved):
LGTM

PR #1330, opened 2023-05-15
remove server var def from class struct, add check in stop() to compe…

ifrit98:
…nsate for error on incorrect calling of bittensor.axon()


Eugene-hu (approved):
LGTM

Eugene-hu (approved):
LGTM

PR #1338, opened 2023-05-18
catches errors from failed metagraph

CreativeBuilds:
sometimes the metagraph testnet will randomly drop out causing a crash
instead of crashing, it now retries with an exponential wait and stops after 4 retry attempts


ifrit98:
LGTM

PR #1340, opened 2023-05-19
Adds strength and seed vars for img2img

CreativeBuilds:
The strength variable was never added to image2image resulting in the default value of 0.75 to always apply to generations. 

This PR fixes that and updates the test file for the miner to show off the difference between high and low strength values.


ifrit98 (approved):
LGTM!

PR #1343, opened 2023-05-19
add wizard lm vicuna uncensored miner, dromedary miners, fastchat_t5, dolly12b miner

ifrit98:


ifrit98:
Moved to: https://github.com/opentensor/bittensor/pull/1346

PR #1350, opened 2023-05-24
re-add taco's addition of QuadraticVotingPower subtensor query.

ifrit98:
Moved to merge with release/5.1.0.

### DIFF
--- a/bittensor/_subtensor/chain_data.py
+++ b/bittensor/_subtensor/chain_data.py
@@ -45,6 +45,7 @@ custom_rpc_type_registry = {
                 ["max_weights_limit", "Compact<u16>"],
                 ["scaling_law_power", "Compact<u16>"],
                 ["synergy_scaling_law_power", "Compact<u16>"],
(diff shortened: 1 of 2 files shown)


mrseeker:
What would be the benefit and dangers of this PR? The benefit would be that validators with a lower stake have more power over the network, while those with a higher stake will have less. However, is that what we want?

This brings me to the "divide and conquer" question: What if I split the stake over multiple validators? Would I gain more or less than keeping it on a single validator? And if I earn more by breaking, then it would be more beneficial to split funds over multiple accounts while using a single validator and simultaneously setting the same weights on all accounts.

What benefits does the solution give over other changes?

shibshib:
Hey @ifrit98 what's the status here? Is this PR still relevant or is it stale now?

ifrit98:
Was attempting to prepare it for last release, I think this is stale now. 

Before closing: @opentaco, is this something you think we should push to include in next release or can we safely disregard?

PR #1353, opened 2023-05-25
Calculates elapsed time at the end of dendrite call apply

steffencruz:
Tested (manually) with text prompting validator neuron. Useful for analysis of dendrite response rates, to be used for `timeout` hyperparameter optimization.


ifrit98 (approved):
LGTM, I see no potential issues.

Eugene-hu (commented):
LGTM, but lets merged into release/5.1.0 and not master directly

PR #1354, opened 2023-05-26
Llama Requirements Added

asifehmad:
Added the Transformer required for Llama, and sentencepiece module which is also required to load Llama.


ifrit98:
@asifehmad Thanks for this fix. Would you mind reopening as 1 PR (combined with your other one) merging into release/5.1.0?

Thanks.

asifehmad:
> @asifehmad Thanks for this fix. Would you mind reopening as 1 PR (combined with your other one) merging into release/5.1.0?
> 
> Thanks.

Sure!

PR #1357, opened 2023-05-26
Added Open_Llama Miner Requirements and Replaced Transformer 

asifehmad:
Added the Transformer required for Llama, and sentencepiece module which is also required to load Llama in the `requirements.txt` file.
Replaced AutoModelForCausalLM and AutoTokenizer with LlamaForCausalLM and LlamaTokenizer, respectively, in the `neuron.py` file.


ifrit98:
Closing and merging your changes in a new PR because we did an update on how we do huggingface miners (created an ABC).

See - https://github.com/opentensor/bittensor/pull/1364

PR #1359, opened 2023-05-28
Reset scores for validators

adriansmares:
#### Summary

References https://github.com/opentensor/bittensor/pull/1307

This PR proposes that the weights for UIDs with a validator permit are reset on every weight check round. The reasoning is that a UID may gain a validator permit after it has been queried in the past, which in turn allows it to 'freeze' its current weights and avoid any form of decay.

#### Changes

- Reset the moving averages for keys which have a validator permit.
  - The reasoning here is that keys with a validator permit are not queried during the training process - their weights do not decay and are instead stable, which affects the evaluation process of the validators.

### DIFF
--- a/neurons/text/prompting/validators/core/neuron.py
+++ b/neurons/text/prompting/validators/core/neuron.py
@@ -712,6 +712,8 @@ class neuron:
         for uid, hotkey in enumerate( self.hotkeys ):
             if hotkey != self.metagraph.hotkeys[ uid ]:
                 self.moving_averaged_scores[ uid ] = 0 #hotkey has been replaced
+            if self.metagraph.validator_permit[ uid ]:
+                self.moving_averaged_scores[ uid ] = 0 # hotkey has validation rights
         if len(self.hotkeys) < len(self.metagraph.hotkeys):
             new_moving_average  = torch.zeros((self.metagraph.n)).to( self.device )
             new_moving_average[:len(self.hotkeys)] = self.moving_averaged_scores


Eugene-hu:
hey @adriansmares, I don't believe the uid weights would freeze. If they are only running a validator, then their weights should decay as a server that isn't replying. All uids are randomly queried on the network, and we pass 0s for any uids that are unsuccessful in replying. (https://github.com/opentensor/bittensor/blob/master/neurons/text/prompting/validators/core/neuron.py#L367)

adriansmares:
The UIDs with validator permit are completely skipped from the `available_uids` - there is no unsuccessful attempt towards the validator UIDs such that they would get a zero score, and in turn decay.

https://github.com/opentensor/bittensor/blob/76e347c951ede506dfe13d471335bc7b45f18392/neurons/text/prompting/validators/core/neuron.py#L279

This means that validators will never be part of either the successful UIDs, or the unsuccessful UIDs. This means that they get to keep their original score during the first `scatter`, and they won't get zeroed in the second `scatter` either, because they are not unsuccessful.

Take a simplified network of 3 UIDs: 1st is the successful one,  the 2nd one is the unsuccessful one, and the 3rd is the validator. The score of the validator is untouched:

```
>>> import torch
>>> moving_average_scores=torch.tensor([125, 150, 200])
>>> event_uids=torch.tensor([0])
>>> unsuccesfull_uids=torch.tensor([1])
>>> scattered_rewards = moving_average_scores.scatter(0, event_uids, torch.tensor([100]))
>>> scattered_rewards
tensor([100, 150, 200])
>>> scattered_rewards = scattered_rewards.scatter(0, unsuccesfull_uids, torch.zeros_like(unsuccesfull_uids))
>>> scattered_rewards
tensor([100,   0, 200])
```

Eugene-hu:
ahh okay, yes, the query uids were changed due to the nature of forwarding within the validators. Very nice catch @adriansmares!

Eugene-hu (approved):

ifrit98 (approved):
LGTM! Great eye @adriansmares 👁️

PR #1362, opened 2023-05-29
blacklist/priority makeover, refactor base_miner to /_neuron

ifrit98:
- separates out _blacklist and _priority into their own dirs.
- clean up blacklist, add vpermit check
- move base_prompting_miner.py to _neuron.

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -209,8 +209,8 @@ from bittensor._dendrite.text_prompting.dendrite import TextPromptingDendrite as
 from bittensor._dendrite.text_prompting.dendrite_pool import TextPromptingDendritePool as text_prompting_pool
 
 # ---- Base Miners -----
-from bittensor._neuron.base_miner_neuron import BaseMinerNeuron as base_miner_neuron
-from bittensor._neuron.base_validator import BaseValidator as base_validator
+from bittensor._neuron.base_miner_neuron import BaseMinerNeuron
+from bittensor._neuron.base_validator import BaseValidator
 from bittensor._neuron.base_prompting_miner import BasePromptingMiner
 
 # ---- Errors and Exceptions -----
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -168,6 +168,8 @@ from bittensor._tokenizer import tokenizer as tokenizer
 from bittensor._serializer import serializer as serializer
 from bittensor._dataset import dataset as dataset
 from bittensor._threadpool import prioritythreadpool as prioritythreadpool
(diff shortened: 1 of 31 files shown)


bittensor/__init__.py
```diff
@@ -207,7 +209,9 @@ def turn_console_off():
 from bittensor._dendrite.text_prompting.dendrite_pool import TextPromptingDendritePool as text_prompting_pool
 
 # ---- Base Miners -----
-from bittensor._synapse.text_prompting.miner import BasePromptingMiner
+from bittensor._neuron.base_miner_neuron import BaseMinerNeuron as base_miner_neuron
```
Eugene-hu: small nit: BasePromptingMiner and base_miner_neuron are two separate conventions. We should stick with one, ie BaseMinerNeuron.

bittensor/_neuron/base_validator.py
```diff
@@ -0,0 +1,167 @@
...
+import torch
+import threading
+import argparse
+import bittensor
+
+from rich import print
+from typing import List, Dict, Union, Tuple, Optional
+from datetime import datetime
+
+class BaseValidator:
```
Eugene-hu, line 29: I am little confused about the inheritance here; what is the use case for this BaseValidator? is it being inherited by the core_validator?

bittensor/_neuron/base_miner_neuron.py
```diff
@@ -0,0 +1,212 @@
...
+import torch
+import threading
+import argparse
+import bittensor
+
+from rich import print
+from typing import Union, Tuple
+from datetime import datetime
+
+class BaseMinerNeuron:
```
Eugene-hu, line 29: Similar question here; is this the base miner for the servers?

ifrit98: We will have:
- TextToImage
- SpeechToText
- TextToCompletion (This is currently text_prompting and needs to extend this, working on it now)

So the main idea is to extend each multi-modal class from this `BaseMinerNeuron` that way we don't have entirely bespoke submodules.

bittensor/__init__.py
ifrit98: Fixed with commit: https://github.com/opentensor/bittensor/commit/07fad43a72bcdcd3c941ef84ca39c72db2e1c9a3

bittensor/_neuron/base_validator.py
ifrit98: Yes, it should be inherited by all future Validator subtypes. (see next comment for details)

Eugene-hu (approved):

Eugene-hu (approved):

PR #1375, opened 2023-06-01
Fix docker and pypi upload commands

eduardogr:
## What will you find here?

- using sudo for docker commands
- change twine command for pip package update
    - now we are using an API token instead of a username/passwd so we expect a .pypirc file with the needed info for bittensor release


ifrit98:
Beautiful.  Thanks 🙏

ifrit98 (approved):

camfairchild (approved):

camfairchild (approved):

PR #1378, opened 2023-06-02
minimal requirements, delete unused.

ifrit98:
Can ditch ~12 completely unused requirements.


requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
...
-rich==12.5.1
 retry==0.9.2
+rich==12.5.1
 requests==2.25.0
 scalecodec==1.2.0
-sentencepiece==0.1.97
+substrate-interface==1.5.0
 termcolor==2.1.1
-torch==1.13.1
+torch>=1.13.1
```
eduardogr, line 33: please, do not specify an unlimited range of versions,
_**Why am i saying this?**_
we should deliver code that is tested and which behavior we know.
with this kind of specifications we can end up with bittensor packages installing versions of torch that we did not tested or that do not work at all at that point.

So, to proceed with this... you may specify an upper bound if you want to provide a range of versions (if that is tested) but, please, do not add an unlimited range of versions because it can lead us to people installing bittensor and not working at all because it may triggers the installation of a version that is not compatible. We should test the versions we integrate in our requirements.

eduardogr:
can u comment in the PR the packages you have deleted? 
since the diff is lines-based is a bit laborious to see what's being removed.

Thanks for this, this was very much needed

requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
...
-tqdm==4.64.1
-qqdm==0.0.7
-ansible_vault>=2.1
-substrate-interface==1.5.0
-jsonschema[format-nongpl]>=4.14.0,<=4.17.0
-markupsafe==2.0.1
-langchain>=0.0.132
-datasets==2.12.0
\ No newline at end of file
+tqdm>=4.64.1
```
eduardogr, line 35: same here

requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
+ansible_vault>=2.1
```
eduardogr, line 1: same here

requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
...
-jinja2==3.0
 fuzzywuzzy==0.18.0
-google-api-python-client==2.7.0
-python-levenshtein==0.12.1
 grpcio==1.42.0
-grpcio-tools==1.42.0
 hypothesis==6.47.4
+idna<3,>=2.5
+jsonschema[format-nongpl]>=4.14.0,<=4.17.0
+langchain>=0.0.132
```
eduardogr, line 11: same here

requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
...
 loguru==0.6.0
 msgpack==1.0.4
 msgpack-numpy==0.4.7.1
 miniupnpc==2.0.2
 munch==2.5.0
 nest_asyncio==1.5.6
 netaddr==0.8.0
+numpy>=1.24.2
```
eduardogr, line 19: same here

requirements/prod.txt
```diff
@@ -1,48 +1,36 @@
...
 retry==0.9.2
+rich==12.5.1
 requests==2.25.0
 scalecodec==1.2.0
-sentencepiece==0.1.97
+substrate-interface==1.5.0
 termcolor==2.1.1
-torch==1.13.1
+torch>=1.13.1
 transformers>=4.20.1
```
eduardogr, line 34: same here

PR #1379, opened 2023-06-02
add default 1024 max stake limit for querying UIDs with vpermit.

ifrit98:
- Small change, make sure that UIDs with a vpermit but under stake limit (1024 by default) are still queried.

### DIFF
--- a/neurons/text/prompting/validators/core/neuron.py
+++ b/neurons/text/prompting/validators/core/neuron.py
@@ -330,7 +330,7 @@ class neuron:
         # Find the available `uids` that are currently serving.
         # If `topk` is larger than the number of available `uids`, set `topk` to the number of available `uids`.
         # Check if we have vpermit and if we do, ensure query only UIDs with less than vpermit_tao_limit.
-        candidate_uids = [uid for uid, ax in enumerate(self.metagraph.axons) if ax.is_serving and (self.metagraph.validator_permit[uid] and self.metagraph.S[uid] < self.config.neuron.vpermit_tao_limit or not self.metagraph.validator_permit[uid])]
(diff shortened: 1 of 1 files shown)


Eugene-hu (approved):

neurons/text/prompting/validators/core/neuron.py
```diff
@@ -328,12 +329,14 @@ def forward(
         # Set `topk` to the number of items in `self.metagraph.n` if `topk` is not provided or is -1.
         # Find the available `uids` that are currently serving.
         # If `topk` is larger than the number of available `uids`, set `topk` to the number of available `uids`.
-        available_uids = torch.tensor( [ uid for uid, ax in enumerate( self.metagraph.axons ) if (ax.is_serving) and (not self.metagraph.validator_permit[uid]) ], dtype = torch.int64 ).to( self.device )
+        # Check if we have vpermit and if we do, ensure query only UIDs with less than vpermit_tao_limit.
+        candidate_uids = [uid for uid, ax in enumerate(self.metagraph.axons) if ax.is_serving and ((self.metagraph.validator_permit[uid] and self.metagraph.S[uid] < self.config.neuron.vpermit_tao_limit) or not self.metagraph.validator_permit[uid])]
```
adriansmares: This condition should also be updated in `check_weights`. 

Can also be simplified to `not self.metagraph.validator_permit[uid] or self.metagraph.S[uid] < self.config.neuron.vpermit_tao_limit`.

ifrit98: Assume it would be as simple as changing the 2nd if condition in `check_weights` to:

```python
for uid, hotkey in enumerate( self.hotkeys ):
    if hotkey != self.metagraph.hotkeys[ uid ]:
        self.moving_averaged_scores[ uid ] = 0 #hotkey has been replaced
    if self.metagraph.validator_permit[ uid ] and self.metagraph.S[ uid ] < self.config.neuron.vpermit_tao_limit:
        self.moving_averaged_scores[ uid ] = 0 # hotkey has validation rights and is below the tao limit
```

Let me know if this is an incorrect assumption.

unconst (approved):

PR #1392, opened 2023-06-16
Textual Inversion and LoRA Support

CreativeBuilds:
allows text2img miners to drop in models for both textual inversion and LoRA

as of right now, miners have no way of expressing which models are available, a TODO in the future is adding an endpoint which enables this


ifrit98 (approved):
LGTM!

PR #1394, opened 2023-06-17
Fix axon requests signature using metadata

unconst:
Fix checks signature of requester during every call to text_prompting forward synapse using the signature generated on GRPC TCP channel creation. 

This removes the possibility for man in the middle attacks where a malicious actor is capable of spoofing the keys of network validators.


unconst:
Honours goes to Adrianmares for the majority of this fix.

adriansmares (approved):

ifrit98 (commented):
Nice! This looks good to me.

Should we also update the int version also in the Auth Interceptor to 510?


https://github.com/opentensor/bittensor/blob/8e3f932d4eede4fe34369203822f5939d3a40fbc/bittensor/_axon/__init__.py#L308

unconst:
Well throw that in a different PR @ifrit98

ifrit98 (approved):

shibshib (approved):

camfairchild (approved):

bittensor/_synapse/synapse.py
```diff
@@ -35,13 +35,22 @@ class SynapseCall( ABC ):
...
+            _,
+            _,
+        ) = synapse.axon.auth_interceptor.parse_signature(metadata)
+        
         self.completed = False
         self.start_time = time.time()
         self.timeout = request_proto.timeout
         self.src_version = request_proto.version
-        self.src_hotkey = request_proto.hotkey
+        self.src_hotkey = sender_hotkey
```
ifrit98, line 53: This indeed prevents spoofing the source hotkey, but still allows the request to go through.

Do we want to check somewhere in the synapse that the request hotkey or still let the request pass?

e.g. in the apply() call:
```python
    def apply( self, call: SynapseCall ) -> object:
        ....
            if call.src_hotkey != call.request_hotkey:
                call.return_code = bittensor.proto.ReturnCode.InvalidRequestHotkey
                call.return_message = 'Invalid request hotkey'
                raise InvalidRequestHotkeyError
```

See: https://github.com/opentensor/bittensor/pull/1396

adriansmares: The thing is that the `request_hotkey` (I am not sure this is the path in the object - the field is `hotkey` in the request proto) field should be removed completely. Older clients do not have this verification anyway, but do need it, while newer clients ignore it completely, and don't need the field - you are not preventing anything by raising the exception.

Having an extra return code for such cases is superflous and just pollutes the codebase with cases which should be completely removed in the end anyway.

ifrit98: Agreed. Thanks for the explanation.

PR #1395, opened 2023-06-18
Update authint version

ifrit98:
Updates the bittensor version that authinterceptor checks against from 370 -> 510.


camfairchild:
~~Should this be into `master` or #1394 be into `new_staging` ?~~
edit: nvm

camfairchild (approved):

PR #1396, opened 2023-06-19
Authint blacklist fix 

ifrit98:
This PR:
- Cleanly separates out blacklist functions for Synapses [whitelist keys, blacklist keys] and for the AuthInterceptor [vpermit, stake check]. (Thanks @Eugene-hu for this idea)
- Prevents further DDoS attacks via hotkey spoofing by checking the hotkey from grpc context against the request hotkey in synapse.apply() call. (Credit goes to @adriansmares  for this one)
- Creates a new InvalidRequestHotkey error code to the protos (26) so we can easily spot this behavior.


bittensor/_axon/__init__.py
```diff
@@ -280,19 +274,19 @@ class AuthInterceptor(grpc.ServerInterceptor):
...
-            black_list (Function, `optional`):
-                black list function that prevents certain pubkeys from sending messages
         """
         super().__init__()
         self.nonces = {}
-        self.blacklist = blacklist
+        self.blacklister = bittensor.auth_blacklister()
+        self.blacklist = self.blacklister.blacklist
         self.receiver_hotkey = receiver_hotkey
+        self.metagraph = metagraph
```
Eugene-hu, line 295: Will this be the same metagraph object as the main metagraph in server? ie. when the metagraph gets synced in the server, will this metagraph also be update?

bittensor/_axon/__init__.py
```diff
@@ -280,19 +274,19 @@ class AuthInterceptor(grpc.ServerInterceptor):
...
         Args:
             receiver_hotkey(str):
                 the SS58 address of the hotkey which should be targeted by RPCs
-            black_list (Function, `optional`):
-                black list function that prevents certain pubkeys from sending messages
         """
         super().__init__()
         self.nonces = {}
-        self.blacklist = blacklist
+        self.blacklister = bittensor.auth_blacklister()
```
Eugene-hu: Is `bittensor.auth_blacklister` necessary inside the AuthInterceptor? If it is not needed, perhaps its better to only attach the blacklist function to keep the interceptor lightweight

bittensor/_axon/__init__.py
```diff
@@ -280,19 +274,19 @@ class AuthInterceptor(grpc.ServerInterceptor):
     def __init__(
         self,
         receiver_hotkey: str,
-        blacklist: Callable = None,
+        metagraph: "bittensor.metagraph",
```
Eugene-hu, line 282: Is there a way to turn off the blacklist feature? Or will the expectation be that it will be on by default?

bittensor/_blacklist/__init__.py
```diff
@@ -110,5 +85,61 @@ def blacklist(
...
+            nargs = '*', 
+            action = 'store',
+            help = 'List of ss58 addresses which are always allowed pass through.', default=[]
+        )
+
+    def blacklist( 
+            self, 
+            forward_call: "bittensor.SynapseCall"
+        ) -> Union[ Tuple[bool, str], bool ]:
+        print("Hi. in synapse_blacklist")
```
Eugene-hu: I think you missed some debug logging haha

bittensor/_synapse/synapse.py
```diff
@@ -134,6 +148,12 @@ def apply( self, call: SynapseCall ) -> object:
         try:
             call.log_inbound()
 
+            # Check if the request proto containes a valid hotkey by comparing with signature hotkey.
+            if call.src_hotkey != call.request_hotkey:
+                call.return_code = bittensor.proto.ReturnCode.InvalidRequestHotkey
+                call.return_message = 'Invalid request hotkey'
+                raise InvalidRequestHotkeyError
```
Eugene-hu, line 155: Amazing!

Eugene-hu (commented):
Amazing work! Left you a few comments, but looks pretty good on my end

bittensor/_axon/__init__.py
ifrit98: Yes, it is the reference to the server metagraph from the axon. I will verify this, but am also having 2nd thoughts about including this if it means we make the AuthInterceptor heavier than need be.

bittensor/_axon/__init__.py
ifrit98: Agreed. Will change.

bittensor/_axon/__init__.py
ifrit98: Set with the expectation to be always on to disallow disabling it but is the ability to disable blacklist in the a good idea to keep? Will add a flag to do this if so.

PR #1399, opened 2023-06-21
fix bug in unstake command to manually enter wallet (bypassed)

ifrit98:
Prompts correctly now for wallet and hotkey name upon calling.  Was just going to default without user input before.


shibshib (approved):

ifrit98:
This is already fixed in `new_staging`

PR #1410, opened 2023-06-23
Staging fast api

ifrit98:
Bring staging up-to-date with `new_fast_api` where we have been adding multimodality miners to be queried using FastAPI

- Adds fast API to the axon server
- Adds various multimodality miners (text-to-speech, text-to-image, etc)


ifrit98:
This PR must be modified to remove neurons as they will be in `openminers` and `openvalidators`

PR #1423, opened 2023-06-28
dont sync metagraph on neuron init

camfairchild:
This fixes errors that require the subtensor config to be passed on the CLI.

### DIFF
--- a/bittensor/_neuron/base_miner_neuron.py
+++ b/bittensor/_neuron/base_miner_neuron.py
@@ -106,7 +106,9 @@ class BaseMinerNeuron:
         bittensor.logging( config = self.config, logging_dir = self.config.neuron.full_path )
         self.subtensor = bittensor.subtensor( self.config )
         self.wallet = bittensor.wallet( self.config )
-        self.metagraph = self.subtensor.metagraph( self.config.netuid )
+        self.metagraph = self.subtensor.metagraph( self.config.netuid, sync=False )
+        self.metagraph.sync( lite = True, subtensor=self.subtensor )
+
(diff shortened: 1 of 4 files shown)


ifrit98 (approved):
Great catch!

Will port this over to `openminers` also.

p-ferreira (approved):
LGTM although I'm not sure if I understood the implication on the validator.
So we want to sync the metagraph with `self.subtensor`?

The current implementation on openvalidators syncs like
` self.metagraph.sync()`. I see in the metagraph code that `lite` is set to true by default, but the `subtensor` is initially empty and later filled with `bt.subtensor`. 

Should we replicate this approach to the `openvalidators` as well?

ifrit98:
> Should we replicate this approach to the `openvalidators` as well?

I believe we should, as (soon-ish) `_neuron` will be removed from `bittensor` proper with the next release following `5.2.1`.

ifrit98:
See https://github.com/opentensor/bittensor/pull/1426

camfairchild:
> LGTM although I'm not sure if I understood the implication on the validator.
> So we want to sync the metagraph with `self.subtensor`?
> 
> The current implementation on openvalidators syncs like
> ` self.metagraph.sync()`. I see in the metagraph code that `lite` is set to true by default, but the `subtensor` is initially empty and later filled with `bt.subtensor`. 

The config for the subtensor, including any chain endpoint, only gets passed to `self.subtensor`

We also need that when syncing the metagraph, so we have to pass that in somehow, meaning we should sync using it and disable the auto-sync when it initializes the object (because it has no config param)

PR #1431, opened 2023-06-30
text_to_image neuron changes

CreativeBuilds:
- adds the ability to load files locally instead off from huggingface
- enables safety checker by default
- fixes missing requirements in requirements/prod.txt


ifrit98 (approved):
LGTM! Like the optionality.

PR #1432, opened 2023-06-30
Remove unnecessary validator code

ifrit98:
This PR removes validator related code as it now lives in `openvalidators` repo.


bittensor/_cli/commands/misc.py
```diff
@@ -28,13 +28,6 @@ class HelpCommand:
     @staticmethod
     def run (cli):
         sys.argv = [sys.argv[0], '--help']
-        # # Run miner.
-        # if cli.config.model == 'core_server':
```
camfairchild, line 32: Should remove btcli run then?

camfairchild (commented):
See comment

bittensor/_cli/commands/misc.py
ifrit98: Might as well gut the whole HelpCommand class, ya?

camfairchild: > Might as well gut the whole HelpCommand class, ya?

Yes for sure

camfairchild (approved):

PR #1433, opened 2023-06-30
Always sync metagraph during subtensor.metagraph

camfairchild:
Should sync on call. This fixes a flaky test as well


ifrit98 (approved):
Beautiful.

PR #1435, opened 2023-07-01
catch error on loading loras

CreativeBuilds:
Miners will error out completely if someone attempts to load a lora that doesnt exist

this fixes it and instead attempts to load the lora, and if it doesnt exist, skips it

### DIFF
--- a/neurons/text_to_image/miners/neuron.py
+++ b/neurons/text_to_image/miners/neuron.py
@@ -208,8 +208,11 @@ def main( config ):
 
                 # load lora weights
                 for model, weight in lora_models:
-                    i2i = load_lora_weights(i2i, model, weight)
-
+                    try:
+                        i2i = load_lora_weights(i2i, model, weight)
+                    except Exception as e:
+                        None
+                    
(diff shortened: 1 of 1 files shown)


ifrit98 (changes requested):
Great, thanks!

Small nit: Would be nice to have a simple log message around this to let users know what happened.
e.g.
```python
                    try:
                        i2i = load_lora_weights(i2i, model, weight)
                    except Exception as e:
                        bittensor.logging.info("Error loading LoRA. Skipping...")
```

ifrit98:
Small nit: Can we add some error handling around this to explain?



> Great, thanks!
> 
> Small nit: Would be nice to have a simple log message around this to let users know what happened. e.g.
> 
> ```python
>                     try:
>                         i2i = load_lora_weights(i2i, model, weight)
>                     except Exception as e:
>                         bittensor.logging.info("Error loading LoRA. Skipping...")
> ```

Nvm, realized this was a hotfix.  Will add logging later.

ifrit98 (approved):

PR #1436, opened 2023-07-03
add weights call to subtensor

camfairchild:


bittensor/_metagraph/__init__.py
```diff
@@ -114,7 +114,15 @@ def sync ( self, block: Optional[int] = None, lite: bool = True, subtensor: Opti
         if lite:
             self.neurons = subtensor.neurons_lite( block = block, netuid = self.netuid )
         else:
-            self.neurons = subtensor.neurons(block = block, netuid = self.netuid )
```
shibshib, line 117: What's the difference between using subtensor.neurons alone and using neurons_lite with all the other calls for weights and etc? Is it just deconstructed or does this deprecate using subtensor.neurons completely?

shibshib (approved):

bittensor/_metagraph/__init__.py
camfairchild: It would deprecate `subtensor.neurons`
It ends up being faster, because of something in the RPC on the node-side being slow for that call

PR #1449, opened 2023-07-12
Use SenateMembers in is_senate_member helper

Rubberbandits:
This PR is required for the new change in subtensor which removes the Senate collective instance, simplifying the runtime for ease-of-use. You can find the matching PR for subtensor [here.](https://github.com/opentensor/subtensor/pull/166)


ifrit98 (approved):

Rubberbandits:
We don't need the subtensor update to use this change in the CLI, merging.

PR #1452, opened 2023-07-14
Fix blacklist vpermit_required

Inquinim:
Specifying --neuron.blacklist.vpermit_required would **REJECT** validators that have the vpermit, instead of rejecting those without.
The return statement was also not providing a tuple (bool, str) either like the others.

@shibshib @ifrit98


ifrit98:
Thanks for pointing this out!

This is addressed in `openminers` now also:  https://github.com/opentensor/miners/commit/19a5c998a9105797738101e1bac3bcc1af37ff3f

PR #1460, opened 2023-07-24
fix torch typehint on some neurons BT-1329

### LINKED ISSUE #1329: Import torch statement missing from Cerebras' neuron.py
**Describe the bug**
And import torch statement is missing from neuron.py in the cerebras directory causing the script to error when run. Adding import torch allowed the script to run properly

**To Reproduce**
pm2 start ~/.bittensor/bittensor/neurons/text/prompting/miners/cerebras/neuron.py

Error: No Module named torch

camfairchild:
Fixes #1329


ifrit98 (approved):
EZ money.

PR #1464, opened 2023-07-25
Faster overview pull

camfairchild:
Suggested by @adriansmares 
Makes use of `concurrent.futures.ProcessPoolExecutor` to multi-process the pulling of each netuid's neurons.
This optimizes the time due to CPU-bound decoding/marshalling of the received `NeuronInfo` objects.

Seems to average around 9-10s faster (12s vs 21-22s)


ifrit98 (approved):
This is a very welcome upgrade!

PR #1476, opened 2023-08-08
Revolution whitelist

ifrit98:
Whitelist based on metagraph validator IP.  Coordinates with https://github.com/opentensor/validators/commit/a7e6da7290274a77ae89c533a1a18ffdf561bd51.

This ensures that only requests from validators with a vpermit are allowed to query using an additional middleware.


ifrit98:
Would like some feedback on this.  

Simplest way seems to be to add the metagraph object to the axon, but unsure if this is optimal.

PR #1480, opened 2023-08-10
add cicleci check for requirements python version support

ifrit98:
Does 2 thins:
- Adds a bash script to verify that all python requirements are supported with the given our officially supported python versions. (e.g. 38, 39, 310, 311)
- Adds a circle CI job so that if any of the tests fail from the bash script, the job fails.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,18 @@ jobs:
             . env/bin/activate
             python -m black --exclude '(env|venv|.eggs)' --check .
 
+  check-compatibility:
+    docker:
+      - image: cimg/python:3.10
+    steps:
+      - checkout
+      - run:
+          name: Install jq
+          command: sudo apt-get update && sudo apt-get install -y jq
+      - run:
+          name: Check Python Version Compatibility
+          command: ./check_compatibility.sh
+
   build-and-test:
     resource_class: medium
     parallelism: 2
@@ -205,6 +217,7 @@ jobs:
 workflows:
   pr-requirements:
     jobs:
+      - check-compatibility
       - black:
           python-version: "3.8.12"
       - build-and-test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,9 @@ jobs:
           name: Install jq
           command: sudo apt-get update && sudo apt-get install -y jq
       - run:
+          name: Make script executable
+          command: chmod +x ./scripts/check_compatibility.sh
+      - run:
           name: Check Python Version Compatibility
           command: ./scripts/check_compatibility.sh
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
           command: sudo apt-get update && sudo apt-get install -y jq
       - run:
           name: Check Python Version Compatibility
-          command: ./check_compatibility.sh
+          command: ./scripts/check_compatibility.sh
 
   build-and-test:
     resource_class: medium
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -42,6 +42,21 @@ jobs:
...
+    docker:
+      - image: cimg/python:3.10
+    steps:
+      - checkout
+      - run:
+          name: Install jq
+          command: sudo apt-get update && sudo apt-get install -y jq
+      - run:
+          name: Make script executable
+          command: chmod +x ./scripts/check_compatibility.sh
```
eduardogr, line 55: i think it is enough adding the permissions in the repository itself, can you try it?

If that is the case we can avoid this step

scripts/check_compatibility.sh
```diff
@@ -0,0 +1,65 @@
...
+        classifiers=$(echo "$response" | jq -r '.info.classifiers[]')
+        requires_python=$(echo "$response" | jq -r '.info.requires_python')
+
+        base_version="Programming Language :: Python :: ${python_version%%.*}"
+        specific_version="Programming Language :: Python :: $python_version"
+
+        if echo "$classifiers" | grep -q "$specific_version" || echo "$classifiers" | grep -q "$base_version"; then
+            echo -e "${GREEN}Supported${NC}"
+        elif [ "$requires_python" != "null" ] && echo "$requires_python" | grep -Eq "==$python_version|>=$python_version|<=$python_version"; then
+            echo -e "${GREEN}Supported${NC}"
```
eduardogr, line 37: do you think that knowing the reason of the failure could enrich the logs?? would it be valuable??

i mean, maybe 3.10 is not compatible bcz, we find a strict < 3.10, for example

scripts/check_compatibility.sh
```diff
@@ -0,0 +1,65 @@
...
+        echo -n "Checking $package_name... "
+
+        url="https://pypi.org/pypi/$package_name/json"
+        response=$(curl -s $url)
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" $url)
+
+        if [ "$status_code" != "200" ]; then
+            echo -e "${RED}Information not available${NC}"
+            continue
+        fi
```
eduardogr, line 26: should this be treated like a non-compatible scenario?
if we cannot ensure this automatically we have to do something at least (like allow listing package versions compatibility even if its harcoded). But maybe we do not want to have green builds and red logs, right?

Becuase if all is OK, but one package is not found, we will have a green build with this error-red message. But we will not see it most probably.

What do you think about that?

scripts/check_compatibility.sh
```diff
@@ -0,0 +1,65 @@
...
+    fi
+done
+
+echo ""
+if $all_passed; then
+    echo -e "${GREEN}All tests passed.${NC}"
+else
+    echo -e "${RED}All tests did not pass.${NC}"
+    exit 1
+fi
```
eduardogr, line 65: very sweet,

Just pointing out, in case you want to address it:
- If you can extract python versions as the input of the script...
  - you can have one run of the script per version. e.g.:
    - ./check_compatibility.sh 3.8
    - /check_compatibility.sh 3.9
    - /check_compatibility.sh 3.10
  - and we could run every version check in parallel. Taking advantage of circleci (same approach we have for test runs)

ifrit98:
Migrating to: https://github.com/opentensor/bittensor/pull/1483

.circleci/config.yml
ifrit98: Thanks Edu! Ended up resolving these in: https://github.com/opentensor/bittensor/pull/1483

PR #1482, opened 2023-08-15
remove miniupnpc

ifrit98:
Removes an unused part of the bittensor codebase.

Default is not to use miniupnpc and often fails when set to True.

The main reason for removing is to avoid having to install bittensor with sudo as miniupnpc requires sudo to install.


camfairchild (approved):
good riddance!

PR #1490, opened 2023-08-23
Call runtime apis using state call

camfairchild:
This switches any custom RPC calls to instead use the `state_call` rpc to directly call the runtime API.   
This *should* allow us to change the runtime API signatures in the future, without modifying any RPCs.  
Meaning we would simply do a python-client code update for the metadata (or pull it live 👀 )


ifrit98 (approved):

Rubberbandits (approved):
LGTM 👍

PR #1491, opened 2023-08-23
[Revolution] call runtime apis using state call

camfairchild:
This is the companion PR for revolution for #1490

### DIFF
--- a/bittensor/chain_data.py
+++ b/bittensor/chain_data.py
@@ -21,7 +21,7 @@ import bittensor
 from enum import Enum
 from dataclasses import dataclass
 from scalecodec.types import GenericCall
-from typing import List, Tuple, Dict, Optional, Any, TypedDict
+from typing import List, Tuple, Dict, Optional, Any, TypedDict, overload, Union
 from scalecodec.base import RuntimeConfiguration, ScaleBytes
 from scalecodec.type_registry import load_type_registry_preset
 from scalecodec.utils.ss58 import ss58_encode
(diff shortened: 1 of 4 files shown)


bittensor/subtensor.py
```diff
@@ -1858,26 +1943,23 @@ def neuron_for_uid_lite(
...
-                )
-
-        json_body = make_substrate_call_with_retry()
-        result = json_body["result"]
-
-        if result in (None, []):
+        if hex_bytes_result == None:
             return NeuronInfoLite._null_neuron()
+        
+        bytes_result = bytes.fromhex(hex_bytes_result[2:])
```
ifrit98: Is this just removing the byte header? I'm always weary of hard-coded list indexing, but if it's a constant than I'm fine with it.  Would it be helpful to have a try block here?

bittensor/chain_data.py
```diff
@@ -225,15 +225,52 @@ class ChainDataType(Enum):
 U16_MAX = 65535
 U64_MAX = 18446744073709551615
 
+@overload
```
ifrit98: What's the role of `@overload` here? Is this substantially different from `@abstractmethod`?

ifrit98 (approved):
LGTM! 

Only a tiny nit about the list indexing.

Rubberbandits (approved):
LGTM :+1:

bittensor/chain_data.py
camfairchild: It allows you to add type hints for multiple signatures of the same function. Meaning for multiple types.
Now that I think about it, this could be a `Union`... instead

ifrit98 (approved):
LGTM!

PR #1495, opened 2023-08-24
Add dereg stake to overview command

camfairchild:
Adds hotkeys to overview if they have no registration, but still exist in the `Stake` map for any of the wallet(s) coldkey(s).   
This should make it easier to find keys which you have stake, but have not automatically unstaked.    

Relies on merge of https://github.com/opentensor/subtensor/pull/179 and subsequent chain upgrade


bittensor/commands/overview.py
```diff
@@ -95,36 +96,144 @@ def run(cli):
...
+                    if len(neurons_result) == 0:
+                        # Remove netuid from overview if no neurons are found.
+                        netuids.remove(netuid)
+                        del neurons[str(netuid)]
+                    else:
+                        # Add neurons to overview.
+                        neurons[str(netuid)] = neurons_result
+
+            total_coldkey_stake_from_metagraph = {
+                coldkey_wallet.coldkeypub.ss58_address: bittensor.Balance(0.0)
```
ifrit98: Only 1 nit with a failing test:

```python
def test_overview_all(self, _):
    config = self.config
    config.command = "overview"
    config.no_prompt = True
    config.netuid = []  # Don't set, so it tries all networks.

    config.all = True
    cli = bittensor.cli(config)
    cli.run()
```

This test complains with `coldkey_wallet` referenced before assignment in the case where `config.all` is passed.

Seems like we hit the first if cond early and iterate over:

```python
cold_wallets  = get_coldkey_wallets_for_path(cli.config.wallet.path)
```

When I try to naively change `cold_wallet` -> `coldkey_wallet` to get the last coldkey_wallet in the list of `cold_wallets` in the first if branch then I get a keyfile not found error :(

```bash
E  bittensor.errors.KeyFileError: Keyfile at: ~/.bittensor/wallets/regen_hotkey_testwallet/coldkeypub.txt does not exist bittensor/keyfile.py:474: KeyFileError
```

ifrit98 (approved):

PR #1251, opened 2023-04-05
Bit 87 coldkey encryption security

isabella618033:
#2 Update wallet from Ansible Vault to NaCl & argon2i password hashing.

Wallets can be updated with the following scenario 
- `btcli update_wallet --all`
- After running a transfer/stake/unstake/delegate command, keyfile version would be checked and advised to get updated.

(successful update with legacy_wallet_2478, failure in update with legacy_wallet_7807, all other wallets had been updated previously)
image

PR #1497, opened 2023-08-25
BIT-87 to revolution wallet upgrade

ifrit98:
Ports https://github.com/opentensor/bittensor/pull/1251 to `revolution`.

```
Update wallet from Ansible Vault to NaCl & argon2i password hashing.

Wallets can be updated with the following scenario

btcli update_wallet --all
After running a transfer/stake/unstake/delegate command, keyfile version would be checked and advised to get updated.
(successful update with legacy_wallet_2478, failure in update with legacy_wallet_7807, all other wallets had been updated previously)

```
Example flow
image


bittensor/keyfile.py
```diff
@@ -409,6 +479,72 @@ def _may_overwrite(self) -> bool:
...
+            return False
+
+        update_keyfile = False
+        if not no_prompt:
+            keyfile_data = self._read_keyfile_data_from_file()
+
+            # If the key is not nacl encrypted.
+            if keyfile_data_is_encrypted(keyfile_data) and not keyfile_data_is_encrypted_nacl(keyfile_data):
+                bittensor.__console__.print(f"You may update the keyfile to improve the security for storing your keys.\nWhile the key and the password stays the same, it would require providing your password once.\n:key:{self}\n")
+                bittensor.__console__.print(f"Please make sure you have the mnemonic stored.", style="white on red")
```
Eugene-hu: maybe `Please make sure you have the mnemonic stored in case an error occurs during the transfer`

Eugene-hu: Do we want to prompt the user if they have a mnemonic stored? Or is that too much trouble you think?

tests/unit_tests/test_wallet.py
```diff
@@ -15,13 +15,200 @@
...
+    
+    def create_wallet(self):      
+        # create an nacl wallet
+        wallet = bittensor.wallet(name=f"mock-{str(time.time())}")  
+        with patch.object(
+            bittensor,
+            "ask_password_to_encrypt",
+            return_value=self.default_updated_password,
+        ):
+            wallet.create()
```
Eugene-hu, line 52: By default, all wallets created from now on will be NaCl encrypted right?

bittensor/cli.py
```diff
@@ -92,6 +92,7 @@
             "regen_coldkeypub": RegenColdkeypubCommand,
             "regen_hotkey": RegenHotkeyCommand,
             "faucet": RunFaucetCommand,
+            "update_wallet": UpdateWalletCommand,
```
Eugene-hu: Do we need to NaCL to the requirements?

Eugene-hu: One followup, what are the python requirements for NaCl? If ansible vault is being depreciated, we might be able to bring back python 3.8.

Eugene-hu (approved):
LGTM, left you a few questions and comments

tests/unit_tests/test_wallet.py
```diff
@@ -15,13 +15,200 @@
...
 
+import json
 import time
 import pytest
+import random
+import getpass
 import unittest
 import bittensor
+from rich.prompt import Confirm
+from ansible_vault import Vault
```
ifrit98, line 26: Is this (`ansible_vault`) needed still?

bittensor/cli.py
ifrit98: > One followup, what are the python requirements for NaCl? If ansible vault is being depreciated, we might be able to bring back python 3.8.

Might be worth brining back. We still have ~1yr of security support for 3.8: https://endoflife.date/python

ifrit98: > Do we need to NaCL to the requirements?

image

`pynacl==1.5.0` is in the apple-m1 req file, just needs to be ported to `requirements/prod.txt`.

bittensor/keyfile.py
isabella618033: oh okay, that's a good idea

tests/unit_tests/test_wallet.py
isabella618033: yes yes

bittensor/cli.py
isabella618033: cannot deprecate ansible vault, cause we still need it to decrpyt some old wallet to update the wallet, 
other then that, we may need to keep supporting the old wallets

ifrit98: Ok, sounds good. Then we likely can't bring back py3.8.

ifrit98 (commented):
LGTM!

tests/unit_tests/test_wallet.py
isabella618033: yes it is needed, cause we need it to decrypt and update the wallet.
also, we should still support decrypting the old user wallets.

PR #1503, opened 2023-08-28
Add body hash to signature for verification, stop overrides

ifrit98:
This does a few things:
* Adds a body_hash property to `Synapse` which computes the hash of the required non-null entities of the synapse, which would normally be considered the "body".
* Makes `body_hash` a read-only property so attempting to set the value will result in a TypeError.
* Adds metaclass override protection on `Synapse` such that `body_hash` is impossible to override.

This enforces that all synapse required elements cannot be modified in transit.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -24,7 +24,6 @@ import uuid
 import copy
 import time
 import asyncio
-import hashlib
 import inspect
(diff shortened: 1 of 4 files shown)


camfairchild (approved):
LGTM

PR #1504, opened 2023-08-28
remove nest asyncio

ifrit98:
Still can run `await coroutine` in jupyter, however if trying to run `loop.run_until_complete(coroutine)` must instead use `asyncio.create_task(coroutine)`.


bittensor/axon.py
```diff
@@ -39,6 +40,9 @@
 from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from typing import Dict, Optional, Tuple, Union, List, Callable
 
+# Force asyncio to use the uvloop created event loop. (Much faster)
+asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```
camfairchild: Should only need this once in the program. Try in the main() ?

camfairchild (approved):
NIT. See comment

bittensor/axon.py
ifrit98: Fixed. ~~Removed from dendrite.py~~ Moved to `__init__`s

PR #1511, opened 2023-09-03
let subnets define schema for required body hash

ifrit98:
Rather than attempting to parse blindly.

### DIFF
--- a/bittensor/synapse.py
+++ b/bittensor/synapse.py
@@ -26,7 +26,6 @@ import hashlib
 import pydantic
 from pydantic.schema import schema
 import bittensor
-from abc import abstractproperty
 from typing import Optional, List, Any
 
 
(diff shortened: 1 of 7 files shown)


bittensor/synapse.py
```diff
@@ -335,8 +336,26 @@ def __setattr__(self, name, value):
...
+        Get the list of required fields for the Synapse instance.
+
+        This method retrieves the list of required fields for the Synapse instance
+        by looking up the schema definition for the current class and returning the
+        list of required fields.
+
+        Returns:
+            List[str]: A list of required fields for the Synapse instance.
+        """
+        return []
```
Eugene-hu: As a default, perhaps it should just return all the hash fields. The property gives subnet designers flexibility if they want to add additional fields in the future while providing backward compatibility.

Eugene-hu (approved):
LGTM, left you a suggestion

PR #1515, opened 2023-09-12
Weight normalization

Eugene-hu:
- root weights weren't normalized before emission, this leads to floats being passed into the chain. 
- adds an emission normalization before submission


ifrit98 (approved):
Ah, makes sense why this would cause issues.

LGTM!

PR #1516, opened 2023-09-14
Header fix

ifrit98:
TL;DR

Headers max out at 8kB, which means we can't send messages >1200 tokens. (Whoops!)

To get around this, don't pass the full body into the headers and instead hash them so that they can be easily reconstructed on the axon side. Added dependency for routes so that we inspect the body, hash it, and compare against the hashes found in the request headers for validation. This was caused by an improper separation of concerns. Now the dendrite handles hashing the body, and the axon reconstructs the signature from the headers, and validates the body in a dependency before passing to the routing function.

Sub-points and other changes:
* Properly separate concerns so that dendrite handles hashing the synapse body for verification.
* Synapse no longer needs override metaclass protection for body_hash property (removed)
* AxonMiddleware::verify now accepts the request body as an argument to validate the body hashes that were constructed by the dendrite and signed with the server key.
* Synapse.to_headers() now creates an empty (dummy) value for the body elements to pass pydantic validation and cannot be removed completely.


bittensor/axon.py
```diff
@@ -648,15 +715,22 @@ async def dispatch(
...
             )
 
-            # Call the verify function
-            await self.verify(synapse)
-
             # Call the blacklist function
             await self.blacklist(synapse)
 
             # Call the priority function
             await self.priority(synapse)
```
Eugene-hu: If `priority` or `blacklist`, this will break the connection, returning an error message correct?

bittensor/axon.py
```diff
@@ -648,15 +715,22 @@ async def dispatch(
...
 
+            # Remove the body for inspection
+            request_body = await request.body()
+
+            # Call verify and return the verified request
+            await self.verify(synapse, request_body)
```
Eugene-hu: I see so the peaking into the body occurs here in the signature verification?

bittensor/axon.py
```diff
@@ -648,15 +715,22 @@ async def dispatch(
...
+
+            # Repackage the request so we can pass it along
+            request = RequestWithBody(request.scope, request_body)
+            del request_body
```
Eugene-hu: Any reason why we explicitly call to del the request body? This should only delete the pointer of the request body and shouldn't delete anything in the garbage collection, right?

bittensor/axon.py
```diff
@@ -604,6 +622,55 @@ def default_verify(self, synapse: bittensor.Synapse) -> Request:
         self.nonces[endpoint_key] = synapse.dendrite.nonce
 
 
+class RequestWithBody(Request):
+    """
+    A specialized HTTP request wrapper that includes a pre-defined request body.
```
Eugene-hu: Really cool wrapper, so we cant access the same body once the information has been read by the middle-ware asynchronously and must recreate a second request?

Eugene-hu (approved):
LGTM, gonna approve for now, but lets wait for @isabella618033's analysis before we merge it

bittensor/axon.py
ifrit98: Yes exactly. It will show up in the JSONResponse that's set in the synapse headers.

bittensor/axon.py
ifrit98: Yep, unfortunately I don't think there's a clean way to break this up.

bittensor/axon.py
ifrit98: More of a signal to the GC to get around to it when it feels like it :D. Probably can remove it

bittensor/axon.py
ifrit98: Yes unfortunately it was impossible to rewrap the same body. The request lifecycle for starlette doesn't allow for this.

Eugene-hu: I see, are there any increases in memory usage due to the repacking of the request?

bittensor/axon.py
```diff
@@ -318,7 +320,10 @@ def attach(
 
         # Add the endpoint to the router, making it available on both GET and POST methods
         self.router.add_api_route(
-            f"/{request_name}", forward_fn, methods=["GET", "POST"]
+            f"/{request_name}",
+            forward_fn,
+            methods=["GET", "POST"],
+            dependencies=[Depends(self.verify_body_integrity)],
```
Eugene-hu, line 326: Really clean design! The api route now depends on the `verify_body_integrity`?

ifrit98: Yessir! Seems to be pretty fast in my initial testing as well.  We await there but don't need to repackage, just simply passes the body dict on to the function.

PR #1520, opened 2023-09-24
updates axon hash to split on _hash_ instead of _

CreativeBuilds:
allows for variables with _'s in their naming


ifrit98:
Addressed here: https://github.com/opentensor/bittensor/pull/1519

PR #1521, opened 2023-09-26
simplify hashing for request verification

ifrit98:
Simplifies the verification process by having a single hash to send over the wire instead of hashes for each field. Also cleans up the naming convention issues with splitting on `_hash_`.

2-step verification process:
* Dendrite calls `synapse.to_headers()` on synapse instance to get required information for the request, which now includes `computed_body_hash` as a hash of the concatenated hashes for each required body field, as well as `hash_fields` which passes the list of required fields that must be hashed in order to verify the request.
* Axon middleware receives request and does intercept verification (blacklist pre-verify) using only header information, which ensures the hashes passed in the request headers were indeed signed by the calling keypair.
* `verify_body_integrity` is called via dependency injection in fastapi to unwrap the body at an appropriate place in its lifecycle. This verifies the body itself was not modified in transit.
* `forward_fn` is processed with the verified body and returned after `run()` completes, returning the response from the server.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -24,6 +24,7 @@ import uuid
 import copy
 import json
 import time
+import base64
 import asyncio
 import inspect
 import uvicorn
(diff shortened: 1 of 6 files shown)


Eugene-hu (approved):
LGTM

PR #1524, opened 2023-09-28
Root Metagraph fixes

Eugene-hu:
`meta = sub.metagraph(0,lite=False)` breaks the current implementation of the metagraph. 
image

This PR adds a new flag `root` that defines the metagraph's behaviour when a root network is expected.
The new correct method: `meta = sub.metagraph(0,lite=False, root=True)`


ifrit98 (approved):
LGTM! 

Would we need to update the cli also to give people the option to check this at the command line?  (Or perhaps make it automatic to set root=True when calling btcli metagraph with netuid 0?)

Rubberbandits (approved):
LGTM

PR #1529, opened 2023-09-30
Metagraph 0 fix

ifrit98:
* metagraph cannot have `self.subtensor` as when validators try to `deepcopy` we hit a pickling error
* replace `self.subtensor` in metagraph by passing where appropriate from single source.


Eugene-hu (approved):
LGTM, nice catch

PR #1532, opened 2023-10-01
Header fix

ifrit98:
* Move from md5 (broken) -> sha3 hashing
* Eliminate parsing `required_hash_fields` and let subclasses define them directly as a pydantic field
* `required_hash_fields` is parsed in the `attach()` method directly from the `synapse` protocol subclass
* Makes `default_verify` async (as it should be)

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -312,7 +312,9 @@ class axon:
         ].annotation
 
         # Parse required hash fields from the forward function protocol defaults
-        required_hash_fields = request_class.__dict__["__fields__"]["required_hash_fields"].default
+        required_hash_fields = request_class.__dict__["__fields__"][
+            "required_hash_fields"
+        ].default
 
         # Assert that the first argument of 'forward_fn' is a subclass of 'bittensor.Synapse'
         assert issubclass(
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -270,6 +270,7 @@ class axon:
         blacklist_fn: Callable = None,
         priority_fn: Callable = None,
         verify_fn: Callable = None,
+        required_hash_fields: List[str] = None,
     ) -> "bittensor.axon":
         """
         Registers an API endpoint to the FastAPI application router.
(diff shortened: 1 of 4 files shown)


Eugene-hu (approved):
LGTM

bittensor/axon.py
```diff
@@ -285,6 +286,7 @@ def attach(
                                               Defaults to None, meaning no priority sorting will be applied.
             verify_fn (Callable, optional): Function to verify requests. It should take the same arguments as 'forward_fn' and return
                                             a boolean value. If None, 'self.default_verify' function will be used.
+            required_hash_fields (List[str], optional): List of fields in the request body that should be hashed and compared against
```
isabella618033: parameter that is not needed?

bittensor/synapse.py
```diff
@@ -542,33 +493,30 @@ def to_headers(self) -> dict:
         headers["header_size"] = str(sys.getsizeof(headers))
         headers["total_size"] = str(self.get_total_size())
         headers["computed_body_hash"] = self.body_hash
-        headers["hash_fields"] = base64.b64encode(
```
isabella618033, line 545: does the axon need to know the other parsed hash_fields other then the required_hash_field?

bittensor/synapse.py
```diff
@@ -689,11 +637,6 @@ def parse_headers_to_inputs(cls, headers: dict) -> dict:
         inputs_dict["header_size"] = headers.get("header_size", None)
         inputs_dict["total_size"] = headers.get("total_size", None)
         inputs_dict["computed_body_hash"] = headers.get("computed_body_hash", None)
-        inputs_dict["hash_fields"] = json.loads(
```
isabella618033, line 692: (same question as the previous one) does the axon need to know the other parsed hash_fields other then the required_hash_field?

isabella618033 (approved):
LGTM, tested axon and synapse on my end.

bittensor/axon.py
ifrit98: Yes, removing it

bittensor/synapse.py
ifrit98: Nope, that is a hold over, nice catch! Will remove.

PR #1533, opened 2023-10-02
fixes for root weights on metagraph

Eugene-hu:
- adds an additional function for mapping the subnets to root weights functions


ifrit98 (approved):
LGTM!

PR #1541, opened 2023-10-08
Cli fix

ifrit98:
* silly error that when a coldkey is overwritten locally but the chain has a different ss58 addr, the cli breaks.


Eugene-hu (approved):
LGTM

PR #1542, opened 2023-10-09
update faucet helpstr

ifrit98:


bittensor/commands/register.py
```diff
@@ -264,7 +264,7 @@ def run(cli):
     @staticmethod
     def add_args(parser: argparse.ArgumentParser):
         run_faucet_parser = parser.add_parser(
-            "faucet", help="""Register a wallet to a network."""
+            "faucet", help="""Perform PoW to receieve test TAO in your wallet."""
```
steffencruz, line 267: Why does this command perform PoW ? Sorry for my ignorance here

ifrit98: It's simply to run the faucet with a very low difficulty so takes ~15-30 sec.  So people don't spam and get instant tokens.

steffencruz: TIL :)

steffencruz (approved):

PR #1543, opened 2023-10-10
Fix typo

steffencruz:
Important to clarify that subnet creation requires that funds are **locked** and not **burned**


ifrit98 (approved):
Nice catch!

PR #1545, opened 2023-10-14
fixes tensors not hashing correctly

CreativeBuilds:
This issue was caused from using `__dict__` which converted the Tensor object into one with just its datatype and shape values, disregarding the actual buffer or content within the Tensor.

The Axon.py file when it was trying to decode the response from the sender would then compare the hash using the buffer that was sent, causing a mismatch of hashes.

This change fixes that by utilizing `.dict()` instead.

### DIFF
--- a/bittensor/synapse.py
+++ b/bittensor/synapse.py
@@ -513,8 +513,7 @@ class Synapse(pydantic.BaseModel):
         hashes = []
 
         # Getting the fields of the instance
-        instance_fields = self.__dict__
-
+        instance_fields = self.dict() # __dict__ converts the Tensor types into strings causing a mismatch in the hash, so we use .dict() instead
         for field, value in instance_fields.items():
             # If the field is required in the subclass schema, hash and add it.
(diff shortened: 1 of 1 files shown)


bittensor/synapse.py
```diff
@@ -513,8 +513,7 @@ def body_hash(self) -> str:
         hashes = []
 
         # Getting the fields of the instance
-        instance_fields = self.__dict__
-
+        instance_fields = self.dict() # __dict__ converts the Tensor types into strings causing a mismatch in the hash, so we use .dict() instead
```
Eugene-hu, line 516: I think @ifrit98 would have better visibility on this change than I do, so we should for his approval before merging.

Eugene-hu: @ifrit98 would this create an issue for the header size? If I am not mistaken, there is a sizing limit on the headers which caused us to create hashes rather than using the full tensor.

ifrit98:
See: https://github.com/opentensor/bittensor/pull/1548

PR #1547, opened 2023-10-16
Added mechanism to sum all delegated tao

shibshib:
Basic change, adds a line when displaying my_delegates for `Total delegated Tao` which shows the full amount of total delegated Tao across all wallets.


Eugene-hu (approved):
LGTM

PR #1551, opened 2023-10-18
Streaming fix

ifrit98:
* Separate out streaming from non-streaming requests (`AsyncGenerator` return value)
* Yield chunks as they are sent


bittensor/dendrite.py
```diff
@@ -133,66 +133,94 @@ async def forward(
...
-                    )
-                # The function then returns a list of responses from all axons.
-                return all_responses
-            else:
-                # Here we build a list of coroutines without awaiting them.
-                coroutines = [
-                    self.call(
+        # Check if synapse is an instance of the StreamingSynapse class or if streaming flag is set.
+        streaming = (
+            issubclass(synapse.__class__, bittensor.StreamingSynapse) or streaming
```
Eugene-hu: Should there be a warning if the two don't match? Should we expect the base class to be streaming if streaming is used?

Eugene-hu (approved):
LGTM, left you a comment

bittensor/dendrite.py
ifrit98: Yes, we should expect to subclass, but if subnet owners decide that the `bt.StreamingSynapse` is too restrictive for their purposes, they should be able to write their own class that inherits from `bt.Synapse`.

ifrit98: Agree we should add a warning here.

PR #1554, opened 2023-10-20
Remove tensor headers

ifrit98:
* This ensures that headers don't exceed 8kb limit for http protocol.
* Removes all tensor related header information in `to_headers()` and `from_headers()`


ifrit98:
See https://github.com/opentensor/bittensor/pull/1555. Included there.

ifrit98:
> Does this mean that tensors are no longer being hashed/secured over transit?

No, any field set in `required_hash_fields` in the subnet protocol will still be verified, regardless of object type. This is specifically referring to metadata about tensor objects, and doesn't have to do with their verification.

PR #1555, opened 2023-10-21
Dendrite & Synapse updates and fixes

ifrit98:
* Fixes dendrite `UnclosedSessionError` by introducing context manager via `__aexit__` and `__del__`
* Updates dendrite documentation
* Fixes inconsistent use of `self.__dict__` vs `self.dict()` in synapse methods. `.dict()` preferred.  See https://github.com/opentensor/bittensor/pull/1548.
* Includes https://github.com/opentensor/bittensor/pull/1554
* Adds tests for dendrite proper close session and `synapse.dict()` call

### DIFF
--- a/bittensor/dendrite.py
+++ b/bittensor/dendrite.py
@@ -42,13 +42,49 @@ class dendrite(torch.nn.Module):
 
     Attributes:
         keypair: The wallet or keypair used for signing messages.
+        external_ip (str): The external IP address of the local system.
+        synapse_history (list): A list of Synapse objects representing the historical responses.
 
     Methods:
         __str__(): Returns a string representation of the Dendrite object.
         __repr__(): Returns a string representation of the Dendrite object, acting as a fallback
                     for __str__().
+        query(self, *args, **kwargs) -> Union[bittensor.Synapse, List[bittensor.Synapse]]:
+            Makes synchronous requests to one or multiple target Axons and returns responses.
 
-    Example:
+        forward(self, axons, synapse=bittensor.Synapse(), timeout=12, deserialize=True, run_async=True, streaming=False) -> bittensor.Synapse:
+            Asynchronously sends requests to one or multiple Axons and collates their responses.
+
+        call(self, target_axon, synapse=bittensor.Synapse(), timeout=12.0, deserialize=True) -> bittensor.Synapse:
+            Asynchronously sends a request to a specified Axon and processes the response.
+
+        call_stream(self, target_axon, synapse=bittensor.Synapse(), timeout=12.0, deserialize=True) -> AsyncGenerator[bittensor.Synapse, None]:
+            Sends a request to a specified Axon and yields streaming responses.
+
+        preprocess_synapse_for_request(self, target_axon_info, synapse, timeout=12.0) -> bittensor.Synapse:
+            Preprocesses the synapse for making a request, including building headers and signing.
+
+        process_server_response(self, server_response, json_response, local_synapse):
+            Processes the server response, updates the local synapse state, and merges headers.
+
+        close_session(self):
+            Synchronously closes the internal aiohttp client session.
+
+        aclose_session(self):
+            Asynchronously closes the internal aiohttp client session.
+
+    NOTE: When working with async aiohttp client sessions, it is recommended to use a context manager.
+
+    Example with a context manager:
+        >>> aysnc with dendrite(wallet = bittensor.wallet() ) as dendrite_obj:
+        >>>     print(dendrite_obj)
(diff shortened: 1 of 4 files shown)


bittensor/dendrite.py
```diff
@@ -42,15 +42,51 @@ class dendrite(torch.nn.Module):
...
+        close_session(self):
+            Synchronously closes the internal aiohttp client session.
+
+        aclose_session(self):
+            Asynchronously closes the internal aiohttp client session.
+
+    NOTE: When working with async aiohttp client sessions, it is recommended to use a context manager.
+
+    Example with a context manager:
+        >>> aysnc with dendrite(wallet = bittensor.wallet()) as d:
```
Eugene-hu, line 79: Any reason why this might be the case? Is the issue with the closure of the stream once we are finished with querying?

bittensor/dendrite.py
```diff
@@ -120,6 +192,7 @@ def query(
             new_loop = asyncio.new_event_loop()
             asyncio.set_event_loop(new_loop)
             result = loop.run_until_complete(self.forward(*args, **kwargs))
+            self.close_session()
```
Eugene-hu: I am a little confused by this loop, when the try block hits an exception, the default would run the same function again?

Eugene-hu: Also once an error has been hit, this change would now close the session and the connection. Any reason why is necessary? Would it have any effects on non-streaming error handling?

bittensor/synapse.py
```diff
@@ -454,28 +453,6 @@ def to_headers(self) -> dict:
             if field in headers or value is None:
                 continue
 
-            # Adding the tensor shape and data type to the headers if the object is a tensor
```
Eugene-hu, line 457: What were these headers used previously? Was it just metadata about the tensors inside the body?

Eugene-hu (approved):
Looks good overall, but might require additional testing before we release it. Any plans for a potential release date on this update?

tests/unit_tests/test_dendrite.py
```diff
@@ -18,11 +18,22 @@
 # DEALINGS IN THE SOFTWARE.
 
 import pytest
+import typing
 import bittensor
 from unittest.mock import MagicMock, Mock, patch
 from tests.helpers import _get_mock_wallet
 
 
+class Dummy(bittensor.Synapse):
```
eduardogr: if we want to add more test classes or share this one with other files, maybe we need to rename this class to `SynapseDummy`so we know what is that about. wdyt about that?

bittensor/dendrite.py
ifrit98: The problem is that unclosed sessions linger and issue an `UnclosedSession` warning on the `aiohttp` side but doesn't break things. Best practice to automate cleanup and wanted to minimize the client having to do that.

bittensor/dendrite.py
ifrit98: The `try` is for if the case where the current event loop doesn't exist (or has been closed) and throws an error, `catch` is to attempt to create a new loop and run it but the `close_session()` should be in `finally`. Nice catch!

RE non-streaming error handling: I haven't observed any change, but will test this out.

bittensor/synapse.py
ifrit98: Yes, merely shape-dtype information that overloaded http (same 8kb header error we ran into with text-prompting >1200 tokens). This broke pre-training subnet when the list of tensor header objects approaches >100. Has no consequence for hashing or verification

tests/unit_tests/test_dendrite.py
ifrit98: Yes, good idea.

bittensor/dendrite.py
ifrit98: Did some more reading. Looks like in the `delete` method in the `aiohttp.Client` is called during teardown but doesn't actually close the client itself, which can lead to blocking/resource hogging.

```python
    def __del__(self, _warnings: Any = warnings) -> None:
        if not self.closed:
            if PY_36:
                kwargs = {"source": self}
            else:
                kwargs = {}
            _warnings.warn(
                f"Unclosed client session {self!r}", ResourceWarning, **kwargs
            )
            context = {"client_session": self, "message": "Unclosed client session"}
            if self._source_traceback is not None:
                context["source_traceback"] = self._source_traceback
            self._loop.call_exception_handler(context)
```

bittensor/dendrite.py
ifrit98: Looks good here.

PR #1561, opened 2023-10-27
Dendrite fixes

ifrit98:
* Check for a session before attempting to close it in query `finally` block.
* Don't iterate over the `AsyncGenerator`, just return it when streaming
* Update typehints


bittensor/dendrite.py
```diff
@@ -164,9 +166,47 @@ async def aclose_session(self):
...
 
+    def _get_endpoint_url(self, target_axon, request_name):
+        endpoint = (
+            f"0.0.0.0:{str(target_axon.port)}"
+            if target_axon.ip == str(self.external_ip)
+            else f"{target_axon.ip}:{str(target_axon.port)}"
+        )
+        return f"http://{endpoint}/{request_name}"
+
+    def _handle_request_errors(self, synapse, request_name, exception):
```
Eugene-hu, line 177: Really like this new functional design!

bittensor/dendrite.py
```diff
@@ -267,17 +324,13 @@ async def single_axon_response(target_axon) -> bt.Synapse:
                     A Synapse object with the response.
                 """
                 if is_stream:
-                    # If in streaming mode, we iterate over the streaming content
-                    # and take the last item as the final result.
-                    final_result = None
-                    async for result in self.call_stream(
+                    # If in streaming mode, return the async_generator
+                    return self.call_stream(
```
Eugene-hu, line 328: If query multiple axons at the same time, this will return multiple async generators for us right? Would this have any issues on the bandwidth if we were to open 25-50 streams at once?

Eugene-hu (approved):
LGTM

bittensor/dendrite.py
ifrit98: Good question, I'll try this out on testnet

PR #1562, opened 2023-10-27
Subtensor Registry 

Eugene-hu:
- adds identity extrinsic calls to subtensor
- This PR requires https://github.com/opentensor/subtensor/pull/191 to be merged first


bittensor/utils/wallet_utils.py
```diff
@@ -102,3 +102,61 @@ def is_valid_bittensor_address_or_public_key(address: Union[str, bytes]) -> bool
...
+    """
+    Creates a dictionary with structure for identity extrinsic.
+
+    Args:
+    display (str): String to be converted and stored under 'display'.
+    legal (str): String to be converted and stored under 'legal'.
+    web (str): String to be converted and stored under 'web'.
+    riot (str): String to be converted and stored under 'riot'.
+    email (str): String to be converted and stored under 'email'.
+    pgp_fingerprint (str): String to be converted and stored under 'pgp_fingerprint'. If None, it remains None.
```
ifrit98: This requires` len(pgp_fingerprint) == 20bytes`, maybe we should make a note?

bittensor/subtensor.py
```diff
@@ -1319,6 +1319,71 @@ def root_set_weights(
...
+                    if block == None
+                    else substrate.get_block_hash(block),
+                )
+            
+        identity_info  = make_substrate_call_with_retry()
+        return bittensor.utils.wallet_utils.decode_hex_identity_dict(identity_info.value['info'])  
+
+    def update_identity(
+        self,
+        wallet: "bittensor.wallet",
```
ifrit98, line 1353: If the wallet coldkey is not the subnet owner, this can essentially set identity for any coldkey on chain. Do we want to limit this to subnet owners only? (check the chain for current subnet coldkey owners)

E.g.:
```python
owner_coldkeys = [sub.get_subnet_owner(i) for i in sub.get_all_subnet_netuids()]
if wallet.coldkey.ss58_address not in owner_coldkeys:
   raise ...
```

bittensor/utils/wallet_utils.py
```diff
@@ -102,3 +102,61 @@ def is_valid_bittensor_address_or_public_key(address: Union[str, bytes]) -> bool
...
+    web: str = "",
+    riot: str = "",
+    email: str = "",
+    pgp_fingerprint: Optional[str] = None,
+    image: str = "",
+    info: str = "",
+    twitter: str = "",
+) -> dict:
+    """
+    Creates a dictionary with structure for identity extrinsic.
```
ifrit98: Do we have the requirements for each parameter? Might be useful to document here if known.

ifrit98 (approved):
LGTM! Left a few comments, but works great!

bittensor/subtensor.py
Eugene-hu: Yeap, this is done by the chain, only subnet owners and root validators can set an identity

bittensor/utils/wallet_utils.py
Eugene-hu: Nothing specific, but there is a max of 64bytes for each param. So there is a limit there

PR #1564, opened 2023-10-28
Add btcli wallet balance

ifrit98:
Main improvement:
* Implements `btcli w balance` with default path as `~/.bittensor/wallets`
* Accepts `--wallet.path` and `--subtensor.network` main arguments

image

Bonus bugfix:
* Handle case in `btcli r get_weights` where no weights are set by a member of the root network.

image


Eugene-hu (approved):
LGTM

PR #1572, opened 2023-11-02
rename logging.py so doesn't circ import

ifrit98:
* `logging.py` in `bittensor/` can cause circular import errors and/or confusion with python's `logging.py` module.
* Rename to `btlogging.py` but don't change the references in imports (e.g. `bittensor.logging` stays the same).
image


bittensor/__init__.py
```diff
@@ -208,7 +208,7 @@ def debug(on: bool = True):
 from .chain_data import *
 from .subtensor import subtensor as subtensor
 from .cli import cli as cli, COMMANDS as ALL_COMMANDS
-from .logging import logging as logging
+from .btlogging import logging as logging
```
Eugene-hu, line 213: so this maintains the functionality for `bittensor.logging`, correct?

Eugene-hu (approved):
Nice catch! I would have that the local importing would have prevented the circular import, but I guess it doesnt cover all of the edge cases

bittensor/__init__.py
ifrit98: Yep, only changes the import. Can still call `bt.logging.<>`.

It's basically covering the edge case in jupyter notebooks where you're running python from the same directory.

PR #1573, opened 2023-11-07
don't print __is_set for recursive objects

ifrit98:


bittensor/config.py
```diff
@@ -287,12 +287,22 @@ def __deepcopy__(self, memo) -> "config":
...
+                self._remove_private_keys(v)
+        return d
+
     def __str__(self) -> str:
         # remove the parser and is_set map from the visible config
         visible = self.toDict()
         visible.pop("__parser", None)
-        visible.pop("__is_set", None)
-        return "\n" + yaml.dump(visible)
+        cleaned = self.remove_private_keys(visible)
```
Eugene-hu, line 304: This removes `__parser` and `__is_set` from the config itself, would this cause issues in the long run? Or are they one-time parameters?

ifrit98: Good to have this concern, but no need to worry! 

The `toDict()` method from the `DefaultMunch` object (which the `config` object is a subclass of) copies the content and returns them as a python dict, so we're actually not popping the state of the config object itself, but rather the copy of its content.

Eugene-hu (approved):

PR #1576, opened 2023-11-10
try/catch on dendrite async close

ifrit98:


bittensor/dendrite.py
```diff
@@ -164,7 +164,10 @@ async def aclose_session(self):
         """
         if self._session:
             await self._session.close()
-            self._session = None
+            try:
+                self._session = None
```
Eugene-hu, line 168: Mind explaining this code block a bit? Would assigning session as `None` raise an exception?

bittensor/dendrite.py
```diff
@@ -164,7 +164,10 @@ async def aclose_session(self):
...
+            except:
+                pass
```
Eugene-hu: Rather than `except` then `pass`, should we have some logging to see what the error would be, just in case?

bittensor/dendrite.py
ifrit98: Sure! It has errored for me in an edge case but I haven't been able to reproduce since.  May be more conservative than necessary. We can hold off on this if no one is having noticeable issues.

bittensor/dendrite.py
ifrit98: Yep. Added.

ifrit98:
Unnecessary. Closing.

PR #1579, opened 2023-11-14
Adds docstrings for CLI for Sphynx documentation

ifrit98:


bittensor/commands/delegates.py
```diff
@@ -53,7 +53,51 @@ def show_delegates(
...
+      Defaults to None.
+    - width (Optional[int]): The width of the console output table. Defaults to None,
+      which will make the table expand to the maximum width of the console.
+
+    The output table includes the following columns:
+    - INDEX: The numerical index of the delegate.
+    - DELEGATE: The name of the delegate.
+    - SS58: The truncated SS58 address of the delegate.
+    - NOMINATORS: The number of nominators supporting the delegate.
+    - DELEGATE STAKE(τ): The stake that is directly delegated to the delegate.
```
rajkaramchedu, line 79: This could use a bit of a clarification. In my delegator doc PR I described it as: "The amount of delegate's own stake (not the TAO delegated from any nominators). This is the the amount of stake that the delegate-owned coldkey has delegated to this delegate's hotkey (distinct from TAO delegated from others)." Is this accurate? If so, let's use the first sentence for the line 79 above.

ifrit98: Yes, that's correct. Updated.

rajkaramchedu (approved):

bittensor/commands/misc.py
```diff
@@ -52,93 +71,3 @@ def add_args(parser: argparse.ArgumentParser):
         )
 
         bittensor.subtensor.add_args(update_parser)
-
-
-class ListSubnetsCommand:
```
Eugene-hu, line 57: Are we disabling the list subnet command? I thought it was quite useful to list all of the subnets at a quick glance

Eugene-hu (approved):

bittensor/commands/misc.py
ifrit98: No, this is a duplicate. :D

PR #1586, opened 2023-11-20
Axon fix

ifrit98:
Catches a `KeyError` exception where axon's do not have the route attached that is being queries. 

This causes issues downstream and makes the error difficult to identify.

Now return a synapse populated with 404 and a helpful error message.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -748,7 +748,7 @@ class AxonMiddleware(BaseHTTPMiddleware):
         # Start of catching all exceptions, updating the status message, and processing time.
         except Exception as e:
             # Log the exception for debugging purposes.
-            bittensor.logging.trace(f"Forward exception: {traceback.format_exc()}")
+            bittensor.logging.error(f"Forward exception: {traceback.format_exc()}")
 
             # Set the status message of the synapse to the string representation of the exception.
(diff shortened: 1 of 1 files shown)


bittensor/axon.py
```diff
@@ -729,10 +729,26 @@ async def dispatch(
...
+            # synapse headers, and an empty content.
+            response = JSONResponse(
+                status_code=404, headers=synapse.to_headers(), content={}
+            )
+
         # Start of catching all exceptions, updating the status message, and processing time.
         except Exception as e:
             # Log the exception for debugging purposes.
-            bittensor.logging.trace(f"Forward exception: {traceback.format_exc()}")
+            bittensor.logging.error(f"Forward exception: {traceback.format_exc()}")
```
Eugene-hu: Just a note, I think we should start making use of the different levels of logging, ie.

`logging.error("Key Error")`
 `logging.trace(traceback)`

This way users will be notified when theres an error vs a developer being notified with the traceback of the full error

Eugene-hu (approved):

bittensor/axon.py
ifrit98: Sweet, I'll change this

PR #1594, opened 2023-11-28
Boost

unconst:
Adds commands:

`btcli r boost --netuid .. --increase ...`
`btcli r slash --netuid .. --decrease ...`

Allows root netuids to quickly increase or decrease the weights on a subnet without having to reset all of their weights across all networks. The operation performs a normalization of the weights to ensure that the weights sum to 1.

image


ifrit98 (approved):
This is nice! LGTM.

PR #1600, opened 2023-12-01
revert subtensor local to finney

ifrit98:
* Return default subtensor to `finney` for now until we get an automated process for starting `local` nodes.
* Add helpful messages to signal the coming change to `local` as default


Eugene-hu (approved):

eduardogr (approved):
clean, concise, direct.
Congratulations sir, this is very nice and helpful

PR #1606, opened 2023-12-05
[Fix] [Bug] Fix subtensor init with chain endpoint specified

camfairchild:
There is a bug during tuple unpacking that prevents one from setting a custom chain endpoint

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -329,7 +329,7 @@ class subtensor:
         self.config = copy.deepcopy(config)
 
         # Setup config.subtensor.network and config.subtensor.chain_endpoint
-        self.chain_endpoint, self.network = subtensor.setup_config(network, config)
+        self.network, self.chain_endpoint = subtensor.setup_config(network, config)
 
         if (
             self.network == "finney"


ifrit98:
It does seem like a bug! You can set a chain endpoint by overloading the `network` arg.

e.g.
```python
import bittensor as bt
bt.subtensor(network="ws://127.0.0.1:1234")
```

The reason this happens is that the [else](https://github.com/opentensor/bittensor/blob/master/bittensor/subtensor.py#L220) assumes the network arg is a `chain_endpoint` if not in the set of networks (finney, local, etc).

image

ifrit98:
This does feel unclear. Maybe we could rename the variables for clarity in this else cond?


E.g. L220
```python
...
        else:
            # Assume passed str is actually a chain endpoint.
            chain_endpoint = network
            if (
                chain_endpoint == bittensor.__finney_entrypoint__
                or "entrypoint-finney.opentensor.ai" in chain_endpoint
            ):
                return "finney", bittensor.__finney_entrypoint__
            elif (
                chain_endpoint == bittensor.__finney_test_entrypoint__
                or "test.finney.opentensor.ai" in chain_endpoint
            ):
                return "test", bittensor.__finney_test_entrypoint__
            elif (
                chain_endpoint == bittensor.__archive_entrypoint__
                or "archive.chain.opentensor.ai" in chain_endpoint
            ):
                return "archive", bittensor.__archive_entrypoint__
            elif "127.0.0.1" in chain_endpoint or "localhost" in chain_endpoint:
                return "local", chain_endpoint
            else:
                return "unknown", chain_endpoint
```

camfairchild:
> It does seem like a bug! You can set a chain endpoint by overloading the `network` arg.
> 
> e.g.
> ```python
> import bittensor as bt
> bt.subtensor(network="ws://127.0.0.1:1234")
> ```
> 
> The reason this happens is that the [else](https://github.com/opentensor/bittensor/blob/master/bittensor/subtensor.py#L220) assumes the network arg is a `chain_endpoint` if not in the set of networks (finney, local, etc).
> 
> image
> 

the bug I was trying to fix is that this doesn't work if the chain endpoint is not one of the set ones.

camfairchild:
Closing as bug wasn't real. I was instantiating a metagraph object, so it would sync without passing a subtensor object and thus default to finney.
My bad

PR #1623, opened 2023-12-15
support subnet protocols

crazydevlegend:
**Example:**
```
from bittensor.subnets.protocols.sn1 import Prompting

sn1 = bt.metagraph(1)
d = bt.dendrite()

sn1_synapse = Prompting(...)

sn1_out = d.query(sn1.axons[1], sn1_synapse)    
```


crazydevlegend:
which of the followings looks better @ifrit98 ?
1. `from bittensor.subnets import Prompting`
2. `from bittensor.subnets.sn1 import Prompting`

ifrit98:
In general, we should be branching off from `stating` rather than `master` as merges into master are limited to release branches.

ifrit98:
> which of the followings looks better @ifrit98 ?
> 
> 1. `from bittensor.subnets import Prompting`
> 2. `from bittensor.subnets.sn1 import Prompting`

We should keep this bespoke (option 2) so that we don't overcrowd the main namespace, and some subnets may be conflicting.

ifrit98:
Nice work!

Overall, this looks good and would work, but biggest concern is that this requires manual maintenance to update these every time a subnet updates its protocol. This is *somewhat* manageable with 32 subnets, but what about 64, or 128 when we potentially have more? Long-term this isn't how we should go about it, IMO.

Might there be an easier way to automatically pull the protocol from the subnet repo itself?

In the mean-time it may be beneficial to keep track of the version of each subnet's protocol (e.g. sn5 is on 1.2.3, sn4 is on 3.4.5, etc) so we can display a warning if the current implemented protocol on bittensor side is below the git repo of the subnet.

ifrit98:
Followup thought: we should probably directly keep track of the git repo URLs so they are accessible to users, potentially even displayed in logging with properl level (e..g debug/trace/warnings)

crazydevlegend:
@ifrit98 please note that protocols are automatically updated by the `Updated Subnet Protocols` action. You can refer to an example commit generated by this scheduled action (executed at 00:00 UTC every day).

PR #1633, opened 2023-12-20
Expand dendrite json and Pyandtic v2 Upgrade

ifrit98:
This updates several things:
* `synapse.dict()` ->` json.loads(synapse.json())`
* pydantic to 2.5.2 (latest version) with a massive overhaul
* fastapi to 0.105.0 (latest version)
* updates related synapse code to leverage these upgrades

This effectively increases the types that can be serialized over the bittensor network natively, e.g. `datetime`

One major bug that this fixes is the limited scope for what kind of objects can be serialized and sent over the wire. For example, `datetime` objects currently are not serializable because of calling `synapse.dict()`. The` synapse.json()` method calls an internal `pydantic` serialization routine which natively supports `datetime` objects.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -597,7 +597,7 @@ class axon:
         self.forward_fns[request_name] = forward_fn
 
         # Parse required hash fields from the forward function protocol defaults
-        required_hash_fields = request_class.__dict__["__fields__"][
+        required_hash_fields = request_class.__dict__["model_fields"][
             "required_hash_fields"
         ].default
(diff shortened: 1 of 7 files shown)


Eugene-hu (approved):
Looks good, mind expanding a bit about the differences between `.dict()` vs `json.loads`? What bug does this fix?

ifrit98:
> Looks good, mind expanding a bit about the differences between `.dict()` vs `json.loads`? What bug does this fix?

Yes, will expand in the main comment.

PR #1635, opened 2023-12-26
Add commitment support to MockSubtensor

agoncharov-reef:
This PR was intended to mock recently-added functions:
```python
subtensor.commit(wallet, netuid, data)
subtensor.get_commitment(netuid, uid)
bittensor.extrinsics.serving.publish_metadata(wallet, netuid, type, data)
bittensor.extrinsics.serving.get_metadata(netuid, hotkey)
```

In order to "fairly" mock `MockSubtensor.commit()` and `MockSubtensor.get_commitment()`, we should mock `MockSubtensor.substrate = MockSubstrate()`. This way `MockSubtensor.commit()` will under the hood call `bittensor.extrinsics.serving.publish_metadata()`  which will use our new `MockSubstrate` - so the behavior will be exactly the same as in production, except that substrate won't make an external call.

However, I see that `MockSubtensor` does not use any mocked substrate, thus it doesn't use "fair" low-level logic. This leads to following consequences:
1) Low-level functions aren't mocked. One won't be able to test `bittensor.extrinsics.serving.publish_metadata` and `bittensor.extrinsics.serving.get_metadata` since there is only "production" substrate class and they use it.
2) High-level `MockSubtensor` does not really use low-level functions, thus if some low-level function is created/modified/deleted, `MockSubtensor` will automatically diverge from original `Subtensor`. 
3) `MockSubtensor` inherits from `subtensor`, so if any new method is added to `subtensor`, then new non-mocked method appears in `MockSubtensor`.

Being said that, I would recommend to mock only real low-level network requests (the `SubstrateInterface`), so that everything else is mocked automatically now and in the future.

Anyway, taking into account existing codebase and approach, I added `MockSubtensor.commit()` and `MockSubtensor.get_commitment()` mocked methods.

Usage:
```python
from bittensor.mock.subtensor_mock import MockSubtensor
from bittensor.wallet import wallet as Wallet

netuid = 9
payload = "some data"

wallet = Wallet(name='mockwall')
wallet = wallet.create_new_hotkey(overwrite=True, use_password=False, suppress=True)
wallet = wallet.create_new_coldkey(overwrite=True, use_password=False, suppress=True)

subtensor = MockSubtensor()
subtensor.create_subnet(netuid)
subtensor.force_register_neuron(netuid, wallet.hotkey.ss58_address, wallet.coldkey.ss58_address) 
subtensor.commit(wallet, netuid, payload)

uid = subtensor.get_uid_for_hotkey_on_subnet(
	hotkey_ss58=wallet.hotkey.ss58_address,
	netuid=netuid,
)
restored_payload = subtensor.get_commitment(netuid, uid)

assert restored_payload == payload
```

### DIFF
--- a/bittensor/mock/subtensor_mock.py
+++ b/bittensor/mock/subtensor_mock.py
@@ -263,6 +263,7 @@ class MockSubtensor(subtensor):
                     "Axons": {},
                     "Prometheus": {},
                     "SubnetOwner": {},
+                    "Commits": {},
                 },
             }
 
@@ -327,6 +328,7 @@ class MockSubtensor(subtensor):
             subtensor_state["EmissionValues"][netuid][0] = 0
             subtensor_state["Burn"][netuid] = {}
             subtensor_state["Burn"][netuid][0] = 0
+            subtensor_state["Commits"][netuid] = {}
 
             # Per-UID/Hotkey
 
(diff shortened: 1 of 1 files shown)


ifrit98:
Thanks for the PR, LGTM! Are there any associated tests we need to add for this change? 

Using `MockSubstrate` is a good idea. We should incorporate this, however will change a lot of the bespoke existing test logic we have established using the current structure, but will be worth the effort. Openly welcome more PRs! (Added this to the backlog into a planned mocking/testing overhaul).

P.S. In the future, would you mind branching off from `staging` so we can streamline the inclusion process? Nice job!

ifrit98 (approved):

agoncharov-reef:
@ifrit98 Thank you! Agreed.
I will branch off from `staging` onwards, sorry for confusion.

PR #1637, opened 2024-01-02
Feature/axon gossip p2p

ifrit98:
New feature: Gossip protocol to discover nodes and extend for message passing among validators (or any nodes) on a subnet. This will be useful for coordinating across nodes (particularly validators) for implementing business logic to more easily facilitate utilization of the subnet for external (and internal) applications.

This feature can be enabled by passing the `enable_gossip` to the `axon` upon creation.

Built-in:
* Peer discovery of validators based on `vtrust`. By running the `axon` with `gossip=True` you turn on this behavior by default.

Extendable:
* Can extend this gossip protocol to create any p2p messaging structure desired on the subnet. 
* Any gossip p2p protocol will be built upon the peer discovery by default, that is easily configurable.


ifrit98:
Stale. May come back to this idea in the future.

PR #1638, opened 2024-01-02
feat(wallets.py): add wallet history command

saqib-codes-11:
Description :-

This PR adds the following new command to fetch all the transfers associated with the provided "Wallet Name".

Command - btcli wallet history --wallet.name xyz 

Screenshot -

Screenshot 2024-01-02 at 09 58 52

### DIFF
--- a/bittensor/commands/wallets.py
+++ b/bittensor/commands/wallets.py
@@ -24,6 +24,7 @@ from rich.table import Table
 from typing import Optional, List
 from . import defaults
 import requests
+from ..utils import RAOPERTAO
 
 
 class RegenColdkeyCommand:
@@ -878,7 +879,7 @@ class GetWalletHistoryCommand:
     def run(cli):
         r"""Check the transfer history of the provided wallet."""
         wallet = bittensor.wallet(config=cli.config)
-        wallet_address = wallet.coldkey.ss58_address
+        wallet_address = wallet.get_coldkeypub().ss58_address
         # Fetch all transfers
         transfers = get_wallet_transfers(wallet_address)
 
@@ -982,11 +983,15 @@ def create_transfer_history_table(transfers):
 
     # Add rows to the table
     for item in transfers:
+        try:
+            tao_amount = int(float(item["amount"])) / RAOPERTAO
+        except:
+            tao_amount = item["amount"]
         table.add_row(
             item["id"],
             item["from"],
             item["to"],
-            item["amount"],
+            str(tao_amount),
             str(item["extrinsicId"]),
             item["blockNumber"],
         )
--- a/bittensor/commands/wallets.py
+++ b/bittensor/commands/wallets.py
@@ -830,6 +830,30 @@ class WalletBalanceCommand:
             ) = bittensor.subtensor.determine_chain_endpoint_and_network(str(network))
 
 
+API_URL = "https://api.subquery.network/sq/TaoStats/bittensor-indexer"
+
+GRAPHQL_QUERY = """
+query ($first: Int!, $after: Cursor, $filter: TransferFilter, $order: [TransfersOrderBy!]!) {
+    transfers(first: $first, after: $after, filter: $filter, orderBy: $order) {
+        nodes {
+            id
+            from
+            to
+            amount
+            extrinsicId
+            blockNumber
+        }
+        pageInfo {
+            endCursor
+            hasNextPage
+            hasPreviousPage
+        }
+        totalCount
+    }
+}
+"""
+
+
 class GetWalletHistoryCommand:
     """
     Executes the 'history' command to fetch the latest transfers of the provided wallet on the Bittensor network.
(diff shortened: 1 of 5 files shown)


bittensor/commands/wallets.py
```diff
@@ -827,3 +828,163 @@ def check_config(config: "bittensor.config"):
                 _,
                 config.subtensor.chain_endpoint,
             ) = bittensor.subtensor.determine_chain_endpoint_and_network(str(network))
+
+
+API_URL = "https://api.subquery.network/sq/TaoStats/bittensor-indexer"
```
ifrit98, line 834: Do we have some metrics on the efficiency and load on this endpoint?  It will be hit quite a lot.

bittensor/commands/wallets.py
```diff
@@ -827,3 +828,163 @@ def check_config(config: "bittensor.config"):
...
+                {"to": {"equalTo": wallet_address}},
+            ]
+        },
+        "order": "BLOCK_NUMBER_DESC",
+    }
+
+    transfers = []
+
+    # Make the request with the provided query and variables until all transfers are fetched
+    while True:
```
ifrit98: Generally weary of unbounded `while True` statements in python. This technically *should* terminate on the desired condition, but may want to explicitly put the exit condition in this line and/or specify some other conditions for exiting. For example, if there is significant txn volume, we may want to limit the output to `n` txns.

For example, adding a `max_txn_history` that we exit upon reaching:

```python
max_txn_history = 1000 # this should be set in the config.
page_info = {"page_info": {"hasNextPage": True}} # for initial loop entry, will be filled by `transfer_data`
while len(transfers) < max_txn_history and page_info.get("hasNextPage") == True:
    ...
```

What do you think?

ifrit98 (changes requested):
Thanks for the contribution! Looks good overall, left you some comments with a request for updating the `while` loop.

ifrit98:
Also, in the future, would you mind branching off from `staging`? This is the preferred workflow outlined in the `contrib` docs for development.

Thanks!

ifrit98:
Any new feature should include proper tests for them, including new `btcli` functions.  Make sure to add this as well in `tests/integration_tests/test_cli.py` and/or `tests/integration_tests/test_cli_no_network.py` where appropriate.

Thanks!

bittensor/commands/wallets.py
saqib-codes-11: Not sure about this. I can perform a basic load test on this using Apache AB. Nonetheless, I am open to suggestions!

bittensor/commands/wallets.py
saqib-codes-11: Yeah, I think it will be a better approach. Will be fixing this in the next commit.

bittensor/commands/wallets.py
ifrit98: That should be enough. Doesn't need to be extravagant, just want to get a sense of how much stress this would take before having issues.

saqib-codes-11: I am attaching the load test results which contain the below scenarios,

1) Concurrency - 50, Total Requests - 1000
2) Concurrency - 100, Total Requests - 10000
3) Concurrency - 250, Total Requests - 10000

250 is the limit, going above it gives socket error - socket: Too many open files (24)

[load_test.txt](https://github.com/opentensor/bittensor/files/13817638/load_test.txt)

saqib-codes-11:
> Any new feature should include proper tests for them, including new `btcli` functions. Make sure to add this as well in `tests/integration_tests/test_cli.py` and/or `tests/integration_tests/test_cli_no_network.py` where appropriate.
> 
> Thanks!

Added

bittensor/commands/wallets.py
```diff
@@ -827,3 +828,171 @@ def check_config(config: "bittensor.config"):
...
+    Note:
+    This command is essential for users to monitor their financial status on the Bittensor network.
+    It helps in fetching info on all the transfers so that user can easily tally and cross check the transactions.
+    """
+
+    @staticmethod
+    def run(cli):
+        r"""Check the transfer history of the provided wallet."""
+        wallet = bittensor.wallet(config=cli.config)
+        wallet_address = wallet.coldkey.ss58_address
```
ifrit98: It is somewhat cumbersome to enter the password each time to simply get the coldkey address, which is not exactly private data.

Suggestion:
```python
wallet_address = wallet.get_coldkeypub().ss58_address
```

bittensor/commands/wallets.py
```diff
@@ -827,3 +828,171 @@ def check_config(config: "bittensor.config"):
...
+            no_wrap=no_wrap,
+        )
+
+    # Add rows to the table
+    for item in transfers:
+        table.add_row(
+            item["id"],
+            item["from"],
+            item["to"],
+            item["amount"],
```
ifrit98: The current amount displayed is in `rao`, and displaying in `tao` may be more user-friendly.  It's a simple divide by `1e9`

Suggestion:
```python
int(item["amount"]) / 1e9
```

May want to do some error handling if the value cannot be converted to an `int`.

ifrit98 (changes requested):
Thanks for the updated tests and testing out the query. Looks good!

Two small nits for you and we should be good to merge: 
(1) Don't unlock coldkey to get `ss58_address`
(2) Display values in `tao` instead of `rao` (and mention this in the column, e.g. `Amount (tao)`

Suggestion for a nice to have, but is not necessary IMO:
(3) May be nice to add a column for `Time`, translating the block number into datetimes. What do you think?
```bash
... |  Block No  | Time
... |   2044685  | Nov 1, 2023 12:21:00 
...
```

bittensor/commands/wallets.py
alex0500: You not need while. Change 10 to 1000 or more

alex0500:
Where is time of transfers?

bittensor/commands/wallets.py
saqib-codes-11: This is done.

bittensor/commands/wallets.py
saqib-codes-11: This is also done. Tested it for the below values, 
2
2.9
"2"
"2.989767"

Any invalid integer will throw an error, in that case, I am using it as it as. Any other suggestions on the same?

saqib-codes-11:
> Thanks for the updated tests and testing out the query. Looks good!
> 
> Two small nits for you and we should be good to merge: (1) Don't unlock coldkey to get `ss58_address` (2) Display values in `tao` instead of `rao` (and mention this in the column, e.g. `Amount (tao)`
> 
> Suggestion for a nice to have, but is not necessary IMO: (3) May be nice to add a column for `Time`, translating the block number into datetimes. What do you think?
> 
> ```shell
> ... |  Block No  | Time
> ... |   2044685  | Nov 1, 2023 12:21:00 
> ...
> ```

I did complete the first two changes but I was not able to get the correct time using the block number, 

I tried the below formula to calculate it but was getting a 1-day difference.

```
from datetime import datetime, timedelta
time_for_target_block = timedelta(seconds=(target_block_number - 1) * substrate_block_time)
target_block_time = first_block_time + time_for_target_block

```

bittensor/commands/wallets.py
saqib-codes-11: @alex0500 removed while and set max_txn to 1000.

ifrit98 (approved):
LGTM. I'll take it from here

PR #1667, opened 2024-01-23
ensure branch off from staging and rm old docs

ifrit98:
Update the contrib docs to clearly mark branching off from `staging` as a requirement and deprecate old docs suggestion.


garrett-opentensor (approved):
Looks good

PR #1681, opened 2024-01-27
burn -> recycle for public facing code

ifrit98:
Make it obvious that we are not literally "burning" tao, but rather recycling it back into the mechanism for client facing functions.


Saevitus0 (approved):
Just changing burn to recycle, shouldn't cause any issues.

thealligatorking (approved):

PR #1687, opened 2024-02-01
wrap set weights in a ttl multiprocessing call so we don't hang past TTL

ifrit98:
Defaults to 100 sec

### DIFF
--- a/bittensor/extrinsics/set_weights.py
+++ b/bittensor/extrinsics/set_weights.py
@@ -22,14 +22,48 @@ import torch
 from rich.prompt import Confirm
 from typing import Union
 import bittensor.utils.weight_utils as weight_utils
+import multiprocessing
 
 from loguru import logger
 
 logger = logger.opt(colors=True)
 
 
-def set_weights_extrinsic(
+def ttl_set_weights_extrinsic(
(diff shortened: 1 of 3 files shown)


garrett-opentensor (approved):

thealligatorking (approved):

bittensor/extrinsics/set_weights.py
```diff
@@ -38,10 +39,47 @@ def set_weights_extrinsic(
...
+    subtensor_endpoint: str,
+    wallet: "bittensor.wallet",
+    netuid: int,
+    uids: Union[torch.LongTensor, list],
+    weights: Union[torch.FloatTensor, list],
+    version_key: int = 0,
+    wait_for_inclusion: bool = False,
+    wait_for_finalization: bool = False,
+    prompt: bool = False,
+    ttl: int = 100,
```
thealligatorking: Looks like the param `ttl` is not used in the func.

garrett-opentensor: Good catch. yeah that parameter should only be relevant for `ttl_set_weights_extrinsic` for the timeout feature

ifrit98: removed

PR #1696, opened 2024-02-08
Fix: MaxAttemptedException -> MaxAttemptsException

thealligatorking:
Except block used undefined `MaxAttemptedException`, removed and replaced with exception that is defined and is raised  `MaxAttemptsException`

image


ifrit98 (approved):
Nice catch!

PR #1711, opened 2024-02-21
fix: Moves test wallets to tmp directory 

thealligatorking:
* Moves test wallets to tmp directory as to not bloat `~/.bittensor/wallets/`
* Fix applied to `mock-` , `mock-legacy`, `mock-empty`


ifrit98 (approved):
Great, simple and avoids headache.

PR #1715, opened 2024-02-22
Replace rich console with loguru

brueningf:
Addresses "Action 4" 

- Replaces rich console print calls with new ConsoleLogger calls while preserving color, and icons.  
  (Except tables, which are still handled with rich console)

- Adds a unit test for ConsoleLogger.
- Modifies MockConsole to work with ConsoleLogger.


ifrit98:
Going a different direction. Python native logger

PR #1717, opened 2024-02-23
Fixed circular function calling in subtensor.py & updated funcs still using the old compose extrinsic calls w/ the new send_extrinsic

Saevitus0:
removed `_do_` extrinsic functions from subtensor.py, and implemented them as inner functions in their related extrinsic func, as well as updated some of the remaining extrinsic calls that weren't using the new send_extrinsic() func.

formatted all with black.

### DIFF
--- a/bittensor/extrinsics/delegation.py
+++ b/bittensor/extrinsics/delegation.py
@@ -33,6 +33,25 @@ def nominate_extrinsic(
     wait_for_finalization: bool = False,
     wait_for_inclusion: bool = True,
 ) -> bool:
+    def _do_nominate() -> bool:
+        response = subtensor.send_extrinsic(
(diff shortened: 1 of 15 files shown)


bittensor/extrinsics/delegation.py
```diff
@@ -33,6 +33,25 @@ def nominate_extrinsic(
     wait_for_finalization: bool = False,
     wait_for_inclusion: bool = True,
 ) -> bool:
+    def _do_nominate() -> bool:
```
thealligatorking: You may want to have the optional string in the type hint here due to line 47

thealligatorking:
@Saevitus0 tests still failing and needs rebase

PR #1722, opened 2024-02-26
Fix: Raises error when exceeding block max on metagraph

thealligatorking:
* Raises error now for this [Note](https://github.com/opentensor/bittensor/blob/master/bittensor/metagraph.py#L495)
* Sets block max at `300`.

### DIFF
--- a/bittensor/metagraph.py
+++ b/bittensor/metagraph.py
@@ -499,6 +499,12 @@ class metagraph(torch.nn.Module):
 
                 subtensor = bittensor.subtensor(network='archive')
         """
+
+        if block > 300:
+            raise ValueError(
+                "Block number is greater than 300. Please use the 'archive' network for subtensor and retry."
+            )
+
         # Initialize subtensor
         subtensor = self._initialize_subtensor(subtensor)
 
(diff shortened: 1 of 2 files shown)


bittensor/metagraph.py
```diff
@@ -499,6 +499,12 @@ def sync(
 
                 subtensor = bittensor.subtensor(network='archive')
         """
+
+        if block > 300:
+            raise ValueError(
```
ifrit98, line 509: Can we use `bittensor.logging.error` here?  Can still raise a value error, but should be logged this way as well

ifrit98 (approved):
LGTM!

bittensor/metagraph.py
thealligatorking: Yep!

ifrit98: Looks like we're getting test errors now

tests/unit_tests/test_metagraph.py
```diff
@@ -133,7 +133,7 @@ def mock_subtensor():
 # Mocking the metagraph instance for testing purposes
 @pytest.fixture
 def metagraph_instance():
-    metagraph = Metagraph(netuid=1, sync=False)
+    metagraph = Metagraph(netuid=1337, sync=False)
```
ifrit98, line 136: :)

PR #1725, opened 2024-02-26
Feature: Synapse passing type check

thealligatorking:
*  `bittensor/synapse.py` updated to pass type checking
*  mypy.ini uncommented `synapse` block so that mypy runs type checking against `bittensor/synapse.py`


bittensor/synapse.py
```diff
@@ -89,7 +89,7 @@ def cast_float(raw: str) -> float:
         float or None: The converted float, or ``None`` if the input was ``None``.
 
     """
-    return float(raw) if raw != None else raw
+    return float(raw) if raw != None else raw  # type: ignore
```
wolfentensor, line 92: Using var != None for checking if a variable is not None is generally considered less pythonic and potentially error-prone compared to using var is not None. The main reasons are:

1. Readability: is not is more readable and explicitly states the intention to check for identity with None, rather than checking for equality. It makes the code easier to understand at a glance.

2. Correctness: is checks for identity, meaning it checks whether both sides point to the same object in memory. This is the recommended way to compare something with None, as there is only one instance of None in a Python program. Using != checks for equality and might lead to unexpected results if the object being compared defines its own __eq__ method to handle equality comparisons.

3. Consistency: Following Python's PEP 8 style guide enhances consistency across Python codebases. PEP 8 suggests using is not for comparisons with None.

Therefore, if var is not None: is preferred for checking if a variable is not None because it's clearer, more accurate, and aligns with Python's best practices.

bittensor/synapse.py
```diff
@@ -73,7 +73,7 @@ def cast_int(raw: str) -> int:
         int or None: The converted integer, or ``None`` if the input was ``None``.
 
     """
-    return int(raw) if raw != None else raw
+    return int(raw) if raw != None else raw  # type: ignore
```
wolfentensor, line 76: See comment at line 92

bittensor/synapse.py
```diff
@@ -24,7 +24,7 @@
 import pydantic
 from pydantic.schema import schema
 import bittensor
-from typing import Optional, List, Any
+from typing import Optional, List, Any, Dict
```
wolfentensor, line 27: Please organize imports according to PEP 8:

1. Standard Library Imports: First, import standard library modules. These are the modules that come with Python itself, such as os and sys.

2. Related Third Party Imports: After the standard library, import modules from third-party libraries that are not part of the Python Standard Library, such as Flask from Flask web framework or requests.

3. Local Application/Library Specific Imports: Lastly, import modules that are specific to your application or library. These will be any bittensor imports.

Additionally, within each of these groups, the imports should be sorted alphabetically, and each group should be separated by a blank line to make the distinctions clear and the code easy to read. This organization helps in quickly understanding the dependencies of the module and distinguishes between different types of modules, which can be especially useful in large codebases.

This can be done automatically with isort:
pip install isort
isort filename.py

bittensor/synapse.py
```diff
@@ -526,7 +526,7 @@ def is_success(self) -> bool:
         Returns:
             bool: ``True`` if dendrite's status code is ``200``, ``False`` otherwise.
         """
-        return self.dendrite.status_code == 200
+        return self.dendrite is not None and self.dendrite.status_code == 200
```
wolfentensor, line 529: A better way to do this might be:

return getattr(self.dendrite, 'status_code', None) == 401

- attempts to get the status_code attribute from self.dendrite.
- If self.dendrite is None or doesn't have a status_code attribute, it defaults to None.
- then checks if the result of getattr is 401.
- returns True / False

This avoids an AttributeError exception in case .status_code does not exist.

bittensor/synapse.py
```diff
@@ -689,7 +691,10 @@ def body_hash(self) -> str:
 
         for field, value in instance_fields.items():
             # If the field is required in the subclass schema, hash and add it.
-            if field in self.required_hash_fields:
+            if (
+                self.required_hash_fields is not None
+                and field in self.required_hash_fields
+            ):
```
wolfentensor, line 697: A better way to do this might be:

if field in (getattr(self, 'required_hash_fields', []) or []):

This expression first tries to get self.required_hash_fields. If self.required_hash_fields is None or does not exist, getattr will return the default value, which is an empty list.

This approach ensures that the code does not raise an AttributeError if self.required_hash_fields does not exist as an attribute of self, it also handles if the attribute is explicitly set to None and avoids a TypeError. The use of getattr with a default empty list simplifies the condition and makes the code safe to run even in scenarios where the attribute might be missing, enhancing the resilience and fault tolerance of the code.

wolfentensor (changes requested):
The code you've written to check for the presence of None in attributes is clear and functionally sound. This explicit handling demonstrates good defensive programming practices. However, considering maintainability and readability, you might find it beneficial to explore more concise ways to achieve the same functionality. Utilizing Python's ability to handle such checks in a more streamlined manner could reduce verbosity and improve code clarity. 

We need to follow PEP 8 styleguide on all code moving forward. The essence of following PEP 8 is not just about adhering to rules for the sake of rules but about writing code that is clean, easy to read, and easy for others to understand and maintain. This will produce high-quality, professional code that stands out in the Python community.

ifrit98 (approved):
LGTM!

PR #1731, opened 2024-03-04
Expands type checking  to subtensor

thealligatorking:
* Efficiencies - `var is None` instead of ` var == None`
* Object paramater assurances - `object.value` -> `if hasattr(object, "value")`
* Optional types
* Notes/comments left on implementations where use of method / func was questionable. (see: TODOs)


bittensor/metagraph.py
```diff
@@ -748,7 +748,7 @@ def _process_root_weights(
 
         """
         data_array = []
-        n_subnets = subtensor.get_total_subnets()
+        n_subnets = subtensor.get_total_subnets() or 0
```
ifrit98, line 751: image

Think this line (or some subtensor change) breaks loading the metagraph.  Please check this and try loading the metagraph to test this out

ifrit98 (approved):
LGTM!

PR #1733, opened 2024-03-06
handle req args by parsing and raising

ifrit98:


bittensor/config.py
```diff
@@ -123,6 +123,14 @@ def __init__(
         if args == None:
             args = sys.argv[1:]
 
+        # Check for missing required arguments before proceeding
+        missing_required_args = self.__check_for_missing_required_args(parser, args)
+        if missing_required_args:
+            # Handle missing required arguments gracefully
+            raise ValueError(
```
wolfentensor, line 130: The proper exception to raise is argparse.ArgumentError:

wolfentensor (approved):
The added code is functionally sound and suited for its purpose. One small comment about the exception type.

bittensor/config.py
ifrit98: This will not work here because we don't have access the the arg reference.

PR #1739, opened 2024-03-07
Fix: proper association with wallet dir and coldkey addr

thealligatorking:
Proper association with wallet directory names and coldkey addresses.


ifrit98 (approved):
Nice catch! LGTM

PR #1741, opened 2024-03-07
Fixed event lookup on new network added

shibshib:
This fixes the event lookup logic when a new sub network is added


ifrit98 (approved):
LGTM!

PR #1745, opened 2024-03-11
Allow set subnet specific hyper parameters.

shibshib:
This PR adds and fixes 4 subnet hyper parameter settings for subnet owners. Owners are now able to explicitly set Rho, Kappa, Difficulty, and bonds moving average on chain for their own subnet.

Users can see hyperparams for subnets using btcli. They can also sudo set hyperparams on subnets they own using btcli sudo set.


Eugene-hu (approved):
LGTM, but the tests are failing and you would want to increase the bittensor version and

PR #1749, opened 2024-03-19
Update install.sh

cisterciansis:
updated the "tip" commands after installation to function with the latest Bittensor version


ifrit98 (changes requested):
Resolve merge conflicts and branch off from `staging` as per `contrib` documentation.

PR #1751, opened 2024-03-19
Logging Refactor

sepehr-opentensor:
This PR removes the `loguru` dependency and replaces the underlying logging mechanism with the standard Python built-in `logging` library for compatibility with 3rd party modules and libraries. 

Making this change will grant visibility into these 3rd party libraries for performance evaluation and debugging, which is inherently necessary for future updates to bittensor and `btcli`. These changes also preserve the existing API, and as such, users should remain unaffected.


ifrit98 (approved):
LGTM!

birkskyum:
This breaking change was introduced in a minor release

PR #1760, opened 2024-03-27
Change extrinsic submission flow

camfairchild:
This changes the flow of extrinsic submission.  
Currently, the flow suffers from a double-compose issue.  
Current flow if the submission has an exception 
(assume the initial submission is included in the chain, regardless of the client exception):
1. call `_do_transfer`
2. create new transfer extrinsic 
3. sign extrinsic using *new nonce*
4. submit extrinsic
5. ERROR!
6. retry from 2.
7. SUCCESS!

In the above example, now there have been 2 *valid* transfer extrinsics submitted (and included) to the chain
This is **bad**, this is a double-*send* by the Python API

This PR changes the flow to retry *only* the *submission* portion. (i.e., 6. retries from 4.)


thealligatorking:
@camfairchild do we still need this?

camfairchild:
> @camfairchild do we still need this?

if we still do this, then yes. I'm not certain we fixed it

PR #1764, opened 2024-04-02
_do_burned_register fix 

DaryxXx:
In _do_burned_register, when wait_for_finalization = False and wait_for_inclusion =  False it was returning only 1 var when expecting a tuple as output.


thealligatorking:
Closing as we have this in staging.

PR #1776, opened 2024-04-07
Axon nonce backwards compat master

ifrit98:


thealligatorking (approved):

Dena327:
Nive

thealligatorking:
@ifrit98 do we still need this or safe to close?

bittensor/axon.py
```diff
@@ -914,14 +988,46 @@ async def default_verify(self, synapse: bittensor.Synapse):
...
-            # Check the nonce from the endpoint key.
-            if (
-                endpoint_key in self.nonces.keys()
-                and self.nonces[endpoint_key] is not None
-                and synapse.dendrite.nonce is not None
-                and synapse.dendrite.nonce <= self.nonces[endpoint_key]
-            ):
-                raise Exception("Nonce is too small")
+            # Use new nonce protocol if the caller is using the new protocol.
+            if hasattr(synapse.dendrite, "unix_timestamp"):
```
mjurbanski-reef, line 988: this will be always `True` and "old protocol" fallback will never be executed

i.e. if server has installed new bittensor, then TerminalInfo class has unix_timestamp field, therefore `hasattr(synapse.dendrite, "unix_timestamp") is True` and client side can do nothing about it.

bittensor/axon.py
```diff
@@ -151,6 +151,61 @@ def stop(self):
...
+
+    nonce_dirname = str(hash(f"{coldkey}:{hotkey}:{ip}:{port}"))
+    nonces_path = os.path.join(nonces_basepath, nonce_dirname, "nonces.json")
+
+    if os.path.exists(nonces_path):
+        with open(nonces_path, "r") as f:
+            nonces = json.load(f)
+        return nonces
+
+    return {}
```
mjurbanski-reef:
```suggestion
    nonce_dirname = str(hash(f"{coldkey}:{hotkey}:{ip}:{port}"))
    nonces_path = os.path.join(nonces_basepath, nonce_dirname, "nonces.json")

    try:
        with open(nonces_path, "r") as f:
            nonces = json.load(f)
        return nonces
    except (FileNotFoundError, json.JSONDecodeError):
        return {}
```
other than less code, the important difference is that this one doesn't contain race condition if file was deleted for some reason.  The second important difference is that, if file was not completed (due error during write, or just because write operation in other process has yet to complete), it will not error out.

I think it would be best to add read/write lock here if we wanted this to be rock solid, but I guess this will do.

bittensor/axon.py
```diff
@@ -151,6 +151,61 @@ def stop(self):
...
+    Args:
+        nonces_path (str): The file path to save the nonces to.
+        coldkey (str): The coldkey to use in the hash.
+        hotkey (str): The hotkey to use in the hash.
+        ip (str): The ip to use in the hash.
+        port (int): The port to use in the hash.
+        nonces (Dict[str, int]): A dictionary containing the (endpoint_key, nonce) pairs to save to the file.
+    """
+    if not os.path.exists(nonces_basepath):
+        os.makedirs(nonces_basepath)
```
mjurbanski-reef:
```suggestion
    os.makedirs(nonces_basepath, exists_ok=True)
```
less code and one less race condition

bittensor/axon.py
```diff
@@ -151,6 +151,61 @@ def stop(self):
...
+
+    nonce_dirname = str(hash(f"{coldkey}:{hotkey}:{ip}:{port}"))
```
mjurbanski-reef, line 198: `hash(...)` is a wrong function to use here, as it is seeded on python startup, so each new python process will have a different value here

```
 rooter@reef-vm ➤ ~ ➤ python3 -c 'print(hash("a"))'
642192925298734290
 rooter@reef-vm ➤ ~ ➤ python3 -c 'print(hash("a"))' 
-3386191752027622546
 rooter@reef-vm ➤ ~ ➤ python3 -c 'print(hash("a"))'
267983391070059639
```

PR #1784, opened 2024-04-14
feat: Create btcli profile config command

thealligatorking:
* Adds `btcli profile create` command to cli.
* When ran, user will be prompted with a series of questions pertaining to profile creation:

image

* Writes config file to be use for profile configuration ingest
image


ibraheem-abe (approved):
Lgtm!

PR #1786, opened 2024-04-15
Replace torch with numpy

sepehr-opentensor:


bittensor/dendrite.py
```diff
@@ -122,6 +121,9 @@ def __init__(
 
         self._session: Optional[aiohttp.ClientSession] = None
 
+    async def __call__(self, *args, **kwargs):
+        return await self.forward(*args, **kwargs)
+
```
thewhaleking, line 126: This is done because of the dendrite no longer subclassing.

bittensor/metagraph.py
```diff
@@ -858,43 +864,38 @@ def load_from_path(self, dir_path: str) -> "metagraph":
...
-        self.validator_permit = torch.nn.Parameter(
-            state_dict["validator_permit"], requires_grad=False
-        )
-        self.uids = torch.nn.Parameter(state_dict["uids"], requires_grad=False)
+        graph_filename = latest_block_path(dir_path)
+        try:
+            with open(graph_filename, "rb") as graph_file:
+                state_dict = pickle.load(graph_file)
+        except pickle.UnpicklingError:
+            import torch
```
thewhaleking: This assumes the unpickling error only relates to trying to load a previous (pre-7.0 save). We should also tell the user what's going on here.

bittensor/utils/__init__.py
```diff
@@ -51,6 +50,8 @@ def unbiased_topk(values, k, dim=0, sorted=True, largest=True):
         indices: (torch.LongTensor)
             indices of the topk values.
     """
+    import torch
```
thewhaleking: This will cause an import error, but without an error message explaining to the user why. Though this is likely only going to be used if the dev requirements are installed, a try catch with error logging is still important.

bittensor/utils/weight_utils.py
```diff
@@ -209,26 +210,26 @@ def convert_weights_and_uids_for_emit(
...
+) -> Tuple[NDArray[np.int64], NDArray[np.float32]]:
     bittensor.logging.debug("process_weights_for_netuid()")
     bittensor.logging.debug("weights", weights)
     bittensor.logging.debug("netuid", netuid)
     bittensor.logging.debug("subtensor", subtensor)
     bittensor.logging.debug("metagraph", metagraph)
 
     # Get latest metagraph from chain if metagraph is None.
-    if metagraph == None:
+    if metagraph is None:
```
thewhaleking, line 227: Nice 😎

bittensor/utils/weight_utils.py
```diff
@@ -209,26 +210,26 @@ def convert_weights_and_uids_for_emit(
...
-    uids,
-    weights: torch.Tensor,
+    uids: NDArray[np.int64],
+    weights: NDArray[np.float32],
     netuid: int,
     subtensor: "bittensor.subtensor",
     metagraph: "bittensor.metagraph" = None,
     exclude_quantile: int = 0,
-) -> torch.FloatTensor:
+) -> Tuple[NDArray[np.int64], NDArray[np.float32]]:
```
thewhaleking, line 219: This is a good catch.

tests/integration_tests/test_metagraph_integration.py
```diff
@@ -56,6 +57,26 @@ def test_load_sync_save(self):
...
+
+        def deprecated_save_torch(metagraph):
+            save_directory = get_save_dir(metagraph.network, metagraph.netuid)
+            os.makedirs(save_directory, exist_ok=True)
+            graph_filename = save_directory + f"/block-{metagraph.block.item()}.pt"
+            print(graph_filename)
+            state_dict = metagraph.state_dict()
+            print(state_dict)
+            for key in METAGRAPH_STATE_DICT_NDARRAY_KEYS:
+                print(key, state_dict[key])
```
thewhaleking: Remove these print calls.

thewhaleking (approved):
Did some small fixes. Overall, g2g

PR #1797, opened 2024-04-16
Logging: Added _primary_loggers

ibraheem-abe:
Edited:
- `_primary_loggers` replaced `_name`
- Usage is updated for `_name` 
- Methods to add/remove loggers in `_primary_loggers`

### DIFF
--- a/bittensor/btlogging/loggingmachine.py
+++ b/bittensor/btlogging/loggingmachine.py
@@ -69,7 +69,7 @@ class LoggingMachine(StateMachine):
         # basics
         super(LoggingMachine, self).__init__()
         self._queue = mp.Queue(-1)
-        self._name = name
+        self._primary_loggers = {name}
         self._config = config
 
         # Formatters
@@ -172,9 +172,15 @@ class LoggingMachine(StateMachine):
         file_handler.setLevel(stdlogging.TRACE)
         return file_handler
 
+    def register_primary_logger(self, name: str):
+        self._primary_loggers.add(name)
+
+    def deregister_primary_logger(self, name: str):
+        self._primary_loggers.remove(name)
+
     def enable_third_party_loggers(self):
         for logger in all_loggers():
-            if logger.name == self._name:
+            if logger.name in self._primary_loggers:
                 continue
             queue_handler = QueueHandler(self._queue)
             logger.addHandler(queue_handler)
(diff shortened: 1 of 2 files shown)


bittensor/btlogging/loggingmachine.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, config: bittensor.config, name: str = BITTENSOR_LOGGER_NAME):
         self._queue = mp.Queue(-1)
         self._name = name
         self._config = config
+        self._first_party_loggers = {name}
```
sepehr-opentensor, line 74: I think, instead of having separate fields, change it to something like this:

```python
def __init__(self, config: bittensor.config, name: str = BITTENSOR_LOGGER_NAME):
    # basics
    super(LoggingMachine, self).__init__()
    self._queue = mp.Queue(-1)
    self._primary_loggers = {name}
    self._config = config

def register_primary_logger(name: str):
    self._primary_loggers.add(name)

def deregister_primary_logger(name: str):
    self._primary_loggers.remove(name)
```

this way the main `bittensor` logger isn't a special case that has to be handled separately.

bittensor/btlogging/loggingmachine.py
```diff
@@ -172,9 +173,23 @@ def _create_file_handler(self, logfile: str):
         file_handler.setLevel(stdlogging.TRACE)
         return file_handler
 
+    def enable_first_party_logger(self, logger_name: str):
```
sepehr-opentensor: these functions aren't necessary, as the enabling/disabling logger logic is handled by the various state transitions (trace, debug, on, and disabled)

bittensor/btlogging/loggingmachine.py
sepehr-opentensor: You'll need to make additional changes, there is another reference to `self._name` (line 229) that needs to be changed to reflect the use of `self._primary_loggers` instead, as described above

sepehr-opentensor (changes requested):
The enabling/disabling of the primary logger is already handled by the state transition logic, you would really only have to replace `self._name` with a set of strings, and handle accordingly.

ifrit98:
Documentation for this?

ibraheem-abe:
Thanks for pointing this out. I'll add documentation for this functionality and also discuss to get these docs ([here](https://docs.bittensor.com/python-api/html/autoapi/bittensor/btlogging/index.html)) updated as well.

thealligatorking (approved):

PR #1802, opened 2024-04-18
Fix: Wallet overwrite functionality

ibraheem-abe:
Bug: The `--overwrite_coldkey` was not working as expected when regenerating coldkeys: the user was still prompted to enter y/N to confirm the action

Description:
- add_argument actions were storing the value False (even if the flag was passed as argument)
- Fix is propagated to other applicable methods as well.


thealligatorking:
@ibraheem-opentensor is it possible to include tests for this?

thealligatorking (approved):

PR #1803, opened 2024-04-21
allow forward_fn to accept more arguments

mjurbanski-reef:
AFAIK there is no downside, and it does allow some extra innovation when handling request by allowing to monitor HTTP headers or other params etc directly in `forward_fn`. This also helps with debugging.

While enabling it, I refactored this function a little to reduce code duplication & improve the readability of it.

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -466,27 +466,19 @@ class axon:
             offered by this method allows developers to tailor the Axon's behavior to specific requirements and
             use cases.
         """
-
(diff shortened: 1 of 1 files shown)


thealligatorking:
@mjurbanski-reef thank you! We will test this out to ensure backwards compatibility.

thealligatorking:
@mjurbanski-reef we plan on getting this into `7.0.0`

mjurbanski-reef:
superseded by #1822 which includes other related changes

PR #1818, opened 2024-04-27
Require required hash fields (security improvement)

mjurbanski-reef:
Previous implementation allowed complete bypass of `body_hash` checks. This PR fixes it.

This is more-or-less breaking change.

I say more-or-less, since, for most it should work without any trouble. Even if they don't update their side of the network. It will just mean they will be still vulnerable, but their nodes will still be able to communicate with the updated nodes.

# Fixed issues

* `required_hash_fields` no longer can be overridden (enabling hash bypass) by the remote node
* all `required_hash_fields` are now STRICTLY required, i.e. if subnet developer, [like in filetao](https://github.com/ifrit98/storage-subnet/blob/b92052e0c33aea0d3d04ce707e183964e442c1b5/storage/protocol.py#L169C40-L169C56), made a typo and had weaker hash because of it, now the code will immediately explode instead of issue going unnoticed for a long time

# Breaking changes

* If someone expected a `required_hash_fields` pydantic field in bittensor.Synapse, they will no longer find it - IDK how much code out there uses it directly. Seeing the amount of copy&pasting going on in bittensor community, probably some does.
* Unused `axon.required_hash_fields` mapping got removed (again, should not affect people in theory, but likely someone will be)

Otherwise, there is some support for legacy subclasses of Synapse, i.e. if they define `required_hash_fields` pydantic fields, we will use it as usual.

# Upgrade steps

* make sure you are not affected by breaking changes listed above
* make sure when redefining `required_hash_fields` as Synapse ClassVar, to list your field names in order of their definition, as otherwise you will get different hash values than before i.e. break compatibility across updated/unupdated nodes. Also if you had a typo in a field name, just don't add it if you don't want to break compatiblity with unupdated nodes.


thealligatorking:
@mjurbanski-reef thank you for the contribution. Since this introduces breaking changes we will be reviewing and getting back to you.

thealligatorking:
@mjurbanski-reef we plan on getting this into `7.0.0`

mjurbanski-reef:
@gus-opentensor did plan change? I see https://github.com/opentensor/bittensor/pull/1899 was opened without this

thealligatorking (approved):

bittensor/synapse.py
```diff
@@ -480,14 +484,7 @@ def set_name_type(cls, values) -> dict:
         repr=False,
     )
 
-    required_hash_fields: Optional[List[str]] = pydantic.Field(
```
opendansor: One thing to call out here is that with the Pydantic V2 update, this optional field will be required, unless explicitly set to None. 

https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields

mjurbanski-reef: As you can see this as a pydantic field is being replaced withit being plain class var (i.e. not something that remote node can influence) from now on.

But the fact remains, since there is also some legacy support in this PR, the pydantic v2 PR will have to also have to address it as well. Probably another reason to merge this sooner than later, before #1889 progresses further and has to fix the related parts of the code.

mjurbanski-reef:
@gus-opentensor @opendansor solved the conflicts caused by pydantic v2 PR, can we get this security fix merged so I don't have to keep redoing it?

thealligatorking (approved):

PR #1832, opened 2024-04-30
Add in check for minimum stake for unstaking

### LINKED ISSUE #1828: Match hotfix from subtensor to factor in threshold and 0 balance when unstaking
Subtensor hotfix: https://github.com/opentensor/subtensor/pull/383/

thewhaleking:
Fixes #1828

Adds a check to ensure that the unstaked balance is greater than or equal to the minimum, or 0.

### DIFF
--- a/bittensor/commands/unstake.py
+++ b/bittensor/commands/unstake.py
@@ -98,7 +98,9 @@ class UnStakeCommand:
                     try:
                         config.amount = float(amount)
                     except ValueError:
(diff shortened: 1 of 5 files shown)


thealligatorking:
@thewhaleking I know this is an ephemeral fix but could we please add a test.

thewhaleking:
> @thewhaleking I know this is an ephemeral fix but could we please add a test.

Yes. One moment

thealligatorking (approved):

PR #1835, opened 2024-05-01
Add version check caching, fix version comparison

olzhasar-reef:
This PR introduces the following changes:
1. Version checking cache:
Currently, version checking works by making an HTTP request to PyPI on every call. To avoid repetitive network requests, we implement a caching mechanism using a local file, ensuring that requests are only made once a day at most.
2. Improved version comparison:
The current version comparison logic relies on integer comparisons, which do not support letters and may lead to errors in cases like "6.0.15" vs "6.1.0". The new comparison mechanism is more robust and powered by the [version handling](https://packaging.pypa.io/en/latest/version.html) capabilities of the packaging library.
3. Fix for `version_checking` silent failure issue:
The CLI module includes logic for displaying an error message with a tip to disable version checking in case of a network request failure. However, this code is unreachable due to version checking failing silently.
This PR introduces a new `check_version` function with a custom exception. The old behavior is retained with a display of a deprecation warning.
The new function name should also be more descriptive to an end user.


thealligatorking:
@olzhasar-reef thank you for the contribution!

thealligatorking:
@ibraheem-opentensor do you mind testing this when you get a chance?

tests/unit_tests/utils/test_version.py
```diff
@@ -0,0 +1,168 @@
...
+    mock_get_version_from_pypi.assert_called_once()
+    assert version_file_path.exists()
+    assert version_file_path.read_text() == pypi_version
+
+
+@pytest.mark.parametrize("elapsed", [0, VERSION_CHECK_THRESHOLD - 5])
+def test_get_and_save_latest_version_file_fresh_check(
+    mock_get_version_from_pypi: MagicMock, version_file_path: Path, elapsed: int
+):
+    now = datetime.utcnow()
```
ibraheem-abe: Thank you for your contribution! 

`utcnow()` is depreciated ([link](https://docs.python.org/3/whatsnew/3.12.html#:~:text=datetime%3A%20datetime.datetime%E2%80%99s%20utcnow()))
Maybe we can use `datetime.now(timezone.utc)`

olzhasar-reef: Sure, fixed

ibraheem-abe:
@gus-opentensor  Lgtm!

thealligatorking (approved):

olzhasar-reef:
@gus-opentensor I've merged recent changes from the staging branch addressing conflicts.
Let me know if there's anything else preventing from merging this PR.

PR #1840, opened 2024-05-04
Add the command btcli root list_delegates_lite to handle the Delegate…

basfroman:
Add the command **btcli root list_delegates_lite** to handle the **DelegateInfoLite** object from the API method **delegateInfo_getDelegatesLite**.

### DIFF
--- a/bittensor/commands/delegates.py
+++ b/bittensor/commands/delegates.py
@@ -46,23 +46,42 @@ def _get_coldkey_wallets_for_path(path: str) -> List["bittensor.wallet"]:
 console = bittensor.__console__
 
 
-def show_delegates_lite(delegates_lite: List["bittensor.DelegateInfoLite"], width: Optional[int] = None):
+def show_delegates_lite(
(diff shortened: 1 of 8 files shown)


bittensor/commands/delegates.py
```diff
@@ -556,6 +722,16 @@ def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
...
+                prev_delegates = None
+
+        if prev_delegates is None:
+            bittensor.__console__.print(
+                ":warning: [yellow]Could not fetch delegates history[/yellow]"
+            )
+
         show_delegates(
             delegates,
             prev_delegates=None,
```
ibraheem-abe: Do we need to pass `prev_delegates` here since we are fetching it earlier

ibraheem-abe (approved):
LGTM 👍

bittensor/commands/delegates.py
basfroman: Nice catch! Thanks

PR #1848, opened 2024-05-07
Update README.md

larry-neil-1206:
Update README file description.


thewhaleking (changes requested):
Hey @larry-neil-1206 ,

Good catch, but there are still only really three ways to install it (despite the numbering). I'd suggest doing 3a and 3b to reflect that instead of this.

thealligatorking:
@larry-neil-1206 please cut off of staging.

PR #1854, opened 2024-05-08
Small fix with receiving delegates based on a 4-hour archive block

basfroman:


thewhaleking (approved):

basfroman:
For proper work with archive blocks, we need to set the fields subtensor.network and subtensor.chain_endpoint before subtensor initialization.

garrett-opentensor (approved):

PR #1863, opened 2024-05-10
Tests: Increased coverage of network.py

ibraheem-abe:
Increased coverage: 53% -> 88%


thealligatorking:
@ibraheem-opentensor black

PR #1871, opened 2024-05-13
Updated error message and a test typo.

### LINKED ISSUE #1867: Typo and bad error message
Image

This is the error given from a `scalecodec.exceptions.RemainingScaleBytesNotEmptyException` in `subtensor.get_balance`, but I don't think this is necessarily the case. Regardless, the typo needs fixed.

thewhaleking:
Fixes #1867

Fixes an error message that claimed a corrupted message from the server was a legacy wallet. Also fixes a typo in a test.

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -4415,7 +4415,7 @@ class subtensor:
             result = make_substrate_call_with_retry()
         except scalecodec.exceptions.RemainingScaleBytesNotEmptyException:
             bittensor.logging.error(
-                "Your wallet it legacy formatted, you need to run btcli stake --ammount 0 to reformat it."
+                "Received a corrupted message. This likely points to an error with the network or subnet."
             )
             return Balance(1000)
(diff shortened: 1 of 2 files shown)


bittensor/subtensor.py
```diff
@@ -4415,7 +4415,7 @@ def make_substrate_call_with_retry():
             result = make_substrate_call_with_retry()
         except scalecodec.exceptions.RemainingScaleBytesNotEmptyException:
             bittensor.logging.error(
-                "Your wallet it legacy formatted, you need to run btcli stake --ammount 0 to reformat it."
+                "Received a corrupted message. This likely points to an error with the network or subnet."
```
garrett-opentensor, line 4418: Maybe we can add the legacy wallet info to your new messages as well.  Like "...likely points to an error with the network/subnet, or may be due to usage of a legacy-formatted wallet."    Open to ideas here. Not sure how many legacy wallets there are floating out there

thealligatorking (approved):

PR #1851, opened 2024-05-08
feat: commit weights

distributedstatemachine:
# Description

This PR closes https://github.com/opentensor/subtensor/issues/389

It introduces a commit-reveal scheme as described here

PR #1875, opened 2024-05-14
Feat/commit reveal weights

distributedstatemachine:
# Description

This PR closes https://github.com/opentensor/subtensor/issues/389

It introduces a commit-reveal scheme as described here

Supersedes  #1851

### DIFF
--- a/bittensor/commands/weights.py
+++ b/bittensor/commands/weights.py
@@ -1,5 +1,4 @@
 import re
-import typing
 import argparse
 import numpy as np
 import bittensor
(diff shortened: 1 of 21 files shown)


bittensor/commands/weights.py
```diff
@@ -0,0 +1,357 @@
...
+
+        if not cli.config.is_set("uids"):
+            cli.config.uids = Prompt.ask(f"Enter UIDs (comma-separated)")
+
+        if not cli.config.is_set("weights"):
+            cli.config.weights = Prompt.ask(f"Enter weights (comma-separated)")
+
+        # Parse from string
+        netuid = cli.config.netuid
+        uids = torch.tensor(
```
thealligatorking: We're trying to deprecate use of torch if you can, use numpy:
```
uids = cli.config.uids  
uid_list = list(map(int, re.split(r"[ ,]+", uids)))
np_array = np.array(uid_list, dtype=np.int64)
```

bittensor/commands/weights.py
```diff
@@ -0,0 +1,357 @@
...
```
thealligatorking: Same, torch -> numpy if you can

thealligatorking (approved):

PR #1886, opened 2024-05-15
fix: deprecated usage of `Balances::transfer` method

orriin:
`transfer` is removed and replaced with `transfer_allow_death` https://github.com/polkadot-js/apps/issues/9854


camfairchild (approved):
LGTM, but perhaps we should be using keep alive

orriin:
That would prevent sending the entire account bal. If you're sure, can change it :)

thealligatorking (approved):

PR #1888, opened 2024-05-15
fix: Incompatible type hint

thealligatorking:
Fixes to support https://docs.python.org/3.8/library/typing.html#typing.List

### DIFF
--- a/bittensor/btlogging/loggingmachine.py
+++ b/bittensor/btlogging/loggingmachine.py
@@ -5,7 +5,7 @@ import atexit
 import argparse
 import multiprocessing as mp
 import logging as stdlogging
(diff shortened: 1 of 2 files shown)


ibraheem-abe (approved):

mjurbanski-reef:
This maybe a wasted effort, since Python 3.9 has https://peps.python.org/pep-0585/ i.e. list[...] is encouraged over List[...] in all python version after 3.8, and 3.8 hits End of life soon https://devguide.python.org/versions/ . 

Since bittensor projects are very much focused on constant development, for bittensor community it maybe better to drop python 3.8 already instead of investing development time in supporting something that is being abandoned by the python community at large.

PR #1889, opened 2024-05-15
Update: Pydantic V2

### LINKED ISSUE #1898: Pydantic V2 integration into Synapse
We are updating a dependency called Pydantic. This library itself got updated from v1 to v2, and we are brining out version of this library to v2. 

There are breaking changes on the Pydantic side between v1 and v2. We are to update this library without the loss of functionality in the Synapse class.

opendansor:
Update Pydantic from V1 to V2. Some of the changes between V1 and V2 have breaking impacts, and our tests will need to be adjusted to reflect those changes.


requirements/prod.txt
```diff
@@ -19,7 +19,7 @@ packaging
 pycryptodome>=3.18.0,<4.0.0
 pyyaml
 password_strength
-pydantic!=1.8,!=1.8.1,<2.0.0,>=1.7.4
+pydantic==2.7.1
```
mjurbanski-reef:
```suggestion
pydantic>=2.7.1,<3
```
in libraries, such like `bittensor` package, it is better not to be very strict with dependency versions when as it very easy to get conflicts in the application. pydantic is using semver, restricting just major version will be fine.

This is extra problematic if you get a security update in one of dependencies, but some library opted to do a strict `==` and you have to do some hacks just to update minor/bugfix version

mjurbanski-reef (commented):
You may want to use https://docs.pydantic.dev/latest/migration/#code-transformation-tool , while it won't solve everything it will be a good starting point.

requirements/prod.txt
thewhaleking: I agree with @mjurbanski-reef here. Better to not restrict specific minor version (beyond 2.7 which we're currently at). I might even suggest doing `>=2.0,<3`, as 2.7.1 was only released a few weeks ago, and all of the changes you've made are only specific to 2.0+

thewhaleking (changes requested):

requirements/prod.txt
opendansor: Agreed. I will increase the scope for the versioning.

opendansor: @thewhaleking I believe you are right, 

https://docs.pydantic.dev/latest/version-policy/

I will do the update.

thewhaleking:
[test_fail.log](https://github.com/opentensor/bittensor/files/15389772/test_fail.log)

Getting a number of errors when I run the tests locally. Not sure why.

RomanCh-OT:
After all the discussions, there was a clear approval.

Not a blocker, just a nitpick: I would advise using a type annotation without quotes, as we decided in [RFC 003: Bittensor Coding Standards](https://docs.google.com/document/d/1ERaRsZY-SFVJY0a179_rX2-8Ff-aFQWTUNn-gXj3UiI/edit#heading=h.dj0jbjcnkmqg)

bittensor/synapse.py
```diff
@@ -357,8 +365,7 @@ class Synapse(pydantic.BaseModel):
     standardized communication in a decentralized environment.
     """
 
-    class Config:
-        validate_assignment = True
+    model_config = ConfigDict(validate_assignment=True)
 
     def deserialize(self) -> "Synapse":
```
ibraheem-abe, line 370: I think we can also remove `parse_raw` in the `deserialize` docstring as it is depreciated. 

Same case with `allow_mutation` which can be removed from:
- `process_server_response` comment (Dendrite.py - line number 703)
- `class Topk( bittensor.Synapse)` - (Readme.md - line number 274)

thewhaleking (commented):
Hey Dan, all looks good, but I did run `bump-pydantic --diff bittensor/` to check, and I got this:

```
[19:40:24] Start bump-pydantic.                                                                                                                                       main.py:61
           Found 66 files to process.                                                                                                                                 main.py:78
--- bittensor/tensor.py
+++ bittensor/tensor.py
@@ -171,7 +171,7 @@
 
     # Represents the tensor buffer data.
     buffer: Optional = Field(
-        title="buffer",
+        None, title="buffer",
         description="Tensor buffer data. This field stores the serialized representation of the tensor data.",
         examples=["0x321e13edqwds231231231232131"],
         frozen=True,
[19:41:04] Found 1 errors. Please check the log.txt file.  
```

and the log.txt file shows:
```
==SUPRESS==
^A syntax error happened on bittensor/utils/__init__.py. This file cannot be formatted.
Check https://github.com/pydantic/bump-pydantic/issues/124 for more information.
Syntax Error @ 1:1.
parser error: error at 1:2: expected one of (, *, +, -, ..., AWAIT, EOF, False, NAME, NUMBER, None, True, [, break, continue, lambda, match, not, pass, ~
```

Just check into those.

bittensor/synapse.py
opendansor: I have updated dendrite.py and readme. 
Also updated `parse_raw` in `deserialize` to `model_validate_json`.

opendansor:
@thewhaleking I have addressed the Pydantic bump change.

thewhaleking (approved):

bittensor/synapse.py
opendansor: https://github.com/orgs/opentensor/projects/28/views/1?pane=issue&itemId=63805700

PR #1903, opened 2024-05-17
Add setting delegate take

gztensor:
This PR adds setting delegate take functionality. 

Command: 

```bash
btcli root set_take
```


bittensor/subtensor.py
```diff
@@ -4382,6 +4449,82 @@ def make_substrate_call_with_retry():
 
         return make_substrate_call_with_retry()
 
+    def _do_increase_take(
+        self,
+        wallet: "bittensor.wallet",
+        hotkey_ss58: str,
+        take: int,
```
thewhaleking, line 4456: Is `take` supposed to be an int or a float? It's defined here as `int`, but it's referenced elsewhere (like `extrinsics/delegation.py` as float.

distributedstatemachine: This is correct. The user passes a  a float via the cli,  is is converted into a [u16 int](https://github.com/opentensor/bittensor/pull/1903/files/f69874d0d9484dbdd5164b7e91ba3a6409efdbe7#diff-bceb1ae9ef839729fe96077619a03ec8768c3c834c145be4fc700392dd289f7eR598). So i think its right. The subtensor function should take a float , while the extrinsics should accept ints

thealligatorking (approved):

Issue #1897, opened 2024-05-17
Restore Torch support for legacy functions.


thewhaleking:
Add in backwards compatibility for the currently existing functionality with Torch that was removed in:

https://github.com/opentensor/bittensor/pull/1786

PR #1904, opened 2024-05-20
Add back compatibility with torch

thewhaleking:
Adds back compatibility with torch, invoked by passing the `USE_TORCH` env var.

Fixes [#1897](https://github.com/opentensor/bittensor/issues/1897)


bittensor/chain_data.py
```diff
@@ -25,7 +25,7 @@
 from scalecodec.type_registry import load_type_registry_preset
 from scalecodec.utils.ss58 import ss58_encode
 
-from .utils import networking as net, U16_MAX, U16_NORMALIZED_FLOAT
+from .utils import networking as net, U16_MAX, U16_NORMALIZED_FLOAT, maybe_get_torch
```
thealligatorking: I know you didn't write or name the func `maybe_get_torch` but it feels like a code smell. I like how you use the env var to check for torch use. Could we update the name of `maybe_get_torch` in this PR as well?

bittensor/chain_data.py
```diff
@@ -264,15 +264,43 @@ def from_neuron_info(cls, neuron_info: dict) -> "AxonInfo":
             coldkey=neuron_info["coldkey"],
         )
 
-    def to_parameter_dict(self) -> dict[str, Union[int, str]]:
-        r"""Returns a dict of the subnet info."""
+    def _to_parameter_dict_torch(self) -> "torch.nn.ParameterDict":
```
thealligatorking: You could also do something like this:

```python

def _to_parameter_dict(self, output_type: str) -> Union["torch.nn.ParameterDict", Dict[str, Union[int, str]]]:
    """Returns a dictionary of the subnet info in the specified format."""
    if output_type == 'torch':
        return torch.nn.ParameterDict(self.__dict__)
    elif output_type == 'numpy':
        return self.__dict__
    else:
        raise ValueError(f"Unsupported output_type: {output_type}. Supported types are 'torch' and 'numpy'.")
```

bittensor/chain_data.py
thewhaleking: I could. I did think about that, and was really looking for someone else to think the same. I'll update it.

bittensor/chain_data.py
thewhaleking: This is probably a better approach. Thanks.

thealligatorking:
Looks great! 

Suggestion: add a file `.example.env` with `USE_TORCH=` and add a comment to indicate what have it set to `0` or `1` does.

bittensor/utils/__init__.py
```diff
@@ -40,35 +40,67 @@ def ss58_to_vec_u8(ss58_address: str) -> List[int]:
...
+        indices = np.argpartition(permuted_values, -k, axis=axis)[-k:]
+        if not sorted:
+            indices = np.sort(indices, axis=axis)
+        if not largest:
+            indices = indices[::-1]
+        topk = np.take(permuted_values, indices, axis=axis)
+        return topk, permutation[indices]
+
+
 def unbiased_topk(
```
basfroman, line 98: add function docstring pls
the same below

bittensor/metagraph.py
```diff
@@ -913,4 +1182,5 @@ def load_from_path(self, dir_path: str) -> "metagraph":
         return self
 
 
-metagraph = Metagraph
+print("USE_TORCH", use_torch())
```
basfroman: why don't use `btlogging` instead of `print`?

bittensor/metagraph.py
```diff
@@ -865,6 +848,294 @@ def load_from_path(self, dir_path: str) -> "metagraph":
             contain valid data for the metagraph. It is essential to ensure that the directory path and the
             state files within it are accurate and consistent with the expected metagraph structure.
         """
+        pass
+
+
+BaseClass = torch.nn.Module if use_torch() else object
```
basfroman: the same as above

bittensor/dendrite.py
```diff
@@ -808,3 +811,24 @@ def __del__(self):
             del dendrite  # This will implicitly invoke the __del__ method and close the session.
         """
         self.close_session()
+
+
+# For back-compatibility with torch
+BaseModel = torch.nn.Module if use_torch() else object
```
basfroman: pls add type annotation to BaseModel

bittensor/commands/root.py
```diff
@@ -520,12 +528,21 @@ def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
...
+        netuids = (
+            torch.tensor(matched_netuids, dtype=torch.long)
+            if os.environ.get("USE_TORCH")
+            else np.array(matched_netuids, dtype=np.int64)
         )
-        weights = np.array(
-            list(map(float, re.split(r"[ ,]+", cli.config.weights))),
-            dtype=np.float32,
+
+        matched_weights = list(map(float, re.split(r"[ ,]+", cli.config.weights)))
```
basfroman: just a nitpick: I would prefer to use a list comprehension for its readability and style, which aligns with Python's recommendations (PEP 8). In terms of performance, both options are practically the same.
`matched_weights = [float(weight) for weight in re.split(r"[ ,]+", cli.config.weights)]`

bittensor/metagraph.py
thewhaleking: This shouldn't be there at all. It was for testing. Good catch!

basfroman (approved):
LGTM

thealligatorking (approved):

PR #1911, opened 2024-05-21
Part 1 for refactoring bittensor/subtensor.py

### LINKED ISSUE #1908: bittensor.subtensor module refactoring

As a part of https://github.com/opentensor/bittensor/issues/1882

- [x] Style refactoring
- [x] Сode refactoring

- [ ] ~~Implement the ability to use functions in an asynchronous implementation.~~ Decided to do this in different PR

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

The current part affects everything up to the subtensor.get_transfer_fee() method

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -16,26 +16,30 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-import functools
-import os
-import copy
-import time
-import logging
-import numpy as np
-from numpy.typing import NDArray
+"""
+The ``bittensor.subtensor`` module in Bittensor serves as a crucial interface for interacting with the Bittensor
+blockchain, facilitating a range of operations essential for the decentralized machine learning network.
+"""
 import argparse
-import bittensor
-import scalecodec
-
-from retry import retry
+import copy
+import functools
+import socket
+import time
 from typing import List, Dict, Union, Optional, Tuple, TypedDict, Any, TypeVar
-from substrateinterface.base import QueryMapResult, SubstrateInterface, ExtrinsicReceipt
-from substrateinterface.exceptions import SubstrateRequestException
+
+import numpy as np
+import scalecodec
+from numpy.typing import NDArray
+from retry import retry
 from scalecodec.base import RuntimeConfiguration
+from scalecodec.exceptions import RemainingScaleBytesNotEmptyException
 from scalecodec.type_registry import load_type_registry_preset
 from scalecodec.types import GenericCall
+from substrateinterface.base import QueryMapResult, SubstrateInterface, ExtrinsicReceipt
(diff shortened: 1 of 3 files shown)


bittensor/subtensor.py
```diff
@@ -261,22 +262,31 @@ def __init__(
...
+            _logger.error(
+                f"Could not connect to %s network with %s chain endpoint. Exiting...",
+                self.network,
+                self.chain_endpoint,
             )
-            bittensor.logging.info(
-                f"You can check if you have connectivity by runing this command: nc -vz localhost {self.chain_endpoint.split(':')[2]}"
+            _logger.info(
+                "You can check if you have connectivity by running this command: nc -vz localhost %s",
+                self.chain_endpoint.split(":")[2],
```
thewhaleking: Why the change from f-strings to the older-style %s here?

bittensor/subtensor.py
```diff
@@ -496,7 +534,7 @@ def nominate(
 
     def delegate(
         self,
-        wallet: "bittensor.wallet",
+        wallet: bittensor.wallet,
```
thewhaleking, line 535: This probably doesn't matter here, but don't we tend to use quotes around bittensor typing to avoid circular imports?

bittensor/utils/balance.py
```diff
@@ -258,9 +258,9 @@ def from_float(amount: float):
...
 
     @staticmethod
-    def from_tao(amount: float):
+    def from_tao(amount: Union[int, float]):
         """
-        Given tao (float), return Balance object with rao(int) and tao(float), where rao = int(tao*pow(10,9))
+        Given tao (int | float), return Balance object with rao(int) and tao(float), where rao = int(tao*pow(10,9))
 
         Args:
             amount: The amount in tao.
```
thewhaleking: While `from_tao` and `from_rao` accept ints or floats, I think the intention of keeping them `tao: float` and `rao: int` is a good idea.

thewhaleking (commented):
See line comments. Not necessarily requesting changes, but want to discuss these couple items first.

bittensor/subtensor.py
basfroman: As we decided in [RFC 003: Bittensor Coding Standards](https://docs.google.com/document/d/1ERaRsZY-SFVJY0a179_rX2-8Ff-aFQWTUNn-gXj3UiI/edit#heading=h.cec1zrvrfmzv) -> Approach -> Typing :: `Prefer imported types for annotations; if not possible, use type names in quotes (for example to avoid circular imports).`
In this case there are no circular imports.

bittensor/utils/balance.py
basfroman: In general, I added this because in different parts of the code for both methods we pass both  types (int and float).
Let's return it as it was and more strictly control which types we will pass.

bittensor/subtensor.py
basfroman: Switch back

thewhaleking (approved):

garrett-opentensor (approved):

PR #1915, opened 2024-05-22
ci: e2e tests

orriin:
Runs the e2e tests in ci to catch regressions like #1914 before they're merged

### DIFF
--- a/.github/workflows/e2e-subtensor-tests.yaml
+++ b/.github/workflows/e2e-subtensor-tests.yaml
@@ -30,7 +30,7 @@ env:
 
 jobs:
   run:
-    runs-on: ubuntu-22.04
+    runs-on: SubtensorCI
     strategy:
       matrix:
         rust-branch:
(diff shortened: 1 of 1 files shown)


thealligatorking:
@orriin taking a look!

.github/workflows/e2e-subtensor-tests.yaml
```diff
@@ -0,0 +1,85 @@
...
+  CARGO_TERM_COLOR: always
+  VERBOSE: ${{ github.events.input.verbose }}
+
+jobs:
+  run:
+    runs-on: SubtensorCI
+    strategy:
+      matrix:
+        rust-branch:
+          - nightly-2024-03-05
```
opendansor, line 37: Will this line need to be updated with new releases?

orriin: If the rust version changes, yes

opendansor (approved):
LGTM, one call out is to either create a PR when/if rust version updates or let us know so we can update the rust version ourselves.

PR #1916, opened 2024-05-22
fix error handling in blacklist processing

cxmplex:
Currently, if attackers are able to cause an error in the blacklist function, as they can with the current subnet example code, this error will propagate downstream to here.

Please see my associated PR in the subnet example repository that addresses the default code having an exploitable line to force an error:

https://github.com/opentensor/bittensor-subnet-template/pull/88


cxmplex:
**Spoke with upstream devs about this as its more appropriate to fix it on SN side as it doesnt appear any function in part of the axon operation has its exceptions caught**. this likely means the o**nus is on the sn dev** to catch them and the existing pattern **should not** be modified. As a result, PR closed.

PR #1917, opened 2024-05-22
fix PoW and other functions not working with USE_TORCH=0 despite torch being available

mjurbanski-reef:
### Bug

This fixes problem that after #1904 uses wanting PoW and other torch-using functionality were forced to use USE_TORCH=1 which also forced the legacy torch-based interface.

### Description of the Change

* users can now use torch-requiring code despite using new interface - this should prevent users being even more reluctant to upgrade just because they use few functions that require torch
* the default value in example.env is now `USE_TORCH=0`, so users don't accidentally run legacy interface - we want old users to move to new one as soon as it is viable AND none of the new users should accidentally run legacy interface
* fixed recursion error in lazy torch loader class which happend when requesting nonexistent attribute 
* refactored the lazy torch loader and added tests for it
* added `legacy_torch_api_compat` for an easier way of maintaining two interfaces, but only used it in limited fashion for now

### Alternate Designs

as written above, no 

### Possible Drawbacks

No real drawbacks that Im aware of.

### Verification Process

Unit tests added and manual tests.

### Release Notes

None, this fixes regression introduced by #1904 that was yet to be released.

### DIFF
--- a/bittensor/utils/registration.py
+++ b/bittensor/utils/registration.py
@@ -30,10 +30,12 @@ def use_torch() -> bool:
 
 
 def legacy_torch_api_compat(func):
-    """Decorator to convert numpy arrays to torch tensors before passing them to the function.
+    """
+    Convert function operating on numpy Input&Output to legacy torch Input&Output API if `use_torch()` is True.
+
     Args:
         func (function):
-            Function to be decorated.
+            Function with numpy Input/Output to be decorated.
     Returns:
         decorated (function):
             Decorated function.
--- a/bittensor/utils/registration.py
+++ b/bittensor/utils/registration.py
@@ -1,16 +1,20 @@
 import binascii
+import functools
 import hashlib
 import math
 import multiprocessing
 import os
 import random
 import time
+import typing
 from dataclasses import dataclass
 from datetime import timedelta
 from queue import Empty, Full
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import backoff
+import numpy
+
 import bittensor
 from Crypto.Hash import keccak
 from rich import console as rich_console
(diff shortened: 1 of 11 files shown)


bittensor/utils/weight_utils.py
```diff
@@ -22,12 +22,13 @@
...
 from numpy.typing import NDArray
 from typing import Tuple, List, Union
-from bittensor.utils.registration import torch, use_torch
+from bittensor.utils.registration import torch, use_torch, legacy_torch_api_compat
 
 U32_MAX = 4294967295
 U16_MAX = 65535
 
 
+@legacy_torch_api_compat
```
mjurbanski-reef, line 31: IMO this approach of just having one implementation of a function and then translating only input and output will be much easier to maintain, but I leave it to others if they want to use it in more places or if they want to keep using `if use_torch()` all over the place

mjurbanski-reef:
@thewhaleking as author of #1904 you maybe interested in this PR 

@gus-opentensor this probably should also land in v7, as it will make users less reluctant to switch to numpy interface

bittensor/utils/registration.py
```diff
@@ -19,62 +23,77 @@
...
 def use_torch() -> bool:
+    """Force the use of torch over numpy for certain operations."""
     return True if os.getenv("USE_TORCH") == "1" else False
 
 
-class Torch:
-    def __init__(self):
-        self._transformed = False
+def legacy_torch_api_compat(func):
+    """Decorator to convert numpy arrays to torch tensors before passing them to the function.
```
thewhaleking: I really like this approach. I would recommend updating this line, as it reads as the opposite of what it's saying. Other than that, this seems great.

thewhaleking (commented):
This overall looks good to me.

thewhaleking (approved):

PR #1919, opened 2024-05-22
thewhaleking/further-legacy-torch-improvements

thewhaleking:
Further implements the improvements of #1917

### DIFF
--- a/bittensor/commands/root.py
+++ b/bittensor/commands/root.py
@@ -24,7 +24,6 @@ from typing import List, Optional, Dict
 from rich.prompt import Prompt
 from rich.table import Table
 from .utils import get_delegates_details, DelegatesDetails
-from bittensor.utils.registration import torch, use_torch
 
 from . import defaults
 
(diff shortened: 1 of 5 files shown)


bittensor/commands/root.py
```diff
@@ -405,7 +400,7 @@ def run(cli: "bittensor.cli"):
     @staticmethod
     def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
         wallet = bittensor.wallet(config=cli.config)
-        subnets: List[bittensor.SubnetInfo] = subtensor.get_all_subnets_info()
+        # subnets: List[bittensor.SubnetInfo] = subtensor.get_all_subnets_info()
```
opendansor: Why comment this out instead of removing it?

bittensor/commands/root.py
```diff
@@ -283,7 +282,7 @@ def run(cli: "bittensor.cli"):
     def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
         r"""Set weights for root network."""
         wallet = bittensor.wallet(config=cli.config)
-        subnets: List[bittensor.SubnetInfo] = subtensor.get_all_subnets_info()
+        # subnets: List[bittensor.SubnetInfo] = subtensor.get_all_subnets_info()
```
basfroman: let's delete this. I checked, it just makes unnecessary code and doesn't have any effect.
the same below

basfroman (approved):
not blocking comments
LGTM

bittensor/commands/root.py
thewhaleking: Good call. Forgot to fully remove them after testing.

PR #1925, opened 2024-05-22
Remove add weighted stakes

gztensor:
Remove stake weight command which effectively removes use of add weighted stake extrinsic. Matches this PR for Subtensor: https://github.com/opentensor/subtensor/pull/456


distributedstatemachine:
lint @gztensor

PR #1934, opened 2024-05-24
Part 5 for refactoring bittensor/subtensor.py

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

This is part 5: current part affects everything in this document except of renaming class `subtensor` according to the CamelCase rule.

Test coverage improved.

Part 1 https://github.com/opentensor/bittensor/pull/1911
Part 2 https://github.com/opentensor/bittensor/pull/1913
Part 3 https://github.com/opentensor/bittensor/pull/1923
Part 4 https://github.com/opentensor/bittensor/pull/1931

PR #1931, opened 2024-05-23
Part 4 for refactoring bittensor/subtensor.py

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

This is part 4: current part affects everything up to the subtensor().tx_rate_limit() method.
Test coverage improved.

Part 1 https://github.com/opentensor/bittensor/pull/1911
Part 2 https://github.com/opentensor/bittensor/pull/1913
Part 3 https://github.com/opentensor/bittensor/pull/1923

PR #1923, opened 2024-05-22
Part 3 for refactoring bittensor/subtensor.py

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

This is part 2: current part affects everything up to the subtensor().tempo() method.
The "Hyper parameter calls" section was refactored by DRY.

Part 1 https://github.com/opentensor/bittensor/pull/1911
Part 2 https://github.com/opentensor/bittensor/pull/1913

PR #1913, opened 2024-05-21
refactoring/roman/subtensor-part-2

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

This is part 2: current part affects everything up to the subtensor().validator_logits_divergence() method.

Part 1 https://github.com/opentensor/bittensor/pull/1911

PR #1935, opened 2024-05-24
Part 6 for refactoring bittensor/subtensor.py - class CamelCase

### LINKED ISSUE #1908: bittensor.subtensor module refactoring

basfroman:
Since the refactoring of this module would turn into a large project, it was decided to divide it into several stages.

This is part 6: Renaming class 1 to comply with PEP8 and avoid namespace conflict. Refactor all references to the changed name.

Part 1 https://github.com/opentensor/bittensor/pull/1911
Part 2 https://github.com/opentensor/bittensor/pull/1913
Part 3 https://github.com/opentensor/bittensor/pull/1923
Part 4 https://github.com/opentensor/bittensor/pull/1931
Part 5 https://github.com/opentensor/bittensor/pull/1934

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -93,7 +93,6 @@ __networks__ = ["local", "finney", "test", "archive"]
 __finney_entrypoint__ = "wss://entrypoint-finney.opentensor.ai:443"
 
 __finney_test_entrypoint__ = "wss://test.finney.opentensor.ai:443/"
-__finney_test_entrypoint__ = "wss://test.finney.opentensor.ai:443/"
 
 __archive_entrypoint__ = "wss://archive.chain.opentensor.ai:443/"
 
@@ -234,9 +233,7 @@ from .errors import (
     UnstakeError,
 )
 
-from substrateinterface import Keypair
-
-_ = Keypair
+from substrateinterface import Keypair  # noqa: F401
 from .config import InvalidConfigFile, DefaultConfig, config, T
 from .keyfile import (
     serialized_keypair_to_keyfile_data,
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -79,7 +79,7 @@ __blocktime__ = 12
 # Pip address for versioning
 __pipaddress__ = "https://pypi.org/pypi/bittensor/json"
 
-# Raw github url for delegates registry file
+# Raw GitHub url for delegates registry file
 __delegates_details_url__: str = "https://raw.githubusercontent.com/opentensor/bittensor-delegates/main/public/delegates.json"
 
 # Substrate ss58_format
@@ -92,6 +92,7 @@ __networks__ = ["local", "finney", "test", "archive"]
 
 __finney_entrypoint__ = "wss://entrypoint-finney.opentensor.ai:443"
 
+__finney_test_entrypoint__ = "wss://test.finney.opentensor.ai:443/"
 __finney_test_entrypoint__ = "wss://test.finney.opentensor.ai:443/"
 
 __archive_entrypoint__ = "wss://archive.chain.opentensor.ai:443/"
@@ -106,7 +107,7 @@ __tao_symbol__: str = chr(0x03C4)
 __rao_symbol__: str = chr(0x03C1)
 
 # Block Explorers map network to explorer url
-## Must all be polkadotjs explorer urls
+# Must all be polkadotjs explorer urls
 __network_explorer_map__ = {
     "opentensor": {
         "local": "https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fentrypoint-finney.opentensor.ai%3A443#/explorer",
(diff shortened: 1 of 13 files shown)


bittensor/__init__.py
```diff
@@ -233,7 +234,9 @@ def debug(on: bool = True):
     UnstakeError,
 )
 
-from substrateinterface import Keypair as Keypair
+from substrateinterface import Keypair
+
+_ = Keypair
```
unconst: can you explain this design?

bittensor/subtensor.py
```diff
@@ -3351,62 +3349,150 @@ def tempo(self, netuid: int, block: Optional[int] = None) -> Optional[int]:
...
+            None
+            if _result is None or not hasattr(_result, "value")
+            else Balance.from_rao(_result.value)
+        )
 
     def get_total_stake_for_coldkey(
         self, ss58_address: str, block: Optional[int] = None
     ) -> Optional["Balance"]:
-        """Returns the total stake held on a coldkey across all hotkeys including delegates"""
+        """
```
unconst: what is the reasoning behind reducing these comments and not having args and return vars?

bittensor/__init__.py
```diff
@@ -287,8 +290,14 @@ def debug(on: bool = True):
...
 )
 
+# Allows avoiding name spacing conflicts and continue access to the `subtensor` module with `subtensor_module` name
 from . import subtensor as subtensor_module
-from .subtensor import subtensor as subtensor
+
+# Double import allows using class `Subtensor` by referencing `bittensor.Subtensor` and `bittensor.subtensor`.
+# This will be available for a while until we remove reference `bittensor.subtensor`
+from .subtensor import Subtensor
+from .subtensor import Subtensor as subtensor
```
unconst, line 296: We shouldn't remove the bt.subtensor factory design.

bittensor/__init__.py
basfroman: 1. We imported `Keypair` as an alias with the same name. I just shortened the code so as not to confuse the programmers.
2. Some IDEs have the ability to optimize imports. But if the imported object is not used within the module (like `Keypair` in __init_.py) then it may remove the import, which will break the package work but not module (but this will not be noticeable immediately). I used `_=Keypair` to avoid this. The variable `_` will be instantly removed by the Py garbage collector, but the import will be preserved.

bittensor/subtensor.py
basfroman: Below this docstring describes all the arguments and the return value. 
Or did I misunderstand your question?

bittensor/__init__.py
basfroman: This is still under discussion. Both options are available now, as we discussed in the chat.

bittensor/__init__.py
mjurbanski-reef: won't the same IDE (&flake) complain about it? i.e. about unused `_` variable so you will still get unhappy linter despite the hack

The generally used solution for these (respected by flake, ruff, PyCharm, ...) is to add `  # noqa: F401` at the end of the line doing import

mjurbanski-reef:
as a Python Developer and bittensor user, I must say that it may seem small, but moving towards getting rid of confusing classes masking over the modules may seem silly, but is an awesome, quality of life change. Both I and my IDE are getting confused with `bt.subtensor` :)

bittensor/__init__.py
basfroman: Oh, I forgot about `# noqa` indicator. Thanks for this!
Also, "_" variables don't make linters angry, that's their point. At least never for me ☺️

basfroman:
> as a Python Developer and bittensor user, I must say that it may seem small, but moving towards getting rid of confusing classes masking over the modules may seem silly, but is an awesome, quality of life change. Both I and my IDE are getting confused with `bt.subtensor` :)

Thank you for such warm feedback! 

We really try to do everything so that the community gets the best experience both as developers and as users who are not related to programming. The more and more detailed community's requests, feedback, and suggestions we receive, the better the Bittensor becomes!

bittensor/utils/wallet_utils.py
```diff
@@ -8,7 +8,7 @@
 # the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
 # and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
-# The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+# The above copyright notice and this permission notice shall be included in all copies or large portions of
```
thewhaleking: You can't change wording on the license.

thewhaleking (changes requested):
I like this, but revert the license wording change. I do like adding some kind of deprecation warning for the `subtensor` that we discussed, but it's not immediately necessary.

bittensor/utils/wallet_utils.py
basfroman: Done

thewhaleking (approved):

ibraheem-abe (approved):

PR #1949, opened 2024-05-29
support Wallet names with hyphens when passing password through ENV vars

mjurbanski-reef:
### Description of the Change

Support Wallet names with hyphens when passing password through ENV vars

### Alternate Designs

Another solutions would be to:
* Explicitly discourage (disallow?) creating wallet names with hyphens - naming is hard problem, so it would likely rattle some people

### Possible Drawbacks

A possible drawback is if someone had multiple wallets which only difference was that one had `-` instead of `_`, but that seems VERY unlikely.

### Verification Process

Added unittest case + manual testing.

### Release Notes

- `BT_COLD_PW_*` env vars now can be used to provide password for wallet names with hyphens (use `_` in place of `-`)


mjurbanski-reef:
anything I can do to move this forward?

thealligatorking (approved):

mjurbanski-reef:
@gus-opentensor anything stopping this from being merged?

thealligatorking:
@mjurbanski-reef tests, I'll merge once they pass.

mjurbanski-reef:
@gus-opentensor the tests are failing since the exactly tests are failing in staging branch that I just merged into this

PR #1973, opened 2024-06-04
ruff autolinter

mjurbanski-reef:
Continues from #1968 

It already successfully detected couple of redefinitions, which in these cases meant couple of tests were in fact never run and it was overlooked before now despite them being worked on fairly recently.

Other than that I think sorting of imports is a nice change, but had to limit it due to the `bittensor.synapse` class/module hack (hopefully it can be removed in future).


mjurbanski-reef:
integration tests don't pass; will have to look at it tomorrow since they seem harder to debug

PR #1974, opened 2024-06-04
deprecate nest_asyncio use

### LINKED ISSUE #1839: Remove nest_asyncio from bittensor to allow uvloop support

Uvloop, which provides supperior speed, does not allow loop nesting.

It is also the case that uvloop is pulled in by popular packages, which [forces some subnets develop hacks to combat this](https://github.com/synapsec-ai/llm-defender-subnet/blob/6c37925c4f34a298607c97dfceebcc01fb74d562/scripts/run_neuron.sh#L140-L146).

And perhaps more importantly, https://github.com/erdewit/nest_asyncio seems to have been abandodend 

Remove nest_asyncio, and let bittensor users decide which asyncio loop they want to run. Perhaps even suggest (not mandate) running uvloop, since it consistently shows better results in benchmarks than CPython asyncio stdlib loop.

Seems like there was some attempt of this in the past https://github.com/opentensor/bittensor/pull/1501 for some reason (?) 

_No response_

### Additional context

_No response_

mjurbanski-reef:
* Deprecate nest_asyncio use allowing for alternative asyncio loops (e.g. uvloop) to be used when `NEST_ASYNCIO=0` env variable is set. In next major  release of `bittensor` package the `nest_asyncio` dependency will be removed and will have to be installed explicitly instead. (Resolves #1839)

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -16,15 +16,28 @@
 # THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
+import os
+import warnings
 
 from rich.console import Console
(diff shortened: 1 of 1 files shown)


thealligatorking (approved):

thewhaleking (approved):

thewhaleking:
@mjurbanski-reef just update the branch, and I'll merge this in. Thanks.

PR #1976, opened 2024-06-04
Feat: Nonce implementation improved

ibraheem-abe:
Updates nonce's implementation:
- After bittensor 7.2, we nonces will converted to timestamps instead of monotonic numbers
- The code will check the synapse bittensor version and branch off in verifying the correct nonce structure accordingly

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -24,7 +24,6 @@ import asyncio
 import contextlib
 import copy
 import inspect
-import ntplib
 import json
 import os
 import threading
@@ -57,6 +56,7 @@ from bittensor.errors import (
     SynapseException,
 )
 from bittensor.threadpool import PriorityThreadPoolExecutor
+from bittensor.utils.networking import BittensorNTPClient
 
 
 class FastAPIThreadedServer(uvicorn.Server):
@@ -898,14 +898,14 @@ class axon:
             # If we don't have a nonce stored, ensure that the nonce falls within
             # a reasonable delta.
             try:
-                ntp_client = ntplib.NTPClient()
-                response = ntp_client.request('pool.ntp.org')
+                ntp_client = BittensorNTPClient()
+                response = ntp_client.request("pool.ntp.org")
                 current_time = int(response.tx_time * 1e9)  # Convert to nanoseconds
             except Exception as e:
                 print(f"Error fetching NTP time: {e}")
                 # Fallback to local time if NTP fails
                 current_time = time.time_ns()
-            
+
             # Updated nonce using NTP implementated at v7.2
             if synapse.dendrite.version >= 720:
                 # If we don't have a nonce stored, ensure that the nonce falls within
@@ -922,10 +922,10 @@ class axon:
                     raise Exception("Nonce is too old")
             else:
                 if (
-                endpoint_key in self.nonces.keys()
-                and self.nonces[endpoint_key] is not None
-                and synapse.dendrite.nonce <= self.nonces[endpoint_key]
-            ):
+                    endpoint_key in self.nonces.keys()
+                    and self.nonces[endpoint_key] is not None
+                    and synapse.dendrite.nonce <= self.nonces[endpoint_key]
+                ):
                     raise Exception("Nonce is too small")
 
             if not keypair.verify(message, synapse.dendrite.signature):
@@ -1190,8 +1190,8 @@ class AxonMiddleware(BaseHTTPMiddleware):
             )
         synapse.name = request_name
         try:
-            ntp_client = ntplib.NTPClient()
-            response = ntp_client.request('pool.ntp.org')
+            ntp_client = BittensorNTPClient()
+            response = ntp_client.request("pool.ntp.org")
             current_time = int(response.tx_time * 1e9)  # Convert to nanoseconds
         except Exception as e:
             print(f"Error fetching NTP time: {e}")
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -24,6 +24,7 @@ import asyncio
 import contextlib
 import copy
 import inspect
+import ntplib
 import json
 import os
 import threading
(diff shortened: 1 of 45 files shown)


bittensor/utils/networking.py
```diff
@@ -168,3 +175,34 @@ def get_formatted_ws_endpoint_url(endpoint_url: str) -> str:
...
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = ntplib.NTPClient()
+        return cls._instance
+
+    @staticmethod
+    def get_current_ntp_time(retries: int = NTP_POOL_RETRIES) -> int:
+        ntp_servers = ["0.pool.ntp.org", "1.pool.ntp.org", "2.pool.ntp.org"]
```
thealligatorking: can we move these to `constants.py` as well?

bittensor/dendrite.py
```diff
@@ -705,7 +706,7 @@ def process_server_response(
                     # Set the attribute in the local synapse from the corresponding
                     # attribute in the server synapse
                     setattr(local_synapse, key, getattr(server_synapse, key))
-                except:
+                except Exception:
```
thealligatorking: What are your thoughts on making a custom exception get raised here 

```suggestion
                except ServerResponseError as e:
                # logging
                raise
```

bittensor/axon.py
```diff
@@ -888,25 +890,38 @@ async def default_verify(self, synapse: bittensor.Synapse):
...
-            if (
-                self.nonces.get(endpoint_key) is not None
-                and synapse.dendrite.nonce <= self.nonces[endpoint_key]
+                synapse.dendrite.version is not None
+                and synapse.dendrite.version >= V_7_2_0
             ):
-                raise Exception("Nonce is too old")
+                # If we don't have a nonce stored, ensure that the nonce falls within
+                # a reasonable delta.
+                current_time = BittensorNTPClient.get_current_ntp_time()
```
mjurbanski-reef: it was accidental here, but if I would be malicious miner, this is the kind of thing I would try to push to official bittensor package so other miners are running suboptimal code

```
>>> client = ntplib.NTPClient()
>>> def ntp_request():
...     try: response = client.request('pool.ntp.org')
...     except: pass
... 
>>> timeit.timeit(ntp_request, number=10)
15.35390254703816
```
**Delay of 1.5s is going to tank your emissions in any subnet where they are affected by latency.**

Sadly, it is against our (the company I work for) financial interests ( :money_with_wings: ) to even report this, but we try to be honest contributors to bittensor as a whole and do not put our interests before the Bittensor community.

Back to the technical part: This is also synchronous code, so this effect will be cumulative i.e. it will become very easy to DoS any server implementing this.

## Alternatives for implementation

My suggestion would be to save the development resources, and **leave NTP synchronization to processes on Operating system side and use UTC time (as acquired from system clock) here instead**.
This would mean a simple note in README about it.

Yes, they are configured a little bit differently (to drift towards synchronization instead of snapping into "right time" right away), but they do that job really well. 
Trying to reimplement NTP synchronization in will be yet another technical challenge consuming bittensor development team and contributors time of which there is always not enough to go around.

If there is enough time, then **option to use in-process NTP synchronization could be added, but it should always be a opt-in / fallback**, since we will never beat the efficiency of the NTP daemon process on OS side.

bittensor/axon.py
```diff
@@ -888,25 +890,38 @@ async def default_verify(self, synapse: bittensor.Synapse):
...
+                if (
+                    self.nonces.get(endpoint_key) is None
+                    and synapse.dendrite.nonce <= current_time - ALLOWED_DELTA
```
mjurbanski-reef:
```suggestion
                    and synapse.dendrite.nonce <= current_time - ALLOWED_DELTA - (synapse.timeout or 0)
```

When transferring large requests there is a lot of delay (both on the wire and serialization/deserialization).
Without taking `timeout` parameter into account here this change will affect the slower axons in at least one of the subnets I know.

bittensor/axon.py
```diff
@@ -888,25 +890,38 @@ async def default_verify(self, synapse: bittensor.Synapse):
...
+                if (
+                    self.nonces.get(endpoint_key) is not None
+                    and synapse.dendrite.nonce <= self.nonces[endpoint_key]
+                ):
+                    raise Exception("Nonce is too old")
+            else:
+                if (
+                    endpoint_key in self.nonces.keys()
+                    and self.nonces[endpoint_key] is not None
+                    and synapse.dendrite.nonce <= self.nonces[endpoint_key]
```
mjurbanski-reef, line 921: there is bug in here, since if we sent N requests at the same time to the same axon, they can very well return in different order, but I guess it can be resolved another time as it is not a common pattern to do that.

bittensor/axon.py
ibraheem-abe: Thanks for your valuable input. 

Previously, we also noticed some delays when hitting the `pool.ntp.org` directly - which is why we changed the endpoints to specific servers e.g: `0.pool.ntp.org` etc. This resulted in faster access times and was averaging 0.015 seconds per request. 

```
>>> timeit.timeit(ntp_request, number=10)
0.12425095890648663
>>> timeit.timeit(ntp_request, number=10)
0.10041512502357364
>>> timeit.timeit(ntp_request, number=10)
0.10449333302676678
>>> timeit.timeit(ntp_request, number=10)
0.13884595804847777
```

However, based on your results, I did a stress-test and found that some requests could take longer (around 1 - 1.5 seconds). Even though this was rare, but I am assuming it would surface more often in practice - considering real distributed network interactions as compared to us testing locally.

```
>>> timeit.timeit(ntp_request, number=10)
10.176145999925211
```

The main motivation to implement NTP was as you mentioned, to counter clock drifts, skews etc, all the issues with clocks in distributed systems. But at this point I would agree that reverting to system UNIX time would be a better approach. 

We can revisit this gap later on when we have cycles and see feedback regarding the new nonces.

bittensor/constants.py
```diff
@@ -0,0 +1,20 @@
...
+
+# THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+# THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
+
+ALLOWED_DELTA = 4000000000  # Delta of 4 seconds for nonce validation
+V_7_2_0 = 720
```
mjurbanski-reef: :bug: 
Please note this is affected by bug discussed in #1982 (since `6.12.2` already had int value >720).
Fix for it was proposed in #1993 - at this time is yet not merged and after it is, this value will need to be changed.

ibraheem-abe: This fix is merged and we'll roll it out along with 7.2. 
Thanks for your contribution 🚀

bittensor/axon.py
ibraheem-abe: Thanks. We've added timeout in our calculation but kept delta the same for now. 
We'll keep monitoring this in testnet and see if it needs any adjustments

thealligatorking (approved):

PR #1978, opened 2024-06-04
Fix IP/Port packing issue

opendansor:
### Description of the Change

The IP address and port are packed together from subtensor. When this happens we see as port being 0 on our end and need to extract the port from axon info, after extraction, the ip looks to be correct again. 

### Release Notes
Credit to [dgagn](https://github.com/dgagn) 
with PR
https://github.com/opentensor/bittensor/pull/1962


thealligatorking (approved):

camfairchild:
Closing for #1981

Issue #1859, opened 2024-05-09
Retrieving error types from the metadata of the Substrate palette "SubtensorModule" for the btcli console


basfroman:
### Is your feature request related to a problem? Please describe.

The feature related with following issues:
- [subtensor: UI and python client get the error types and error docs from metadata](https://github.com/opentensor/subtensor/issues/384)
- [subtensor: Improve error granularity](https://github.com/opentensor/subtensor/issues/422)

### Describe the solution you'd like

With functionality of `substrateinterface` lib, we retrieve the metadata from `Subtensor`. 
### This allows us to:
1. Avoid hardcoding the error description.
2. Always have up-to-date information about all errors on the server.
3. Automatically receive updates to the list of errors and their descriptions.

### Acceptance Criteria
- [x] Research: Exploring the logic of how the `substrateinterface` package interacts with metadata.
- [x] Research: Exploring structure of metadata by the provided data.
- [x] Implementation: Retrieving metadata from `SubtensorModule`.
- [x] Implementation: Caching a JSON object with a list of errors in the user's directory.
- [x] Implementation: Logic to control the relevancy of cached data relative to server metadata.
- [x] Implementation: Covering new code with unit tests.
- [ ] Implementation: a wrapper for presenting the error in a human-readable view.
- [ ] Testing new logic on a local network.

### Describe alternatives you've considered

Previously, the team considered the possibility of implementing an API endpoint that would return data about the list of errors. After starting the implementation, the decision was canceled in favor of using the capabilities of the substrateinterface package.

basfroman:
I decided to split https://github.com/opentensor/bittensor/issues/1859 into 2 PRs
The first one https://github.com/opentensor/bittensor/pull/1862 (merged)

basfroman:
Updated with:
- [subtensor: UI and python client get the error types and error docs from metadata](https://github.com/opentensor/subtensor/issues/384)
- [subtensor: Improve error granularity](https://github.com/opentensor/subtensor/issues/422)

PR #1980, opened 2024-06-04
A wrapper for presenting extrinsics errors in a human-readable form.

basfroman:
Related with  https://github.com/opentensor/bittensor/issues/1859


bittensor/extrinsics/network.py
```diff
@@ -15,11 +15,26 @@
...
-import bittensor
 
 from rich.prompt import Confirm
 
+import bittensor
+from bittensor.utils import format_error_message
+from ..commands.network import HYPERPARAMS
+
+
+def _find_event_attributes_in_extrinsic_receipt(response, event_name) -> list:
```
thealligatorking: Is it possible to add type hints here?

bittensor/utils/__init__.py
```diff
@@ -260,3 +258,25 @@ def hash(content, encoding="utf-8"):
 
     # Produce the hash
     return sha3.hexdigest()
+
+
+def format_error_message(error_message: dict) -> str:
```
gus-opentensor, line 263: Same here with type hints

bittensor/extrinsics/network.py
basfroman: Sure. Good catch!
Done

thealligatorking (approved):

PR #1989, opened 2024-06-07
Async Subtensor implementation + async substrain

basfroman:
Files `bittensor/asyncbt/subtensor.py`, `bittensor/asyncbt/utils/async_substrate.py` need to be reviewed.

The rest of the files are just copied.


bittensor/asyncbt/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
+from rich.console import Console
+from rich.traceback import install
+
+# Install and apply nest asyncio to allow the async functions
+# to run in a .ipynb
+import nest_asyncio
```
thewhaleking: This is removed in staging, so let's scrap this so it doesn't accidentally make the migration.

thewhaleking (commented):
I'm very confused by this. Specifically the move of the init stuff from `__init__` to `init`, and the creation of `async_subtensor` and the execution (using `asyncio.run`) of `_create_async_subtensor()`. 
No args are passed to the subtensor object on instantiation, and creating it with `asyncio.run` would create a need to nest asyncio to use that object.
What's wrong with the original idea for the implementation of just instantiating it as we have been, and then calling `async with self.substrate` for the calls? This is a standard practice.

thewhaleking (approved):

opendansor:
I think we should remove the 'beginner' label on this PR. 🙂

basfroman:
> I think we should remove the 'beginner' label on this PR. 🙂

Let's leave this as a gentle reminder of where it all began)
Anyway, it's already merged

PR #1993, opened 2024-06-08
fix __version_as_int__ for >10 minor/patch release vers (resolves #1982)

### LINKED ISSUE #1982: bitensor version as int supports only minor,patch versions <10
### Describe the bug

https://github.com/opentensor/bittensor/blob/master/bittensor/__init__.py#L33C1-L37

If bittensor `7.10.1` gets released, it will get encoded as version `801`, and break version comparisons.

While releasing 100 of patch releases maybe far fetched (but not impossible), the release of 10+ minor version is quite common (e.g. Python is currently at 3.13 ).

### To Reproduce

-

### Expected behavior

I would expect `bittensor.__version_as_int__` to be always > than previous version, even if we have more than 10 releases.

A possible solution would be to use `*100` multiplier - after all if we used `*10` we would still require uint32.
This would be also backwards compatible change - `>` comparison between new released version and old one would still give the expected results.

### Screenshots

_No response_

### Environment

-

### Additional context

subtensor seems to encode bittensor/subnet version as u32 int, so we have plenty of room there

mjurbanski-reef:
### Bug

#1982 

### Description of the Change

Uses 1000 base, to fix `6.12.2` `__version_as_int__` being larger than one from release `7.2.1` .

### Alternate Designs

already discussed in #1982 

### Possible Drawbacks

If for some reason int16 was used to store the version number it could be a problem, but glance at subtensor code seems to indicate, int32 was used there.

### Verification Process

Manual + added simple asserts to prevent similar problems from ever occurring in the future (it will basically immediately explode on import)

### Release Notes

- Fix `bittensor.__version_as_int__` from release 6.12.x being higher number than  one from `7.x` line by encoding it using 1000 base instead of 10.
- Deprecated `bittensor.version_split` - use `bittensor.__version__`, `bittensor.__version_info__` or `bittensor.__version_as_int__` instead

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -42,12 +42,15 @@ if (NEST_ASYNCIO_ENV := os.getenv("NEST_ASYNCIO")) in ("1", None):
 # Bittensor code and protocol version.
 __version__ = "7.0.0"
 
-version_split = __version__.split(".")
-__version_as_int__: int = (
-    (100 * int(version_split[0]))
-    + (10 * int(version_split[1]))
-    + (1 * int(version_split[2]))
+_version_split = __version__.split(".")
+__version_info__ = tuple(map(int, _version_split))
+_version_int_base = 1000
+assert max(__version_info__) < _version_int_base
+
+__version_as_int__: int = sum(
+    e * (_version_int_base**i) for i, e in enumerate(reversed(__version_info__))
 )
+assert __version_as_int__ < 2**31  # fits in int32
 __new_signature_version__ = 360
 
 # Rich console.
@@ -56,6 +59,16 @@ __use_console__ = True
 
 # Remove overdue locals in debug training.
 install(show_locals=False)
+
+
+def __getattr__(name):
+    if name == "version_split":
+        warnings.warn(
+            "version_split is deprecated and will be removed in future versions. Use __version__ instead.",
+            DeprecationWarning,
+        )
+        return _version_split
+    raise AttributeError(f"module {__name__} has no attribute {name}")
 
 
 def turn_console_off():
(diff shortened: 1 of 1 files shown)


bittensor/__init__.py
```diff
@@ -42,12 +42,15 @@
 # Bittensor code and protocol version.
 __version__ = "7.0.0"
 
-version_split = __version__.split(".")
-__version_as_int__: int = (
-    (100 * int(version_split[0]))
-    + (10 * int(version_split[1]))
-    + (1 * int(version_split[2]))
+_version_split = __version__.split(".")
+__version_info__ = tuple(map(int, _version_split))
```
basfroman: let's use 
```python
__version_info__ = tuple(int(part) for part in _version_split)
``` 
instead of 
```python
__version_info__ = tuple(map(int, _version_split))
```

in this example it is the same in terms of optimization, but code readability is much better

ibraheem-abe: @RomanCh-OT I can make this change when this gets merged. Just waiting for confirmation from subtensor side

mjurbanski-reef: Applied;

as for version being u32,  AFAIK, that was needed to be checked was in 
https://github.com/opentensor/subtensor/blob/0c77062a10cadfbab8507b7dccb2e1ea49848e58/pallets/subtensor/src/serving.rs#L57C9-L57C16
even with prometheus is u32 , but that version number is arbitrary set by node so it doesn't even matter in our context
https://github.com/opentensor/subtensor/blob/0c77062a10cadfbab8507b7dccb2e1ea49848e58/pallets/subtensor/src/serving.rs#L160

So it should be safe.

thewhaleking (approved):

PR #1994, opened 2024-06-10
Metagraph async

ibraheem-abe:
Async migration for the following modules:

- metagraph.py
- commands/metagraph.py
- metagraph in subtensor

### DIFF
--- a/bittensor/commands/metagraph.py
+++ b/bittensor/commands/metagraph.py
@@ -15,6 +15,7 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
+import asyncio
 import argparse
 import bittensor
 from rich.table import Table
@@ -76,13 +77,13 @@ class MetagraphCommand:
             subtensor: "bittensor.subtensor" = bittensor.subtensor(
                 config=cli.config, log_verbose=False
             )
-            MetagraphCommand._run(cli, subtensor)
+            asyncio.run(MetagraphCommand._run(cli, subtensor))
         finally:
             if "subtensor" in locals():
                 subtensor.close()
                 bittensor.logging.debug("closing subtensor connection")
 
-    def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
+    async def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
         r"""Prints an entire metagraph."""
         console = bittensor.__console__
         console.print(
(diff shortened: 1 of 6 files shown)


bittensor/commands/metagraph.py
```diff
@@ -76,27 +77,31 @@ def run(cli: "bittensor.cli"):
             subtensor: "bittensor.subtensor" = bittensor.subtensor(
                 config=cli.config, log_verbose=False
             )
-            MetagraphCommand._run(cli, subtensor)
+            asyncio.run(MetagraphCommand._run(cli, subtensor))
```
basfroman, line 80: Since we are doing async work on the entire Bittensor package, it is assumed that all the code is async, include `%CommsndName%.run` methods.
The entry point to Bittensor SDK is the `btcli` utility. This is where we will create an event loop to execute/run/use async code.
@thewhaleking

thewhaleking: I actually disagree. I think that with the current architecture, each command is basically its own entrypoint. While I don't necessarily think that's the best architecture design, fixing it currently will delay our async efforts. Because we'd ideally not create a new subtensor object in each `run` method.

basfroman: IMHO
My approach is this:

1. In btcli we create an event loop
2. In btcli we create a Subtensor instance
3. We pass the Subtensor instance to the Commands.run() as a argument.
4. The same way the commander will do it.

In this case, we get:

1. a completely asynchronous SDK
2. the correct architecture, which will not have many sync entry points. 

This is best practices of the async impl.

thewhaleking: While I agree wholeheartedly that is the correct approach overall, I think getting the async stuff sorted first, and then going back and fixing the architecture is the best method, because of the 26 June deadline for this.

basfroman: I'm OK with this approach. 
Migrate the bittensor -> migrate the commands -> apply the abstract class to the commands (bring those to one controlled structure) -> change the use bittensor (sync -> async) in btcli.
Sounds like a great plan!

basfroman (approved):

thewhaleking:
@ibraheem-opentensor hey I didn't get a chance to review this, but let's update this:
```python
difficulty = await subtensor.difficulty(cli.config.netuid)
subnet_emission = bittensor.Balance.from_tao(
    await subtensor.get_emission_value_by_subnet(cli.config.netuid)
)
total_issuance = bittensor.Balance.from_rao(
    (await subtensor.total_issuance()).rao
)
```
to
```python
difficulty, subnet_emission_, total_issuance_ = await asyncio.gather(
    subtensor.difficulty(cli.config.netuid),
    subtensor.get_emission_value_by_subnet(cli.config.netuid),
    subtensor.total_issuance()
)
subnet_emission = bittensor.Balance.from_tao(subnet_emission_)
total_issuance = bittensor.Balance.from_rao(total_issuance_).rao
```

We want to take full advantage of asyncio with big push.

Issue #1990, opened 2024-06-07
Async migration for bittensor/extrinsic sub-package


basfroman:
### Is your feature request related to a problem? Please describe.

As a next step after `bittensor/subtensor.py` migrated to async the `bittensor/extrinsic` sub-package have to be next one.

### Describe the solution you'd like

sync -> async for `bittensor/extrinsic sub-package`

RomanCh-OT:
https://github.com/opentensor/bittensor/pull/1991

basfroman:
Related fix https://github.com/opentensor/bittensor/pull/1992

basfroman:
Part 2 https://github.com/opentensor/bittensor/pull/1995
Part 3 https://github.com/opentensor/bittensor/pull/1996

basfroman:
Part 4 https://github.com/opentensor/bittensor/pull/2000

basfroman:
Part 5 https://github.com/opentensor/bittensor/pull/2008

basfroman:
`extrinsics/unstaking.py` https://github.com/opentensor/bittensor/pull/2001

PR #1995, opened 2024-06-10
Async for `bittensor/extrinsics/network.py`

basfroman:
Part 2 of https://github.com/opentensor/bittensor/issues/1990

### DIFF
--- a/bittensor/extrinsics/network.py
+++ b/bittensor/extrinsics/network.py
@@ -15,10 +15,19 @@
 # THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
+
+"""
+Network module provides functions to interact with the Subtensor blockchain, specifically for registering subnetworks
+and setting hyperparameters.
+"""
+
 import time
-import bittensor
 
 from rich.prompt import Confirm
+
+import bittensor
+from bittensor.utils import balance
+from ..commands.network import HYPERPARAMS
 
 
 def register_subnetwork_extrinsic(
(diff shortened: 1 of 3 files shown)


bittensor/extrinsics/network.py
```diff
@@ -15,37 +15,44 @@
...
+
         success (bool):
-            Flag is ``true`` if extrinsic was finalized or included in the block.
-            If we did not wait for finalization / inclusion, the response is ``true``.
+            Flag is ``true`` if extrinsic was finalized or included in the block. If we did not wait for finalization / inclusion, the response is ``true``.
     """
-    your_balance = subtensor.get_balance(wallet.coldkeypub.ss58_address)
-    burn_cost = bittensor.utils.balance.Balance(subtensor.get_subnet_burn_cost())
+    your_balance = await subtensor.get_balance(wallet.coldkeypub.ss58_address)
+    burn_cost = balance.Balance(await subtensor.get_subnet_burn_cost())
```
thewhaleking: This should be gathered.

bittensor/extrinsics/network.py
```diff
@@ -62,48 +69,58 @@ def register_subnetwork_extrinsic(
...
+            wait_for_inclusion=wait_for_inclusion,
+            wait_for_finalization=wait_for_finalization,
+        )
+
+        # We only wait here if we expect finalization.
+        if not wait_for_finalization and not wait_for_inclusion:
+            return True
+
+        # process if registration successful
+        # TODO: ? passably have to be awaited after `submit_extrinsic` implemented in `async_substrate.py`
```
thewhaleking, line 96: "Possibly". This is a good callout. Not yet finished with the extrinsics stuff in substrate.

bittensor/extrinsics/network.py
```diff
@@ -167,46 +174,44 @@ def set_hyperparameter_extrinsic(
...
-            # Successful registration, final check for membership
-            else:
-                bittensor.__console__.print(
-                    f":white_heavy_check_mark: [green]Hyper parameter {parameter} changed to {value}[/green]"
+        # We only wait here if we expect finalization.
+        if not wait_for_finalization and not wait_for_inclusion:
+            return True
+
+        # process if registration successful
+        # TODO: ? passably have to be awaited after `submit_extrinsic` implemented in `async_substrate.py`
```
thewhaleking, line 206: "Possibly"

bittensor/extrinsics/network.py
```diff
@@ -167,46 +174,44 @@ def set_hyperparameter_extrinsic(
...
+        response.process_events()
+        if not response.is_success:
+            bittensor.__console__.print(
+                ":cross_mark: [red]Failed[/red]: error:{}".format(
+                    response.error_message
                 )
-                return True
+            )
+            await asyncio.sleep(0.5)
```
thewhaleking, line 214: Why do we wait here? I know we do in the original also, but why?

bittensor/subtensor.py
```diff
@@ -5106,7 +5106,7 @@ async def get_subnet_burn_cost(self, block: Optional[int] = None) -> Optional[st
         if lock_cost is None:
             return None
 
-        return lock_cost
+        return int(lock_cost)
```
thewhaleking, line 5109: The docs state this should be in Tao (which is usually float). Is it actually in Rao? Even if so, why are we changing the type here?

bittensor/utils/async_substrate.py
```diff
@@ -344,7 +349,8 @@ async def get_block_number(self, block_hash: str) -> int:
                 return int(response["result"]["number"], 16)
 
     def close(self):
-        raise NotImplementedError()
+        self.substrate.close()
+        # TODO: Websocket close need to be implemented and added here.
```
thewhaleking, line 353: Is there a reason we want to manually allow closing the connection?

thewhaleking (changes requested):

bittensor/extrinsics/network.py
basfroman: Good catch! Done

bittensor/extrinsics/network.py
basfroman: Honestly, I don't see any valuable reasons for this action. But I just convert this into async. Let's remove sleep at all?

bittensor/utils/async_substrate.py
basfroman: For the same reasons as in original SubstrateInterface. Then it used in Subtensor.

thewhaleking (approved):

PR #1997, opened 2024-06-11
Async for `commands/register.py` 

ibraheem-abe:
- Adds async migration to `commands/register.py`

### DIFF
--- a/bittensor/commands/register.py
+++ b/bittensor/commands/register.py
@@ -17,6 +17,7 @@
 
 import sys
 import argparse
+import asyncio
 import bittensor
 from rich.prompt import Prompt, Confirm
 from .utils import check_netuid_set, check_for_cuda_reg_config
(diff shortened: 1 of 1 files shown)


bittensor/commands/register.py
```diff
@@ -68,27 +69,27 @@ def run(cli: "bittensor.cli"):
...
             bittensor.__console__.print(
                 f"[red]Subnet {cli.config.netuid} does not exist[/red]"
             )
             sys.exit(1)
 
         # Check current recycle amount
-        current_recycle = subtensor.recycle(netuid=cli.config.netuid)
-        balance = subtensor.get_balance(address=wallet.coldkeypub.ss58_address)
+        current_recycle = await subtensor.recycle(netuid=cli.config.netuid)
+        balance = await subtensor.get_balance(address=wallet.coldkeypub.ss58_address)
```
thewhaleking, line 92: `asyncio.gather`

bittensor/commands/register.py
```diff
@@ -127,7 +128,7 @@ def add_args(parser: argparse.ArgumentParser):
         bittensor.subtensor.add_args(register_parser)
 
     @staticmethod
-    def check_config(config: "bittensor.config"):
+    async def check_config(config: "bittensor.config"):
```
thewhaleking, line 131: If we change `check_config` to a coroutine here, we're going to need to update `cli.check_config`.

thewhaleking (changes requested):

basfroman:
Let's forget about `bittensor/commands` for now? 
My suggestion:
1. Complete the full async migration of the entire bittensor (except for `bittensor/commands` and `bin/btcli`)
2. Implement an abstract class for all modules in the commands sub-package
3. Migrate commands modules
4. Migrate btcli

We still need to do this. Now it looks like double work.
Make sense?

PR #2000, opened 2024-06-11
Asyc `bittensor/extrinsics` (part 4)

basfroman:
Inclused async migration for:
- `bittensor/extrinsics/root.py`
- `bittensor/extrinsics/senate.py`.
- `bittensor/extrinsics/serving.py`
All of them refactored, has some small bug fixes, and async migrated.

For easy review you can do this via commits.

### DIFF
--- a/bittensor/extrinsics/set_weights.py
+++ b/bittensor/extrinsics/set_weights.py
@@ -96,7 +96,7 @@ def set_weights_extrinsic(
         ":satellite: Setting weights on [white]{}[/white] ...".format(subtensor.network)
     ):
         try:
-            success, error_message = subtensor._do_set_weights(
+            success, error_message = subtensor.do_set_weights(
                 wallet=wallet,
                 netuid=netuid,
                 uids=weight_uids,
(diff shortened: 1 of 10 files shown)


opendansor (approved):
LGTM

ibraheem-abe (approved):

bittensor/extrinsics/set_weights.py
```diff
@@ -96,7 +96,7 @@ def set_weights_extrinsic(
         ":satellite: Setting weights on [white]{}[/white] ...".format(subtensor.network)
     ):
         try:
-            success, error_message = subtensor._do_set_weights(
+            success, error_message = subtensor.do_set_weights(
```
thewhaleking, line 99: Coroutine is never awaited.

thewhaleking (changes requested):

bittensor/extrinsics/set_weights.py
basfroman: Wait. I almost broke down. But then I realized that this file should not be in this PR :)

PR #2029, opened 2024-06-14
Fix incorrent import order of `bittensor/commands/utils.py:defaults` a potential import error

basfroman:


bittensor/commands/utils.py
```diff
@@ -15,6 +15,7 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
+from munch import Munch, munchify
 import sys
 import os
 import bittensor
```
camfairchild:
```suggestion
import sys
import os
from munch import Munch, munchify
import bittensor
```
nit: move munch after os,sys

camfairchild (dismissed):

bittensor/commands/utils.py
basfroman: Good catch! I forgot to optimize the import in this file.

bittensor/commands/utils.py
```diff
@@ -15,25 +15,69 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from dataclasses import dataclass
 import os
-import requests
 import sys
+from dataclasses import dataclass
```
thewhaleking, line 20: I understand almost all of these import order changes, but I don't get this one. This is non-blocking. Just looking for the rationale.

bittensor/commands/wallets.py
```diff
@@ -73,11 +73,11 @@ async def run(cli: "bittensor.cli"):
             # Password can be "", assume if None
             json_password = cli.config.get("json_password", "")
         wallet.regenerate_coldkey(
+            use_password=cli.config.use_password,
+            overwrite=cli.config.overwrite_coldkey,
```
thewhaleking, line 77: Why are these changed? Again, it's non-blocking. Just looking for rationale

thewhaleking (approved):
Approved. Just make absolutely sure Raj is aware of the change of `defaults` for release, because this is breaking.

bittensor/commands/wallets.py
basfroman: In the function `wallet.regenerate_coldkey`, `use_password` and `overwrite` are two first arguments. 
This is making readability more obvious and shows us that the last arguments passed as a part of `**kwargs`

bittensor/commands/utils.py
basfroman: There is a convention for the sequence of correct import of packages: first those built into Python-SDK, then installed (third party) and then everything that is implemented locally (in the product being developed). This is mentioned in the PEP.

thewhaleking: Correct. But `dataclasses` is a stdlib package.

basfroman:
> Approved. Just make absolutely sure Raj is aware of the change of `defaults` for release, because this is breaking.

Could you please explain why this is breaking? I didn't change `defaults` at all, just moved to logically correct place. This is a local `defaults` for a sub-package of `commands`, not related to the `defaults` in the bittensor root (__ init __.py).
Am I missing something?

bittensor/commands/utils.py
basfroman: Indeed, if I did `import dataclasses`, then use it in the code as `@ dataclasses. dataclass`
But in every section I mentioned, `import %package%` is done first, and then everything that is `from %package% import %object%`

thewhaleking: Okay. I've never heard of that style of importing. I've always just used the PEP8 convention of stdlib,pypi,local (sorted alphabetically).

thewhaleking:
> > Approved. Just make absolutely sure Raj is aware of the change of `defaults` for release, because this is breaking.
> 
> Could you please explain why this is breaking? I didn't change `defaults` at all, just moved to logically correct place. This is a local `defaults` for a sub-package of `commands`, not related to the `defaults` in the bittensor root (__ init __.py). Am I missing something?

It's breaking because you're moving its location.

basfroman:
> > > Approved. Just make absolutely sure Raj is aware of the change of `defaults` for release, because this is breaking.
> > 
> > 
> > Could you please explain why this is breaking? I didn't change `defaults` at all, just moved to logically correct place. This is a local `defaults` for a sub-package of `commands`, not related to the `defaults` in the bittensor root (__ init __.py). Am I missing something?
> 
> It's breaking because you're moving its location.

Raj is informed

PR #2035, opened 2024-06-17
Adds normalization for alpha hyperparams

ibraheem-abe:
- Adds normalization for alpha hyperparams

### DIFF
--- a/bittensor/commands/utils.py
+++ b/bittensor/commands/utils.py
@@ -186,10 +186,10 @@ def filter_netuids_by_registered_hotkeys(
         )
         netuids_with_registered_hotkeys.extend(netuids_list)
 
-    if cli.config.netuids == None or cli.config.netuids == []:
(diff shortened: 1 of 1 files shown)


bittensor/commands/utils.py
```diff
@@ -186,10 +186,10 @@ def filter_netuids_by_registered_hotkeys(
         )
         netuids_with_registered_hotkeys.extend(netuids_list)
 
-    if cli.config.netuids == None or cli.config.netuids == []:
+    if cli.config.netuids is None:
```
thewhaleking: Are we sure we don't want to do `if not cli.config.netuids:`?

ibraheem-abe: Good one 👍 incorporated

thealligatorking (approved):

thewhaleking (approved):

PR #2042, opened 2024-06-17
Async tests fix for `tests/integration_tests` directory + e2e tests commented

basfroman:
Async tests fix for `tests/integration_tests` directory

### DIFF
--- a/bittensor/mock/subtensor_mock.py
+++ b/bittensor/mock/subtensor_mock.py
@@ -1,30 +1,30 @@
 # The MIT License (MIT)
 # Copyright © 2022-2023 Opentensor Foundation
-
+#
 # Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
 # documentation files (the “Software”), to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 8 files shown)


ibraheem-abe (approved):

thewhaleking (changes requested):
None of these coroutine tests appear to actually run. You can test it by adding `self.assertEqual(1, 2)` into any of them, and running the suite, and they will still pass.

basfroman:
> None of these coroutine tests appear to actually run. You can test it by adding `self.assertEqual(1, 2)` into any of them, and running the suite, and they will still pass.

You defenately right! We have to use `unittest.IsolatedAsyncioTestCase` instead of `unittest.TestCase` for making correct async tests based on unittest lib.

basfroman:
**Additional context:**
- real fix `tests/integration_tests/test_subtensor_integration.py` with async unittest.
- commented blocking test in `tests/integration_tests/test_cli.py` because those tests based on sync `cli.py` module. Will be fixed later.
- test_cli.py and e2e tests marked commented until commands and cli migrated. 
- `TODO: fix after commands and cli are async migrated`

thewhaleking (approved):

PR #2050, opened 2024-06-20
Enhancement: Detailed nonce information in-case of failures

ibraheem-abe:
- Increases level of information logged in-case nonces fail. 
- Differentiates between different nonce failure scenarios and logsaccordingly.
- Ensures synapse timeout is calculated in ns

### DIFF
--- a/bittensor/axon.py
+++ b/bittensor/axon.py
@@ -44,7 +44,7 @@ from starlette.responses import Response
 from substrateinterface import Keypair
 
 import bittensor
-from bittensor.utils.axon import allowed_nonce_window_ns, calculate_diff_seconds
+from bittensor.utils.axon_utils import allowed_nonce_window_ns, calculate_diff_seconds
 from bittensor.constants import V_7_2_0
 from bittensor.errors import (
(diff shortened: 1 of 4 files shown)


bittensor/axon.py
```diff
@@ -893,33 +895,44 @@ async def default_verify(self, synapse: bittensor.Synapse):
...
+                synapse_timeout_ns = (synapse.timeout or 0) * NANOSECONDS_IN_SECOND
+                allowed_window_ns = current_time_ns - ALLOWED_DELTA - synapse_timeout_ns
                 if (
                     self.nonces.get(endpoint_key) is None
-                    and synapse.dendrite.nonce
-                    <= time.time_ns() - ALLOWED_DELTA - (synapse.timeout or 0)
+                    and synapse.dendrite.nonce <= allowed_window_ns
                 ):
-                    raise Exception("Nonce is too old")
+                    diff_seconds = (
```
thealligatorking: can we move this to a helper method so we can get coverage on it?

thealligatorking (approved):

PR #2055, opened 2024-06-21
Async for `commands`, `cli`, `btcli`

basfroman:
Not final solution

### DIFF
--- a/bin/btcli
+++ b/bin/btcli
@@ -30,6 +30,9 @@ import bittensor
 from bittensor import cli as btcli
 
 
+bittensor.logging.on()
+
+
 async def main():
     # Create the parser with shtab support
     parser = btcli.__create_parser__()
--- a/bin/btcli
+++ b/bin/btcli
@@ -41,7 +41,7 @@ async def main():
 
     try:
         async with btcli(args=sys.argv[1:]) as cli_instance:
-            cli_instance.run()
+            await cli_instance.run()
     except KeyboardInterrupt:
         print('KeyboardInterrupt')
     except RuntimeError as e:
--- a/bin/btcli
+++ b/bin/btcli
@@ -1,5 +1,26 @@
 #!/usr/bin/env python
 
+# The MIT License (MIT)
+# Copyright © 2021 Yuma Rao
+# Copyright © 2022 Opentensor Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the “Software”), to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+# the Software.
+#
+# THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+# THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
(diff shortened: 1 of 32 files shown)


bin/btcli
```diff
@@ -19,8 +40,8 @@ def main():
         return
 
     try:
-        cli_instance = btcli(args=sys.argv[1:])
-        cli_instance.run()
+        async with btcli(args=sys.argv[1:]) as cli_instance:
+            cli_instance.run()
```
thewhaleking: Needs to be `await`

bittensor/commands/delegates.py
```diff
@@ -554,10 +556,12 @@ def check_config(config: "bittensor.config"):
...
                 subtensor = bittensor.subtensor(config=config, log_verbose=False)
-                delegates: List[bittensor.DelegateInfo] = subtensor.get_delegates()
+                delegates: List[
+                    bittensor.DelegateInfo
+                ] = await subtensor.get_delegates()
                 try:
-                    prev_delegates = subtensor.get_delegates(
-                        max(0, subtensor.block - 1200)
+                    prev_delegates = await subtensor.get_delegates(
+                        max(0, await subtensor.block - 1200)
```
thewhaleking, line 565: I don't think this will work, but I might be wrong. Awaiting properties is weird behavior.

bittensor/commands/delegates.py
```diff
@@ -657,22 +658,22 @@ def run(cli: "bittensor.cli"):
...
+    async def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
+        """List all delegates on the network."""
         with bittensor.__console__.status(":satellite: Loading delegates..."):
-            delegates: list[bittensor.DelegateInfo] = subtensor.get_delegates()
+            delegates: list[bittensor.DelegateInfo] = await subtensor.get_delegates()
 
             try:
-                prev_delegates = subtensor.get_delegates(max(0, subtensor.block - 1200))
+                prev_delegates = subtensor.get_delegates(
+                    max(0, await subtensor.block - 1200)
```
thewhaleking: Same thing here with `await` on a property.

bittensor/commands/delegates.py
```diff
@@ -733,37 +734,37 @@ class NominateCommand:
...
     @staticmethod
-    def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
+    async def _run(cli: "bittensor.cli", subtensor: "bittensor.subtensor"):
         r"""Nominate wallet."""
         wallet = bittensor.wallet(config=cli.config)
 
         # Unlock the wallet.
         wallet.hotkey
-        wallet.coldkey
+        wallet.coldkeya
```
thewhaleking: Typo, unless this was changed from `wallet.coldkey`

thewhaleking (changes requested):

bittensor/commands/delegates.py
basfroman: await added to the call.
Awaiting `properties` is normal thing for async

```python
In [19]: class Some:
    ...:     @property
    ...:     async def value(self):
    ...:         return "123"
    ...: 

In [20]: async def main():
    ...:     s = Some()
    ...:     print(await s.value)
    ...: 

In [21]: def main1():
    ...:     s = Some()
    ...:     print(s.value)
    ...: 

In [22]: main()
Out[22]: <coroutine object main at 0x1086931d0>

In [23]: asyncio.run(main())
123
```

bittensor/commands/delegates.py
basfroman: done

bittensor/commands/delegates.py
basfroman: Awaiting `properties` is normal thing for async

bin/btcli
basfroman: strange, I was going to commit this)

thewhaleking (approved):

bittensor/commands/delegates.py
RomanCh-OT: it's works correctly

bittensor/subtensor.py
```diff
@@ -4456,10 +4470,17 @@ async def get_delegates(self, block: Optional[int] = None) -> List[DelegateInfo]
 
         """
 
-        @retry(stop=stop_after_attempt(3), wait=wait_fixed(1))
+        @retry(
+            stop=stop_after_attempt(3),
+            wait=wait_fixed(1),
+            retry=retry_if_exception_type(SubstrateRequestException),
```
thewhaleking: You're actually going to want to use `retry=retry_if_not_exception_type(SubstrateRequestException)` rather than `retry_if_exception_type`, because otherwise it will never raise `SubstrateRequestException`.

thewhaleking (changes requested):

bittensor/subtensor.py
basfroman: Sounds reasonable. I applied this in a fix to this project, which I am now preparing for review.

PR #2059, opened 2024-06-24
async retry reraise logic

thewhaleking:
Added in reraising logic for all `@retry` attempts, to keep consistency with older retry logic.

The old `@retry` would reraise the last exception. This changes in tenacity. To get the same execution, I added the `reraise=True` arg to each call.


basfroman:
This has already been fixed in the PR I'm working on, as I mentioned on the channel.
You can close or merge this PR.

basfroman (approved):
lgtm

PR #2065, opened 2024-06-24
Fix: Logging configs not being set

ibraheem-abe:
Please see this for more context:
https://github.com/opentensor/KnowledgeBase/issues/30#issuecomment-2187469983

### DIFF
--- a/bittensor/btlogging/loggingmachine.py
+++ b/bittensor/btlogging/loggingmachine.py
@@ -432,9 +432,7 @@ class LoggingMachine(StateMachine):
         prefix_str = "" if prefix is None else prefix + "."
         try:
             default_logging_debug = os.getenv("BT_LOGGING_DEBUG") or False
-            print(f"BT_LOGGING_TRACE {os.getenv('BT_LOGGING_TRACE')}")
             default_logging_trace = os.getenv("BT_LOGGING_TRACE") or False
-            print(default_logging_trace)
             default_logging_record_log = os.getenv("BT_LOGGING_RECORD_LOG") or False
             default_logging_logging_dir = (
                 os.getenv("BT_LOGGING_LOGGING_DIR") or "~/.bittensor/miners"
--- a/bittensor/btlogging/loggingmachine.py
+++ b/bittensor/btlogging/loggingmachine.py
@@ -70,19 +70,19 @@ class LoggingMachine(StateMachine):
         | Default.to(Default)
     )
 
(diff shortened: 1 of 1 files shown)


thealligatorking (approved):

bittensor/btlogging/loggingmachine.py
```diff
@@ -70,19 +70,19 @@ class LoggingMachine(StateMachine):
...
     )
 
-    disable_trace: Default = Trace.to(Default)
+    disable_trace = Trace.to(Default)
 
-    disable_debug: Default = Debug.to(Default)
+    disable_debug = Debug.to(Default)
 
-    disable_logging: Disabled = (
+    disable_logging = (
```
thewhaleking, line 85: Why are we removing type hints?

bittensor/btlogging/loggingmachine.py
```diff
@@ -107,14 +107,31 @@ def __init__(self, config: bittensor.config, name: str = BITTENSOR_LOGGER_NAME):
...
         # set up all the loggers
         self._logger = self._initialize_bt_logger(name)
         self.disable_third_party_loggers()
+        self._enable_initial_state(self._config)
+
+    def _enable_initial_state(self, config):
+        """Set correct state action on initializing"""
+        if config.trace:
+            self.enable_trace()
+        elif config.debug:
```
thewhaleking, line 124: This `elif` would only enable debug if trace is not enabled. Is that what we want?

bittensor/btlogging/loggingmachine.py
ibraheem-abe: I think the type hints here are not required as, for example, 'Debug' is not a type here; instead it is an instance of the State class in the Statemachine. 
The IDE also warns with the error
`Variable not allowed in type expression`

bittensor/btlogging/loggingmachine.py
ibraheem-abe: Yes, that is the intended behavior. At a time, we can enable 1 exclusive logging state—Trace, Debug, or Default—based on the configuration provided.

**Trace**: This is the most verbose level, logging all messages including trace level messages. This has a level of 5. It contains all logging levels i.e. trace + debug + default

**Debug**: This has a level of 10. It contains debug + default logging levels. 

**Default**: This has a level of 20. It only contains default levels i.e. info, error, warning, success

thewhaleking (approved):

PR #2066, opened 2024-06-24
Feat: Parallel e2e tests 

ibraheem-abe:
We are currently going with 2 parallel machines running our tests. 
The workload is distributed according to the folder directory in our e2e tests. 

**Note**: Tests where we wait for an epoch must be distributed amongst these folders otherwise one machine may have to wait for too long if there are multiple tests. 

**Additional note**: We also tested spinning up an additional machine exclusively for tests where we are waiting for epochs, but the execution and cost downsides didn't justify the minimum improvement in results (yet). We can revisit this case as we increase our coverage.


thewhaleking:
Is this still needed, or can this be closed?

ibraheem-abe:
We can close this

PR #2084, opened 2024-06-28
Merger `staging` into `merge-async`

basfroman:
4 tests requires fix after this PR:
- `tests/e2e_tests/multistep/test_axon.py` - regarding `metagraph` changes;
-  `tests/e2e_tests/subcommands/hyperparams/test_liquid_alpha.py` and `tests/e2e_tests/subcommands/stake/test_stake_show.py` require captured log fix. Possibly because of async.
-  `tests/e2e_tests/multistep/test_emissions.py` - an investigation into the behavior of the async implementation of `wait_interval` is required.

Since e2e tests take a lot of time to investigate, it was decided to do this in the next PR, after the merger.


tests/e2e_tests/multistep/test_axon.py
```diff
@@ -96,13 +96,9 @@ async def test_axon(local_chain):
...
     # ignore, don't await coroutine, just write logs to file
-    await asyncio.create_task(
-        write_output_log_to_file("axon_stdout", axon_process.stdout)
-    )
+    asyncio.create_task(write_output_log_to_file("axon_stdout", axon_process.stdout))
     # ignore, dont await coroutine, just write logs to file
-    await asyncio.create_task(
-        write_output_log_to_file("axon_stderr", axon_process.stderr)
-    )
+    asyncio.create_task(write_output_log_to_file("axon_stderr", axon_process.stderr))
```
thewhaleking, line 104: These need to be awaited, or else they won't write the logs to the file.

tests/e2e_tests/multistep/test_dendrite.py
```diff
@@ -126,11 +125,11 @@ async def test_dendrite(local_chain):
     # record logs of process
     # Create tasks to read stdout and stderr concurrently
     # ignore, dont await coroutine, just write logs to file
-    await asyncio.create_task(
+    asyncio.create_task(
         write_output_log_to_file("dendrite_stdout", dendrite_process.stdout)
     )
     # ignore, dont await coroutine, just write logs to file
-    await asyncio.create_task(
+    asyncio.create_task(
```
thewhaleking, line 133: These also need to be awaited. Same reason.

tests/e2e_tests/multistep/test_incentive.py
```diff
@@ -120,13 +120,9 @@ async def test_incentive(local_chain):
...
     # ignore, don't await coroutine, just write logs to file
-    await asyncio.create_task(
-        write_output_log_to_file("miner_stdout", miner_process.stdout)
-    )
+    asyncio.create_task(write_output_log_to_file("miner_stdout", miner_process.stdout))
     # ignore, dont await coroutine, just write logs to file
-    await asyncio.create_task(
-        write_output_log_to_file("miner_stderr", miner_process.stderr)
-    )
+    asyncio.create_task(write_output_log_to_file("miner_stderr", miner_process.stderr))
```
thewhaleking, line 126: These need to be awaited.

tests/e2e_tests/multistep/test_incentive.py
```diff
@@ -163,11 +159,11 @@ async def test_incentive(local_chain):
 
     # Create tasks to read stdout and stderr concurrently and write output to log file
     # ignore, don't await coroutine, just write logs to file
-    await asyncio.create_task(
+    asyncio.create_task(
         write_output_log_to_file("validator_stdout", validator_process.stdout)
     )
     # ignore, dont await coroutine, just write logs to file
-    await asyncio.create_task(
+    asyncio.create_task(
```
thewhaleking, line 168: Need to be awaited.

tests/e2e_tests/utils.py
```diff
@@ -1,17 +1,19 @@
...
 
 from substrateinterface import SubstrateInterface
 
 import bittensor
 from bittensor import Keypair, logging
 
 template_path = os.getcwd() + "/neurons/"
-repo_name = "templates repository"
+templates_repo = "templates repository"
+ASYNC_TEMPL_URL = "https://api.github.com/repos/opentensor/bittensor-subnet-template/commits/async-metagraph-for-async-e2e-tests-only"
```
thewhaleking, line 18: I don't like this being a hardcoded github link for longterm maintenance. Add a TODO comment here.

thewhaleking (changes requested):

tests/e2e_tests/multistep/test_axon.py
basfroman: According to Daniel, he uses this for debugging, so he asked to ignore it.

tests/e2e_tests/multistep/test_dendrite.py
basfroman: According to Daniel, he uses this for debugging, so he asked to ignore it.

tests/e2e_tests/multistep/test_incentive.py
basfroman: According to Daniel, he uses this for debugging, so he asked to ignore it.

tests/e2e_tests/multistep/test_incentive.py
basfroman: According to Daniel, he uses this for debugging, so he asked to ignore it.

tests/e2e_tests/utils.py
basfroman: There is an expression: if you deny something, then offer a solution ;)
Now this is used not just as hardcode, but as a constant within the test. Our goal is to provide the opportunity for this constant to exist outside the test. No matter how we try to do this, inside the test it will all be a constant.
Does it make sense?

thewhaleking (approved):

PR #2105, opened 2024-07-09
Feature/roko/profile and config

unforkableco:
WIP

Adds Profile Commands:

```
    ProfileCommand,
    ProfileListCommand,
    ProfileShowCommand,
    ProfileDeleteCommand,
    ProfileSetValueCommand,
    ProfileDeleteValueCommand,
    ProfileSetCommand
```

Usage:

```
$ btcli profile create
$ btcli p create
```

WIP


unforkableco:
# Input Needed on Handling value precedence. Args Parser vs. Profile Config.

## Status Quo:

Implemented: Profile Commands 
(Ticket #7 ROKOxBittensor Project Board)

Implemented: Parameter Overlay via Merging Strategy as Outined in Ticket
(Ticket #3 ROKOxBittensor Project Board)

https://github.com/opentensor/bittensor/pull/2105/files#diff-373be912eb8f8235ff39a9568ef098023b56c305a2e0b5d4c92b8d2d314e8109R252

## issue Description: What is the preferred way we should implement this? Input needed from Bittensor:

Currently default values are written into the config using a check via the __is_set function if they are default values.

https://github.com/opentensor/bittensor/pull/2105/files#diff-373be912eb8f8235ff39a9568ef098023b56c305a2e0b5d4c92b8d2d314e8109R432

We are adding more configs now (generic, profile, command line args)

Because of the way the Argument Parser works and since Arguments should always have precedence over Profile simply merging the profile config does not work as intended.

We can either change the commands like create, delete etc. to always prompt the user which to us (Bad UX)

vs.

Complete Refactoring of the Parameter handling

## Solution #1 (hacky and not clean, worse user experience)
### Go through the Commands and always ask which wallet etc. should be used.

This is bad for UX since the reason we use Profile should make it less work and "just work"
e.g.: 
```btcli wallet balance``` should just print the balance of the set wallet in profile or manually set wallet via args.


## Solution #2 (clean but deeper codebase changes, better UX)
### Refactor Argument Parser.

Cleaner Solution but changes the behavior of the Argument Parser which makes changes in almost all CLI related files necessary.

Doing this could also Refactor other things like Config Initialisation which is done x4 times right now.

PR #2106, opened 2024-07-09
Adds feature to get arbitration period for coldkey swap

thewhaleking:
Adds method in subtensor to get remaining blocks til previously scheduled coldkey swap. Displays this in human readable time when the user begins a coldkey swap using the btcli with a key that has already been scheduled.

### DIFF
--- a/bittensor/utils/formatting.py
+++ b/bittensor/utils/formatting.py
@@ -20,3 +20,11 @@ def millify(n: int):
     )
 
     return "{:.2f}{}".format(n / 10 ** (3 * millidx), millnames[millidx])
+
+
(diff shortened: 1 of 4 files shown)


bittensor/utils/formatting.py
```diff
@@ -20,3 +20,11 @@ def millify(n: int):
     )
 
     return "{:.2f}{}".format(n / 10 ** (3 * millidx), millnames[millidx])
+
+
+def convert_blocks_to_time(blocks: int, block_time: int = 12) -> tuple[int, int, int]:
```
basfroman, line 25: pls add at least single line docstring

basfroman (approved):
LGTM

thealligatorking (approved):

PR #2117, opened 2024-07-11
feat: Enhance dendrite error messaging

thealligatorking:
## Summary 

- Extends error messages with granular status codes and messages.


bittensor/constants.py
```diff
@@ -15,7 +15,29 @@
...
+#### Dendrite ####
+DENDRITE_ERROR_MAPPING: Dict[Type[Exception], tuple] = {
+    aiohttp.ClientConnectorError: ("503", "Service unavailable"),
+    asyncio.TimeoutError: ("408", "Request timeout"),
+    aiohttp.ClientResponseError: (None, "Client response error"),
+    aiohttp.ClientPayloadError: ("400", "Payload error"),
+    aiohttp.ClientError: ("500", "Client error"),
+    aiohttp.ServerTimeoutError: ("504", "Server timeout error"),
+    aiohttp.ServerDisconnectedError: ("503", "Service disconnected"),
+    aiohttp.ServerConnectionError: ("503", "Service connection error"),
```
thewhaleking, line 39: I'm totally fine with the hard-coding here, but just for future reference, the error message is embedded in the docstring, so like `aiohttp.ServerTimeoutError.__doc__`.

tests/unit_tests/test_dendrite.py
```diff
@@ -334,3 +341,75 @@ async def test_dendrite__call__handles_http_error_response(
...
+    exception,
+    expected_status_code,
+    expected_message,
+    synapse_timeout,
+    synapse_ip,
+    synapse_port,
+    request_name,
+):
+    # Arrange
+    dendrite = DendriteMixin()
```
thewhaleking: Are we sure we want to use `DendriteMixin` here rather than the `dendrite` class?

thewhaleking (approved):

tests/unit_tests/test_dendrite.py
thealligatorking: Good catch - updating now.

thewhaleking (approved):

bittensor/constants.py
```diff
@@ -15,7 +15,29 @@
...
+# 3rd Party
+import aiohttp
+
 
 ALLOWED_DELTA = 4_000_000_000  # Delta of 4 seconds for nonce validation
 V_7_2_0 = 7002000
 NANOSECONDS_IN_SECOND = 1_000_000_000
+
+#### Dendrite ####
+DENDRITE_ERROR_MAPPING: Dict[Type[Exception], tuple] = {
```
basfroman, line 31: Why do we store a constant in an external `bittensor/constants.py` module if it is only used in the `bittensor/dendrite.py` module? Let's store it in the same module where it's using.

bittensor/dendrite.py
```diff
@@ -225,7 +229,17 @@ def _get_endpoint_url(self, target_axon, request_name):
         )
         return f"http://{endpoint}/{request_name}"
 
-    def _handle_request_errors(self, synapse, request_name, exception):
+    def log_exception(self, exception: Exception):
```
basfroman, line 232: pls add single line docstring

basfroman (changes requested):

bittensor/dendrite.py
thealligatorking: added

bittensor/constants.py
thealligatorking: this is subjective and shouldn't be a blocking comment.

bittensor/dendrite.py
```diff
@@ -237,22 +263,32 @@ def _handle_request_errors(self, synapse, request_name, exception):
...
+
+        self.log_exception(exception)
+
+        error_info = DENDRITE_ERROR_MAPPING.get(type(exception), DENDRITE_DEFAULT_ERROR)
+        status_code, status_message = error_info
+
+        if status_code:
+            synapse.dendrite.status_code = status_code  # type: ignore
+        elif isinstance(exception, aiohttp.ClientResponseError):
+            synapse.dendrite.status_code = str(exception.code)  # type: ignore
```
basfroman, line 281: `code` is deprecated for exactions. we need to use `status` instead of this in the future.

basfroman (approved):
Left no blocking comment, but `LGTM`

PR #2123, opened 2024-07-12
Merget staging into merge-async for the task https://github.com/opentensor/KnowledgeBase/issues/69

basfroman:
Preperation for the task https://github.com/opentensor/KnowledgeBase/issues/69
- `staging-async` crated from commit `0d11b42182c19d6579fd9ec319fc6a0d79f60243`
- `merge-staging-to-staging-async` created from the same commit
- `staging` merged into `merge-staging-to-staging-async`
- changed bittensor version to `alfa` think as `async` version
- fixed all new sync code to async
- tests checked locally


VERSION
```diff
@@ -1 +1 @@
-7.2.0
\ No newline at end of file
+7.2.0a1
```
ibraheem-abe: I am not sure if this versioning will work with the version checking in  `utils/version`

basfroman: all tests passed and I added some logic in the bittensor.__init__.py for __version__

PR #2126, opened 2024-07-12
Re-adds check_coldkey_swap command

ibraheem-abe:
This re-adds `check_coldkey_swap` command for release 7.3.0


thewhaleking (approved):

ibraheem-abe:
Tested locally after adding keys in arbitration in local subtensor
image

PR #2132, opened 2024-07-13
fix: coldkeypub usage instead of coldkey for arbitration_stats

Rapiiidooo:
### Usage of the coldkeypub instead of coldkey

Change the method that performs a check of the current arbitration data (if any), use the coldkeypub instead of the coldkey.
It increases the performance as the file decryption doesn't need to occur.
It's also more secure as the password should not be required for that stat.

### Quantitative Performance Benefits

Few seconds.

### Verification Process

Re-ran the method and it works correctly.

### Release Notes

- The Arbitration stats check method now use coldkeypub to improve the performance and the security.

### DIFF
--- a/bittensor/commands/check_coldkey_swap.py
+++ b/bittensor/commands/check_coldkey_swap.py
@@ -30,14 +30,16 @@ def fetch_arbitration_stats(subtensor, wallet):
     """
     Performs a check of the current arbitration data (if any), and displays it through the bittensor console.
     """
-    arbitration_check = len(subtensor.check_in_arbitration(wallet.coldkey.ss58_address))
+    arbitration_check = len(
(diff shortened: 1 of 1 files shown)


alex-drocks (approved):

ibraheem-abe:
Hey @Rapiiidooo  
Can you fetch the latest staging and rebase off of it? We back-merged master recently into staging to get all changes.

Rapiiidooo:
Thank you 😄 ! It is all good now @ibraheem-opentensor

camfairchild (approved):

thealligatorking (approved):

ibraheem-abe (approved):
LGTM 👍

PR #2141, opened 2024-07-16
Adds timeout for e2e tests

ibraheem-abe:
In some cases where the e2e tests keep running indefinitely, we need a timeout to halt the tests to preserve credits. 
Currently, we've set 45 minutes - we will be observing if this is good enough with PRs in staging before rolling it out.


thealligatorking (approved):

basfroman (approved):
I think 45 mins is enough for now.
LGTM

PR #2153, opened 2024-07-21
btlogging/loggingmachine.py: Fix bw compat API.

mvds00:
The bt logging mechanism breaks regular logging in various ways, that are solved by this commit.

### Bug

Logging from outside of bittensor is broken, for instance, `retry` logs using:

    logger.warning('%s, retrying in %s seconds...', e, _delay)

and this will output the literal `%s`, as the `args` are hijacked by the attempt at bw compatibility in bittensor, because the args `e` and `_delay` would land on kwargs `prefix` and `suffix`.

Also, all logging originates quite visibly from `loggingmachine.py`, which, in a literal sense, is correct, but not very helpful information.

### Description of the Change

- add `stacklevel=2` so that the filename of the originating call is logged (requires Python >= 3.8)
- move `*args` before `prefix` and `suffix` kwargs so regular logging calls using `args`, such as in `retry`, still work as intended

### Alternate Designs

Reworking the logging entirely is also a possibility. But this seemed to be the easy fix.

### Possible Drawbacks

Any code relying on prefix and suffix also being positional arguments, is now broken. This is intentional.

### Verification Process

I ran code that relies on `bittensor` and `btlogging` to verify that logging is now better.

### Release Notes

- Fixed and improved logging

### DIFF
--- a/bittensor/btlogging/loggingmachine.py
+++ b/bittensor/btlogging/loggingmachine.py
@@ -22,6 +22,7 @@ Default, Debug, Trace, and Disabled.
 """
 
 import argparse
+import platform
 import atexit
 import copy
 import logging as stdlogging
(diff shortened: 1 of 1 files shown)


mvds00:
I pushed an additional commit to get rid of the ` - ` prefixes and suffixes that are added when prefix and/or suffix are not used.

basfroman:
Looks great, but you need to apply this PR to staging. I tried this, but it affects 30 files. Please fix this.

PR #2158, opened 2024-07-22
Ruff complains about an extra line

basfroman:
Somehow the state of the code base was destroyed when the ruff formatter missed a line by mistake. After this we get an error for all subsequent PRs

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -109,7 +109,6 @@ from .utils.subtensor import get_subtensor_errors
 KEY_NONCE: Dict[str, int] = {}
 
 
-
 class ParamWithTypes(TypedDict):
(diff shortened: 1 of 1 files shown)


mvds00:
I don't think ruff made the mistake. The mistake was to add commits after approval, see my Discord post.

ibraheem-abe (approved):

basfroman:
> I don't think ruff made the mistake. The mistake was to add commits after approval, see my Discord post.

You're right. That's exactly what happened. I'll merge the PR and the error will go away.

thewhaleking:
>Somehow the state of the code base was destroyed when the ruff formatter missed a line by mistake.

Need a bit more context here. How did this destroy the codebase?

PR #2159, opened 2024-07-23
fix streaming synapse regression

mjurbanski-reef:
### Bug

Streaming Synapse is broken since https://github.com/opentensor/bittensor/commit/22b77d7c23b1d4f8bd86114777277a9857453464 that got released in 7.2.0

### Description of the Change

Exempt Streaming Synapse from being immediately attempted to be converted into JSON response effectively returing the original behavior 

### Alternate Designs

Reverting the original commit that fixed HTTP error status codes. IMO that would be step backwards therefore I advise against it.
One of the reasons for this regression is that there was not enough test coverage and that commit introduced additional tests, but the Streaming use case was not covered.

### Possible Drawbacks

No that I know of.

### Verification Process

Added tests for axon serving streaming synapse.

### Release Notes

- Fixed StreamingSynapse regression introduced in bittensor 7.2.0


thealligatorking:
@mjurbanski-reef thank you for this. We'll get streaming added to our internal test setup as well.

cc: @ibraheem-opentensor

thewhaleking:
Looks good overall. Just let me do a little more testing. Can you update the branch with the base please?

thewhaleking (approved):

mjurbanski-reef:
Implementation note:
This implementation does not set headers from Synapse value on the streaming response.
After this was noticed, I was asked not to implement it. IMO it probably would be better to do so for backwards compatibility at very least, but in terms of usefullness it was deemed non-critical and returning Synapse tends to have more adverse effects (i.e. can be often used for cheating) than positive ones.

Discord thread: https://discord.com/channels/1120750674595024897/1120799375703162950/1265413339391983748

mjurbanski-reef:
Fixed streaming response headers not being set as they were in the pre 7.2 ; covered that change in unit tests.

thealligatorking (approved):

ibraheem-abe (approved):

PR #2164, opened 2024-07-23
fix(subtensor.set_weights): break early if no wait flags are set

0ximjosh:
### Bug

subtensor.set_weights will try to double fire if neither wait flag is set

### Description of the Change

break early if both wait flags are unset

### Alternate Designs

N/A

### Possible Drawbacks

N/A

### Verification Process

Manually patched in local bittensor package

### Release Notes

- Fixes set_weights throwing `priority too low` when neither wait flags are set

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -759,7 +759,6 @@ class Subtensor:
     ###############
     # Set Weights #
     ###############
-    # TODO: still needed? Can't find any usage of this method.
     def set_weights(
         self,
         wallet: "bittensor.wallet",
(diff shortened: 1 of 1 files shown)


0ximjosh:
tests that failed dont seem to be related to this code change in any capacity

thealligatorking:
@GentikSolm thank you, we will be taking a look here shortly.

bittensor/subtensor.py
```diff
@@ -817,6 +816,8 @@ def set_weights(
                     wait_for_finalization=wait_for_finalization,
                     prompt=prompt,
                 )
+                if not wait_for_finalization and not wait_for_inclusion:
```
thewhaleking, line 819: Would it not be better to break on `if success is True`?

0ximjosh: not sure if that will have the same desired output, but am open to it.

PR #2167, opened 2024-07-25
CI/E2E test improvements

mvds00:
### Bug

CI is broken because of issues in E2E tests. Issues are:
- epoch calculation is off in various ways
- time.sleep() is used in an asyncio context
- wait_for_epoch() was an unused duplicate of wait_for_interval()

This is more of a style / bug waiting to happen thing:
- various constants are hardcoded repeatedly ("magic constants")

### Description of the Change

- epoch calculation is now exact (derived mathematically and verified by comparing epoch block calculations with actual epoch blocks)
- wait_for_epoch() now waits for the epoch without the need to specify the tempo
- tempo is taken from chain instead of hardcoded value
- netuid==1 is now a variable and not a magic constant at several places

The epoch calculation is detailed in the related commit message, with reference to subtensor code.

### Alternate Designs

No alternate designs were considered. Looking forward however, the following is considered:
- using a different netuid should be possible, for example when creating multiple netuids or when testing whether the tempo logic (which takes netuid as a parameter) is correct.
- using a different tempo should be possible, one the one hand to speed up the test, on the other hand to validate all still works with a different tempo than 360. Code is locally available to do this via an extrinsic, but I have to find the root key first. Using a different tempo (for example 12 or 36), by hardcoding it in subtensor, resulted in some tests failing, which might hint at possible issues.

### Possible Drawbacks

N/A

### Verification Process

The E2E CI code was run locally and showed that the errors that were seen before, are now gone. No new errors appeared.

### Release Notes

N/A

### DIFF
--- a/tests/e2e_tests/multistep/test_axon.py
+++ b/tests/e2e_tests/multistep/test_axon.py
@@ -31,12 +31,13 @@ are set correctly, and that the miner is currently running
 
 @pytest.mark.asyncio
 async def test_axon(local_chain):
+    netuid = 1
     # Register root as Alice
     alice_keypair, exec_command, wallet = setup_wallet("//Alice")
     exec_command(RegisterSubnetworkCommand, ["s", "create"])
 
-    # Verify subnet 1 created successfully
-    assert local_chain.query("SubtensorModule", "NetworksAdded", [1]).serialize()
+    # Verify subnet <netuid> created successfully
+    assert local_chain.query("SubtensorModule", "NetworksAdded", [netuid]).serialize()
 
     # Register a neuron to the subnet
     exec_command(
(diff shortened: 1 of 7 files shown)


basfroman:
Hi @mvds00 , appreciate for your PR. At first glance everything looks good, let's update the branch from the staging version and wait for the test to complete. Tomorrow we will explore this in more detail.

Also, could you please use a naming convention for the branches. We have [this document](https://github.com/opentensor/bittensor/blob/master/contrib/DEVELOPMENT_WORKFLOW.md#development-model).
I recommended to use something like this `feature/<your_github_account_name>/<descriptive-isentence>`.

Thank you!

mvds00:
> Hi @mvds00 , appreciate for your PR. At first glance everything looks good, let's update the branch from the staging version and wait for the test to complete. Tomorrow we will explore this in more detail.
> 
> Also, could you please use a naming convention for the branches. We have [this document](https://github.com/opentensor/bittensor/blob/master/contrib/DEVELOPMENT_WORKFLOW.md#development-model). I recommended to use something like this `feature/<your_github_account_name>/<descriptive-isentence>`.
> 
> Thank you!

Well I did use a naming convention: PR (Pull Req) s (staging) 002 (second in line).

But I'll close this one and continue with the renamed branch!

PR #2168, opened 2024-07-25
CI/E2E test improvements

mvds00:
### Bug

CI is broken because of issues in E2E tests. Issues are:
- epoch calculation is off in various ways
- time.sleep() is used in an asyncio context
- wait_for_epoch() was an unused duplicate of wait_for_interval()

This is more of a style / bug waiting to happen thing:
- various constants are hardcoded repeatedly ("magic constants")

### Description of the Change

- epoch calculation is now exact (derived mathematically and verified by comparing epoch block calculations with actual epoch blocks)
- wait_for_epoch() now waits for the epoch without the need to specify the tempo
- tempo is taken from chain instead of hardcoded value
- netuid==1 is now a variable and not a magic constant at several places

The epoch calculation is detailed in the related commit message, with reference to subtensor code.

### Alternate Designs

No alternate designs were considered. Looking forward however, the following is considered:
- using a different netuid should be possible, for example when creating multiple netuids or when testing whether the tempo logic (which takes netuid as a parameter) is correct.
- using a different tempo should be possible, one the one hand to speed up the test, on the other hand to validate all still works with a different tempo than 360. Code is locally available to do this via an extrinsic, but I have to find the root key first. Using a different tempo (for example 12 or 36), by hardcoding it in subtensor, resulted in some tests failing, which might hint at possible issues.

### Possible Drawbacks

N/A

### Verification Process

The E2E CI code was run locally and showed that the errors that were seen before, are now gone. No new errors appeared.

### Release Notes

N/A

### DIFF
--- a/tests/e2e_tests/utils.py
+++ b/tests/e2e_tests/utils.py
@@ -34,6 +34,7 @@ def setup_wallet(uri: str):
             "--wallet.path",
             wallet_path,
         ]
+        logging.info(f'executing command: {command} {" ".join(args)}')
         config = bittensor.config(
             parser=parser,
             args=args,
@@ -123,18 +124,22 @@ def call_add_proposal(substrate: SubstrateInterface, wallet: bittensor.wallet) -
     return response.is_success
 
 
-def wait_interval(interval, subtensor):
+def wait_interval(tempo, subtensor, netuid=1):
+    interval = tempo + 1
     current_block = subtensor.get_current_block()
-    next_tempo_block_start = (current_block - (current_block % interval)) + interval
+    last_epoch = current_block - 1 - (current_block + netuid + 1) % interval
+    next_tempo_block_start = last_epoch + interval
+    last_reported = None
     while current_block < next_tempo_block_start:
         time.sleep(1)  # Wait for 1 second before checking the block number again
         current_block = subtensor.get_current_block()
-        if current_block % 10 == 0:
+        if last_reported is None or current_block - last_reported >= 10:
+            last_reported = current_block
             print(
-                f"Current Block: {current_block}  Next tempo at: {next_tempo_block_start}"
+                f"Current Block: {current_block}  Next tempo for netuid {netuid} at: {next_tempo_block_start}"
             )
             logging.info(
-                f"Current Block: {current_block}  Next tempo at: {next_tempo_block_start}"
+                f"Current Block: {current_block}  Next tempo for netuid {netuid} at: {next_tempo_block_start}"
             )
 
 
--- a/tests/e2e_tests/utils.py
+++ b/tests/e2e_tests/utils.py
@@ -1,4 +1,5 @@
 import logging
+import asyncio
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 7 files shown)


thealligatorking:
@mvds00 thank you for the submission - we're taking a look.

mvds00:
> @mvds00 thank you for the submission - we're taking a look.

Seems like only the swap hotkey fails, but that's simply missing in subtensor and I was requested to leave it as-is.

basfroman:
> > @mvds00 thank you for the submission - we're taking a look.
> 
> Seems like only the swap hotkey fails, but that's simply missing in subtensor and I was requested to leave it as-is.

Yes, the "swap hotkey" test will still fail due to the subtensor. We are waiting for these changes.
Let's wait for the remaining tests to complete and I would like to approve this PR. Also let's merge this PR to https://github.com/opentensor/bittensor/pull/2155.

thewhaleking:
> Also let's merge this PR to #2155.

Why would we do that? This is unrelated. I specifically asked him to create a separate PR because it is unrelated.

thewhaleking (approved):

tests/e2e_tests/utils.py
```diff
@@ -123,18 +125,37 @@ def call_add_proposal(substrate: SubstrateInterface, wallet: bittensor.wallet) -
...
+    tempo = q_tempo[0]
+    logging.info(f"tempo = {tempo}")
+    await wait_interval(tempo, subtensor, netuid)
+
+
+async def wait_interval(tempo, subtensor, netuid=1):
+    interval = tempo + 1
     current_block = subtensor.get_current_block()
-    next_tempo_block_start = (current_block - (current_block % interval)) + interval
+    last_epoch = current_block - 1 - (current_block + netuid + 1) % interval
```
thewhaleking, line 144: Hey man. What's the reason for the netuid here? @ibraheem-opentensor and I were wondering.

mvds00: It's the algorithm as used in subtensor!

mvds00: I'm not making this up ;-)
There is extensive explanation in one of the commit messages, quoting a part from there:

For tempo T, the epoch interval is T+1, and the offset depends on the
netuid. See subtensor/src/block_step.rs, blocks_until_next_epoch(), with
the comment stating:

https://github.com/opentensor/subtensor/blob/1332d077ea73bc7bf40f551c7f1adea3370df2bd/pallets/subtensor/src/block_step.rs#L33
"Networks run their epoch when (block_number + netuid + 1 ) % (tempo + 1) = 0"

This comment from the subtensor code is not correct

mvds00: It only accidentally worked before, and only for the first few epochs before the differences start to matter

basfroman:
> > Also let's merge this PR to #2155.
> 
> 
> 
> Why would we do that? This is unrelated. I specifically asked him to create a separate PR because it is unrelated.

Yeah, not `to` 2155, but `before`. I meant this PR has to be merged first.
@mvds00

tests/e2e_tests/utils.py
thewhaleking: Cool. Assumed you had a good reason for it. Just didn't seem immediately apparent. Thanks.

PR #2191, opened 2024-07-29
Part1: Removing content related to the wallet. Start use the pip installable package.

basfroman:


bittensor/__init__.py
```diff
@@ -1,21 +1,20 @@
 # The MIT License (MIT)
-# Copyright © 2021 Yuma Rao
-# Copyright © 2022-2023 Opentensor Foundation
-# Copyright © 2023 Opentensor Technologies Inc
-
+# Copyright © 2024 OpenTensor Foundation
```
thewhaleking: Opentensor not OpenTensor

bittensor/mock/subtensor_mock.py
```diff
@@ -1,30 +1,31 @@
 # The MIT License (MIT)
-# Copyright © 2022-2023 Opentensor Foundation
-
+# Copyright © 2024 OpenTensor Foundation
```
thewhaleking: Opentensor

bittensor/__init__.py
basfroman: This license has been through so many reviews. No one noticed before.

thealligatorking (approved):

thewhaleking (approved):
I don't want to say "LGTM" because it looks atrocious with the `# type: ignore` comments everywhere, but I understand the need to have them to preserve backwards compatibility. One day, when we're no longer hampered by this, I know you will create a beautiful new design.

PR #2195, opened 2024-08-01
feat: add `register_cost` command for registering a neuron to a subnet

0xxfu:
Executes the ``register_cost`` command to return the cost of registering a neuron on the Bittensor network by recycling some TAO.

    This command is used to determine the registration cost for a neuron on a specified subnet within the network.

    Usage:
        The command verifies the existence of the specified subnet and returns the current recycle amount required for registration. The cost is displayed in TAO, the network's native token.

        The registration cost is essential for users to understand the financial commitment required to add a neuron to the network. It is a critical factor in the decision-making process for contributing to the network.

    Example usage::

        btcli register_cost --netuid 1

    Note:
        This command is useful for users who wish to understand the registration cost before proceeding with the registration process. It provides valuable information for planning and budgeting the registration of a neuron on the network.


thealligatorking:
@0xxfu thank you for the submission - we're taking a look.

thewhaleking:
Sorry about this. The btcli has moved to its own [repo](https://github.com/opentensor/btcli). If you would like to port this feature to the new btcli, we can certainly review that.

PR #2203, opened 2024-08-04
Tests for metagraph, highlighting bug

mvds00:
### Bug

The e2e test that tests metagraph has some flaws. Some assertions don't work as intended and a bug in metagraph is not caught. While I was at it I fixed some low hanging fruit, that is, some fragile constructions that may turn into bugs later on.

### Description of the Change

- don't split lines, assert particular strings in complete output so specific line number doesn't matter
- fix `assert 'a' and 'b' in s` as it doesn't test whether both `'a'` and `'b'` are in `s`
- replace magic constant `1` with `netuid`
- create an additional subnet to make the test coverage broader
- test `.save()` and `.load()` on metagraph

note that the last test fails, but that is due to a bug in bittensor that has to be solved. This PR demonstrates there is a bug, and provides a test to see whether a (to be written) fix solves the bug.

### Alternate Designs

Not relevant.

### Possible Drawbacks

N/A

### Verification Process

By doing `metagraph_pre_dave = subtensor.metagraph(netuid=netuid)` instead of `metagraph_pre_dave.load()` it was verified that the assertions comparing distinct metagraph objects should work, if the metagraphs are in sync.

### Release Notes

N/A

### DIFF
--- a/bittensor/commands/stake.py
+++ b/bittensor/commands/stake.py
@@ -58,7 +58,7 @@ def get_netuid(
         )
         return False, -1
     netuid = cli.config.netuid
-    if netuid < 0 or netuid > 2**32 - 1:
+    if netuid < 0 or netuid > 65535:
         console.print(
             "[red]Invalid input. Please enter a valid integer for netuid in subnet range.[/red]"
         )
--- a/bittensor/commands/stake.py
+++ b/bittensor/commands/stake.py
@@ -44,19 +44,25 @@ MAX_CHILDREN = 5
 
 
 def get_netuid(
-    cli: "bittensor.cli", subtensor: "bittensor.subtensor"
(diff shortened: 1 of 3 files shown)


thealligatorking:
@ibraheem-opentensor could you review when you have time please

ibraheem-abe:
Hey @mvds00 
Thank you for your contribution. The test looks good.
Just waiting if you would like to push your fix regarding the .neurons in save/load - otherwise we can also take the task up!

mvds00:
I will submit a PR.

mvds00:
I rebased the PR on staging, the changes are still relevant. It tests save/load - I saw another PR was posted already fixing that.

thewhaleking (dismissed):

mvds00:
I just rebased it on staging, some tests fail but that is on purpose IIRC.

PR #2209, opened 2024-08-07
test_root_register_add_member_senate.py: improvements

mvds00:
### Bug

The e2e test in `test_root_register_add_member_senate.py` searches for specific strings in specific lines (line numbers) of output. This does not make sense as the line number is not necessarily stable, nor relevant. When testing today the line numbers did not add up (did not investigate why exactly). 

### Description of the Change

As the ordering of items could be relevant, a function `assert_sequence()` is added that check whether expected items occur in the expected order. If lines would be injected between the expected lines, this doesn't affect the assertion. This is by design.

Also, the magic netuid number 1 is replaced by the variable netuid == 1.

### Alternate Designs

It was considered to also assert that lines are adjacent. In practice the `assert_sequence()` will be most useful if injection of additional lines doesn't break the test. Such checks may be added later as a flag. It was considered to place `assert_sequence()` in some `utils.py` but it was decided to keep the change local to `test_root_register_add_member_senate.py` for now.

### Possible Drawbacks

Not expected.

### Verification Process

The new logic was verified by changing the expected string and observing the assertion being triggered.

### Release Notes

N/A

### DIFF
--- a/tests/e2e_tests/subcommands/root/test_root_register_add_member_senate.py
+++ b/tests/e2e_tests/subcommands/root/test_root_register_add_member_senate.py
@@ -13,8 +13,25 @@ from bittensor.commands.senate import SenateCommand
 from ...utils import setup_wallet
 
 
+def assert_sequence(lines, sequence):
+    sequence_ptr = 0
+    for line in lines:
+        words_in_line = set(line.split())
+        current_seq_set = sequence[sequence_ptr]
+        if current_seq_set.issubset(words_in_line):
+            sequence_ptr += 1
+            if sequence_ptr == len(sequence):
+                break
+
+    assert sequence_ptr == len(
+        sequence
+    ), f"Did not find sequence[{sequence_ptr}] = '{sequence[sequence_ptr]}' in output"
+
+
 def test_root_register_add_member_senate(local_chain, capsys):
     logging.info("Testing test_root_register_add_member_senate")
+    netuid = 1
+
     # Register root as Alice - the subnet owner
     alice_keypair, exec_command, wallet = setup_wallet("//Alice")
     exec_command(RegisterSubnetworkCommand, ["s", "create"])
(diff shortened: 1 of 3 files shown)


thealligatorking (dismissed):

tests/e2e_tests/subcommands/root/test_root_register_add_member_senate.py
```diff
@@ -13,8 +13,25 @@
...
+        if lineset.intersection(seqset) == seqset:
+            sequence_ptr += 1
+            if sequence_ptr >= len(sequence):
+                # all items seen
+                break
+    assert sequence_ptr == len(
+        sequence
+    ), f"Did not find sequence[{sequence_ptr}] = '{sequence[sequence_ptr]}' in output"
+
+
```
thewhaleking: Let's maybe make this a bit more readable:

```py
def assert_sequence(lines: list[str], sequence: Collection[set[str]]):
    sequence_ptr = 0
    for line in lines:
        words_in_line = set(line.split())
        current_seq_set = sequence[sequence_ptr]
        if current_seq_set.issubset(words_in_line):
            sequence_ptr += 1
            if sequence_ptr == len(sequence):
                break

    assert sequence_ptr == len(sequence), (
        f"Did not find sequence[{sequence_ptr}] = '{sequence[sequence_ptr]}' in output"
    )
```

thewhaleking (changes requested):
Other than improving the readability of the `assert_sequence` function, this is good to go.

tests/e2e_tests/subcommands/root/test_root_register_add_member_senate.py
mvds00: Other than replacing `>=` with `==`, I agree. I have seen such constructs, where a value has to land on an exact value to break out of a loop, lead to very hard to find bugs.

mvds00: agreed, pushed this

thewhaleking (dismissed):

basfroman:
not relevant with SDK

PR #2221, opened 2024-08-09
Adds e2e setup & tests

ibraheem-abe:
Adds e2e tests to btsdk.

- chain_interactions: Contains code to interact with the chain and submit extrinsics. For functionality not present in btsdk but necessary for tests

Tests added:
- test_axon
- test_dendrite
- test_incentive
- test_transfer

Next PR will include remaining tests


tests/e2e_tests/utils/chain_interactions.py
basfroman, line 1: Pls add module description

basfroman (approved):

PR #2223, opened 2024-08-11
btcli/bt.config: various improvements

mvds00:
### Bug

- You can pass any argument to code that uses `bt.config()` with default kwargs, and it will not complain. This leads to confusion on all levels (developers, devops, users).
- `bt.config()` does some special magic that kills the `required=True` feature of argparse; it will complain that the required argument is not set, even when it is set. This leads to developers implementing workarounds such as setting semi-random default values and/or additional post-argparse checks that could/should already be done in argparse. This bug seems to originate in the recursive nature of argparse as used in bittensor.
- When using `strict=True`, subparsers will wrongly complain about not recognizing arguments like `--no_prompt`

I'm not entirely sure what `bt.config` intends to do at every turn, but I think the changes described below don't make it worse.

### Description of the Change

While fixing the bugs described above, various other small (cosmetic) improvements were done. Taken from the commit message:

- implement recursive apply function for (sub^n)argparser
- use recursive apply to eliminate repeated code
- move adding standard bt.config args to separate function
- \* use recursive apply to add standard bt.config args to all parsers
- \* make strict=True the default
- eliminate redundant COMMANDS.item.name field
- remove obsolete COMMANDS looping code
- remove redundant dest= argument
- fix some comments

The changes marked * implement a functional improvement, the rest is for improved maintainability.

### Alternate Designs

Rework `bt.config`, as it is not clear what it intends to add to argparse, that already supports tons of features.

### Possible Drawbacks

People may experience code not starting anymore because they supply unknown arguments that now lead to errors due to strict checking. This is intended. Having non-functional arguments is not useful and confusing.

### Verification Process

I tested this on some bittensor code I'm writing right now, by adding a `required=True` argument and observing that the argument is now required and that the code doesn't bail out on a supposedly missing argument. I tested to see that I now cannot add `--whatevercrapinsertedhere` without argparse complaining about an unknown arg.

The CI checks further serve as a broad verification of the argparse mechanics.

### Release Notes

Any invalid arguments will now lead to an error, instead of being silently ignored, or treated otherwise depending on your exact implementation. In case you encounter an invalid argument error, as a first step remove the offending argument(s) and see what happens.

### DIFF
--- a/bittensor/config.py
+++ b/bittensor/config.py
@@ -148,7 +148,7 @@ class config(DefaultMunch):
         ## Make a dict with keys as args and values as argparse.SUPPRESS
         defaults_as_suppress = {key: argparse.SUPPRESS for key in all_default_args}
 
-        def local_set_defaults(local_parser:argparse.ArgumentParser):
+        def local_set_defaults(local_parser: argparse.ArgumentParser):
             ## Set the defaults to argparse.SUPPRESS, should remove them from the namespace
             local_parser.set_defaults(**defaults_as_suppress)
             local_parser._defaults.clear()  # Needed for quirk of argparse
@@ -173,7 +173,11 @@ class config(DefaultMunch):
         }
 
     @staticmethod
-    def apply_to_parser_recursive(parser:argparse.ArgumentParser, callback:Callable[[argparse.ArgumentParser],None], depth:int=0):
+    def apply_to_parser_recursive(
+        parser: argparse.ArgumentParser,
+        callback: Callable[[argparse.ArgumentParser], None],
+        depth: int = 0,
+    ):
         """
         Recursively apply callback() to parser and its subparsers.
         """
@@ -187,7 +191,7 @@ class config(DefaultMunch):
                 config.apply_to_parser_recursive(cmd_parser, callback, depth=depth + 1)
 
     @staticmethod
-    def add_args(parser:argparse.ArgumentParser):
+    def add_args(parser: argparse.ArgumentParser):
         """
         Add standard arguments to argument parser.
         """
--- a/bittensor/config.py
+++ b/bittensor/config.py
@@ -26,7 +26,7 @@ import yaml
 import copy
 from copy import deepcopy
 from munch import DefaultMunch
-from typing import List, Optional, Dict, Any, TypeVar, Type
+from typing import List, Optional, Dict, Any, TypeVar, Type, Callable
 import argparse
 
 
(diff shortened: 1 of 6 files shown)


bittensor/config.py
```diff
@@ -231,6 +172,63 @@ def __init__(
...
+        if not parser._subparsers:
+            return
+        for action in parser._subparsers._actions:
+            if not isinstance(action, argparse._SubParsersAction):
+                continue
+            for cmd_parser in action.choices.values():
+                config.apply_to_parser_recursive(cmd_parser, callback, depth=depth + 1)
+
+    @staticmethod
+    def add_standard_args(parser):
```
basfroman: To keep a consistent naming of adding arguments across the entire bittensor, pls rename`add_standard_args` to `add_args`.

bittensor/config.py
```diff
@@ -182,37 +139,21 @@ def __init__(
...
-                                    cmd_parser: argparse.ArgumentParser
-                                    for cmd_parser in action.choices.values():
-                                        cmd_parser.set_defaults(**defaults_as_suppress)
-                                        cmd_parser._defaults.clear()  # Needed for quirk of argparse
-                        else:
-                            cmd_parser.set_defaults(**defaults_as_suppress)
-                            cmd_parser._defaults.clear()  # Needed for quirk of argparse
+
+        def l_set_defaults(l_parser):
+            ## Set the defaults to argparse.SUPPRESS, should remove them from the namespace
```
basfroman, line 152: Convert this to a dostring. Also, please make the function name `l_set_defaults` a little more descriptive. Example: `load_set_defaults`.
Also, add type annotation.

bittensor/config.py
```diff
@@ -231,6 +172,63 @@ def __init__(
             ]
         }
 
+    @staticmethod
+    def apply_to_parser_recursive(parser, callback, depth=0):
```
basfroman: Pls add single line docstring

basfroman (changes requested):
Left a few comments.
Overall looks good!

mvds00:
Fixed first round of comments, for now as a separate commit. Also fixed two issues that came up in CI, as we are now seeing previously buried issues pop up due to strict checking. Some unknown `--wallet.name` issues still under investigation.

basfroman:
> Fixed first round of comments, for now as a separate commit. Also fixed two issues that came up in CI, as we are now seeing previously buried issues pop up due to strict checking. Some unknown `--wallet.name` issues still under investigation.

Nice. Could you pls fix ruff. I would like to see that all checkers passed.
Thank you!

thewhaleking:
So this is definitely an improvement, but config and flags in general are changed drastically with the new btcli (soon to release fully): https://github.com/opentensor/btcli

PR #2226, opened 2024-08-13
Feature/mvds00/test axon subprocess error checking

mvds00:
### Bug

While working with the e2e tests, it appeared that the miner code did not function well, but the test never gave any indication of this. The two issues that this PR fixes are:
- the miner code is not run with NEST_ASYNCIO disabled, leading to errors generated by uvicorn, that is used by class axon of bittensor.
- `test_axon.py` doesn't check any output of the miner code.
Although the miner code gets far enough to register an axon, making the test succeed, going forward it is not practical to not have everything working flawlessly under the hood.

### Description of the Change

The subprocess to run a miner is now started with `asyncio.create_subprocess_exec` instead of `asyncio.create_subprocess_shell` so that it can be easily terminated by the test, to examine the output.
The status of the subprocess is monitored. The NEST_ASYNCIO environment variable is set to 0 to disable nest_asyncio and fix the miner.

### Alternate Designs

Rework of the test is actively worked on, this will be a separate PR on top of these commits.

### Possible Drawbacks

Not known.

### Verification Process

The test was run and with additional debugging it was observed that the miner now runs better.

### Release Notes

N/A

### DIFF
--- a/tests/e2e_tests/multistep/test_axon.py
+++ b/tests/e2e_tests/multistep/test_axon.py
@@ -1,5 +1,6 @@
 import asyncio
 import sys
+import textwrap
 
 import pytest
 
@@ -28,6 +29,8 @@ Verify that:
 are set correctly, and that the miner is currently running
 
 """
+
+INDENT = "   "
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


mvds00:
fixed ruff formatting issue

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -68,35 +71,52 @@ async def test_axon(local_chain):
...
+    axon_process = await asyncio.create_subprocess_exec(
+        sys.executable,
+        *args,
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
+        env={"NEST_ASYNCIO": "0"},
     )
-    logging.info("Neuron Alice is now mining")
+    await asyncio.sleep(1)  # wait for the miner to start up or fail
+    if axon_process.returncode != None:
```
thewhaleking, line 101: `is not` instead of `!=`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -68,35 +71,52 @@ async def test_axon(local_chain):
...
+        stderr = textwrap.indent(stderr.decode("UTF8"), INDENT)
+        assert (
+            axon_process.returncode == None
+        ), f"miner process stopped immediately or did not start:\n{cmd}\nstdout:\n{stdout}\nstderr:\n{stderr}"
+
+    logging.info(f"Neuron Alice is now mining: {axon_process}")
     await asyncio.sleep(
         5
     )  # wait for 5 seconds for the metagraph to refresh with latest data
+    if axon_process.returncode != None:
```
thewhaleking, line 113: `is not` instead of `!=`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -68,35 +71,52 @@ async def test_axon(local_chain):
...
+        env={"NEST_ASYNCIO": "0"},
     )
-    logging.info("Neuron Alice is now mining")
+    await asyncio.sleep(1)  # wait for the miner to start up or fail
+    if axon_process.returncode != None:
+        stdout, stderr = await axon_process.communicate()
+        stdout = textwrap.indent(stdout.decode("UTF8"), INDENT)
+        stderr = textwrap.indent(stderr.decode("UTF8"), INDENT)
+        assert (
+            axon_process.returncode == None
```
thewhaleking, line 106: `is` instead of `==`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -68,35 +71,52 @@ async def test_axon(local_chain):
...
+    logging.info(f"Neuron Alice is now mining: {axon_process}")
     await asyncio.sleep(
         5
     )  # wait for 5 seconds for the metagraph to refresh with latest data
+    if axon_process.returncode != None:
+        stdout, stderr = await axon_process.communicate()
+        stdout = textwrap.indent(stdout.decode("UTF8"), INDENT)
+        stderr = textwrap.indent(stderr.decode("UTF8"), INDENT)
+        assert (
+            axon_process.returncode == None
```
thewhaleking, line 118: `is` instead of `==`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -109,4 +129,29 @@ async def test_axon(local_chain):
     assert updated_axon.port == 8091
     assert updated_axon.hotkey == alice_keypair.ss58_address
     assert updated_axon.coldkey == alice_keypair.ss58_address
+
+    # Terminate miner and make sure nothing bad happened.
+    if axon_process.returncode == None:
```
thewhaleking, line 134: `is`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -109,4 +129,29 @@ async def test_axon(local_chain):
...
+        logging.info("terminating miner")
+        try:
+            # try/except to prevent exception in case the process just ends now
+            axon_process.terminate()
+        except:
```
thewhaleking, line 139: Let's not do bare `except`s. Ideally, we should know what we're trying to catch, but if not, at least `except Exception as e` and log it.

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -109,4 +129,29 @@ async def test_axon(local_chain):
...
+            pass
+    await asyncio.sleep(1)
+
+    if axon_process.returncode == None:
```
thewhaleking, line 143: `is`

tests/e2e_tests/multistep/test_axon.py
```diff
@@ -109,4 +129,29 @@ async def test_axon(local_chain):
...
+        logging.info("killing miner")
+        try:
+            # try/except to prevent exception in case the process just ends now
+            axon_process.kill()
+        except:
```
thewhaleking, line 148: bare `except`

thewhaleking (changes requested):
Few changes. I like the idea of this test, but I really do want nest_asyncio to completely go away in the future.

basfroman:
No activity for over 2 weeks

PR #2227, opened 2024-08-13
bittensor/axon.py: thread and exception handling

mvds00:
### Bug

Various issues were encountered trying to run and understand e2e tests:
- if uvicorn fails to start, an uncaught exception is emitted to stderr
- axon keeps spinning, waiting for `self.started`, indefinitely
- exceptions are not propagated from threads
- there is no way to (simply) test from the outside whether an axon started and/or runs
- axon creates a thread that only creates another thread, which seems redundant

### Description of the Change

This patch addresses some of these issues, in `FastAPIThreadedServer`:
- add thread safe `set`/`get_exception()` to set/get exceptions
- `run_in_thread()` yields the created thread, so that the code using it can check whether the thread is alive
- `uvicorn.Server.startup()` is wrapped to set a thread-safe flag using `self.set_started(True)` to indicate startup succeeded
- `run_in_thread()` times out after one second to prevent infinite loop in case `self.get_started()` never becomes `True`
- `run_in_thread()` raises an exception if it fails to start the thread
- `_wrapper_run()` tests whether the thread is still alive

and in `class axon`, the following are added:
- `@property axon.exception()`, returning any exception
- `axon.is_running()`, returning `True` when the axon is operational

The seemingly redundant thread is left in until feedback is received on the reasons for including it.

### Alternate Designs

It is a deliberate choice to add status calls on `class axon`, and not expect the user instantiating an axon to look into implementation details such as `axon.fast_server`, even though they are not explicitly marked private by the naming convention used.

### Possible Drawbacks

Not expected. Perhaps issues that are currently buried will suddenly pop up. This is then intended and an expected effect of improving error handling.

### Verification Process

These changes are part of an effort to improve e2e tests, and as such they are part of various other developments and tested along with other changes.

### Release Notes

N/A


bittensor/axon.py
```diff
@@ -100,26 +101,72 @@ class FastAPIThreadedServer(uvicorn.Server):
...
 
+    """
+    Provide a channel to signal exceptions from the thread to our caller.
+    """
+    _exception: Exception = None
+    _lock: threading.Lock = threading.Lock()
+    _thread: threading.Thread = None
+    _started: bool = False
+
+    def set_exception(self, ex):
```
basfroman: Pls use more descriptive variable name instead of `ex`.
Also, pls add type annotation.

bittensor/axon.py
```diff
@@ -100,26 +101,72 @@ class FastAPIThreadedServer(uvicorn.Server):
...
+        with self._lock:
+            self._exception = ex
+
+    def get_exception(self):
+        with self._lock:
+            return self._exception
+
+    def set_thread(self, thread):
```
basfroman: Pls add type annotation.

bittensor/axon.py
```diff
@@ -100,26 +101,72 @@ class FastAPIThreadedServer(uvicorn.Server):
...
+        with self._lock:
+            self._thread = thread
+
+    def get_thread(self):
+        with self._lock:
+            return self._thread
+
+    def set_started(self, started):
```
basfroman: Pls add type annotation.

bittensor/axon.py
```diff
@@ -405,6 +458,24 @@ def info(self) -> "bittensor.AxonInfo":
             placeholder2=0,
         )
 
+    # Our instantiator should be able to test axon.exception to see if any
+    # exception occurred.
+    @property
+    def exception(self):
+        # for future use: setting self._exception to signal an exception
+        e = getattr(self,'_exception',None)
```
basfroman: Pls use more descriptive variable name instead of `e`

bittensor/axon.py
```diff
@@ -100,26 +101,72 @@ class FastAPIThreadedServer(uvicorn.Server):
...
+            thread: a running thread
+
+        Raises:
+            Exception: in case the server did not start (as signalled by self.get_started())
         """
         thread = threading.Thread(target=self.run, daemon=True)
         thread.start()
         try:
-            while not self.started:
+            t0 = time.time()
```
basfroman: Pls use more descriptive variable name instead of `t0`

bittensor/axon.py
```diff
@@ -405,6 +458,24 @@ def info(self) -> "bittensor.AxonInfo":
...
+        # for future use: setting self._exception to signal an exception
+        e = getattr(self,'_exception',None)
+        if e:
+            return e
+        return self.fast_server.get_exception()
+
+    # Our instantiator should be able to test axon.is_running() to see if all
+    # required threads etc are running.
+    def is_running(self):
+        t = self.fast_server.get_thread()
```
basfroman: Pls use more descriptive variable name instead of `t`.

bittensor/axon.py
```diff
@@ -405,6 +458,24 @@ def info(self) -> "bittensor.AxonInfo":
...
+    # exception occurred.
+    @property
+    def exception(self):
+        # for future use: setting self._exception to signal an exception
+        e = getattr(self,'_exception',None)
+        if e:
+            return e
+        return self.fast_server.get_exception()
+
+    # Our instantiator should be able to test axon.is_running() to see if all
```
basfroman: It's better to keep it in docstring. Then it will help not only the developer who reads the code, but also the one who calls `help(function/method/property)`

bittensor/axon.py
```diff
@@ -405,6 +458,24 @@ def info(self) -> "bittensor.AxonInfo":
             placeholder2=0,
         )
 
+    # Our instantiator should be able to test axon.exception to see if any
```
basfroman: It's better to keep it in docstring. Then it will help not only the developer who reads the code, but also the one who calls `help(function/method/property)`

basfroman (changes requested):
Left a few comments for changes.  Also it required ruff formatting. 
But looks pretty good!

mvds00:
Pushed fixes for first round of feedback, for now as a separate commit, plus some further tweaks.

thewhaleking (dismissed):
I'm good with these changes. Just rebase, and we'll merge it in.

thewhaleking (dismissed):

thewhaleking (approved):
Fixed merge conflicts in #2459

basfroman:
No activity for over 2 weeks

PR #2229, opened 2024-08-13
Child hotkeys handle excess normalization

opendansor:
In staking.py, introduced the floor function for normalized proportions and added logic to handle excess sums. Updated formatting.py to use floor for float-to-u64 conversion, ensuring values stay within bounds.

### DIFF
--- a/bittensor/extrinsics/staking.py
+++ b/bittensor/extrinsics/staking.py
@@ -1,6 +1,7 @@
 # The MIT License (MIT)
 # Copyright © 2021 Yuma Rao
 # Copyright © 2023 Opentensor Foundation
+from math import floor
 
 # Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
 # documentation files (the “Software”), to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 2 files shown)


bittensor/extrinsics/staking.py
```diff
@@ -649,4 +650,20 @@ def normalize_children_and_proportions(
...
+
+    # if the sum is more, subtract the excess from the first child
+    if sum_norm > u64_max:
+        if abs(sum_norm - u64_max) > 10:
+            raise ValueError(
+                "The sum of normalized proportions is out of the acceptable range."
+            )
+        normalized_children[0] = (
+            normalized_children[0][0] - (sum_norm - (u64_max - 1)),
+            normalized_children[0][1],
```
basfroman, line 666: It would be useful to understand what such indexes mean. Pls add a comment.

basfroman (approved):

bittensor/extrinsics/staking.py
opendansor: There is a comment on line 658. 
It adjusts the first child +- 10 to make sure we dont go over u64 max.

Issue #2208, opened 2024-08-07
Subtensor.get_subnets excludes subnets past netuid=100 


camfairchild:
### Describe the bug

The `subtensor.get_subnets` function returns *at most* 100 subnets, regardless of the true subnet count.

### To Reproduce

```python
import bittensor

sub = bittensor.subtensor("wss://test.chain.opentensor.ai:443")
subnets = sub.get_subnets()
n_subnets = sub.get_total_subnets()
len(subnets) == n_subnets
# False
```

### Expected behavior

```python
import bittensor

sub = bittensor.subtensor("wss://test.chain.opentensor.ai:443")
subnets = sub.get_subnets()
n_subnets = sub.get_total_subnets()
len(subnets) == n_subnets
# True
```

### Environment

Bittensor 7.3.0

### Additional context

This seems to happen because the function uses the `subtensor.query_map_subtensor("NetworksAdded", block)` which only pulls map entries a page at a time, with the default page size of `100`. 

We should expect it to pull *all* pages for this function.

This potentially affects other functions that use the maps in the same way.

niljub:
Working on this issue :)

PR #2240, opened 2024-08-15
Fixed a bug in the get_subnets function

niljub:
### Bug

The get_subnets function was truncating to max. 100 results due to list comprehensions handling of iterators. The bug has been described in https://github.com/opentensor/bittensor/issues/2208.

### Description of the Change

Replaced list comprehension with a manual iteration to ensure all subnet UIDs are retrieved, addressing the issue of truncated results.

### Alternate Designs

Considered increasing fetch limits but chose manual iteration for more reliable results.

### Possible Drawbacks

No significant drawbacks, just a minor increase in code complexity.

### Verification Process

Added a test to check the length of the result list, ensuring all records are processed correctly. Ran all existing tests to confirm no regressions.

### Release Notes

Fixed an issue in get_subnets where the result list was being truncated, ensuring complete retrieval of subnets.


tests/unit_tests/test_subtensor.py
```diff
@@ -1914,6 +1914,24 @@ def test_get_subnets_no_block_specified(mocker, subtensor):
     subtensor.query_map_subtensor.assert_called_once_with("NetworksAdded", None)
 
 
+def test_get_subnets_correct_length(mocker, subtensor):
+    """Test get_subnets returns a list of the correct length."""
+    # Prep
+    block = 123
+    num_records = 50
```
camfairchild: Shouldn't this test for length > 100?

camfairchild:
Thanks for the contribution! 
How was this fix tested? 
The issue can be reproduced on the testnet and I don't believe this fixes it.

See example:
```python
import bittensor
subtensor = bittensor.subtensor(network="wss://test.chain.opentensor.ai:443")
s = subtensor.get_subnets()
len(s)
# 100; Expected > 100
```

I would consider using the paging functionality of pysub-interface

tests/unit_tests/test_subtensor.py
niljub: You're right—I missed a zero. I've added it and pushed a changed version along with the updated test, but the test still doesn't catch the issue. I also updated get_subnets again and verified it by running your snippet.

        
```python
import bittensor

sub = bittensor.subtensor("wss://test.chain.opentensor.ai:443")
subnets = sub.get_subnets()
n_subnets = sub.get_total_subnets()

len(subnets) == n_subnets
# True; Expected: True
```

niljub:
> Thanks for the contribution! How was this fix tested?

I initially tested the fix using your snippet from the bug report issue. However, during the cleanup process, I introduced an issue that wasn't caught even with 500 networks. I've now pushed a working version and thoroughly tested it with the snippet, though the solution is a bit more barebones.

> I would consider using the paging functionality of pysub-interface.

I can certainly rewrite the function to use pysub-interface. However, this would also require modifying the query_map_subtensor function and incorporating a start_key. This way, we can leverage the substrate to retrieve results page by page as QueryMapResults. If this approach is preferred, let me know and I will implement it.

thewhaleking:
Hi. Taking a look at this today.

thewhaleking (changes requested):
I think this is a great feature to add. Thanks!

But the logic used in the test is bad. You can see that the old function logic passes the new test:
Screenshot 2024-08-28 at 19 20 40

TL;DR: improve the tests please.

basfroman:
implemented in SDK

PR #2244, opened 2024-08-16
feat: return error message instead of raising exception

thealligatorking:
## Overview
- Previously we would raise an exception and stdout the whole stack trace, this update makes enables a terse error messages.

Ex: 

image

### DIFF
--- a/bittensor/commands/delegates.py
+++ b/bittensor/commands/delegates.py
@@ -752,7 +752,13 @@ class NominateCommand:
 
         # Unlock the wallet.
         wallet.hotkey
-        wallet.coldkey
+        try:
+            wallet.coldkey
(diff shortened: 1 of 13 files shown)


thewhaleking (commented):
In the instances here where we also unlock the hotkey, do we want to include those in the try block?

ibraheem-abe:
> In the instances here where we also unlock the hotkey, do we want to include those in the try block?

Interesting thought - but confirming if we actually unlock the hotkey currently. afaik, hotkey is without password in our current setup

ibraheem-abe (approved):
LGTM. 
I think we dont password protect our hotkeys but we can confirm that

opendansor (approved):

PR #2252, opened 2024-08-20
Increase test coverage for subtensor.py

basfroman:
PyCharm says that current `subtensor.py` coverage is 80%

### DIFF
--- a/bittensor/core/subtensor.py
+++ b/bittensor/core/subtensor.py
@@ -1839,6 +1839,7 @@ class Subtensor:
             return None
         return Balance.from_rao(result.value)
 
+    # Community uses this method
     def commit_weights(
         self,
         wallet: "Wallet",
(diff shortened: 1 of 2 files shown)


bittensor/core/subtensor.py
```diff
@@ -1915,6 +1916,7 @@ def commit_weights(
 
         return success, message
 
+    # Community uses this method
```
garrett-opentensor, line 1919: Should we include this comment in the code? If so, perhaps something more subtle like “function used for XYZ purpose”

basfroman: > Should we include this comment in the code? If so, perhaps something more subtle like “function used for XYZ purpose”

I don't think so we need to move the comments like this into the methods/functions. I keep it outside especially for us/developers. It means that marked method is used for backwards compatibility  only (not only within bittensor). 
I have a plan to delete all of same comments when SDK is ready for release.

thealligatorking (approved):

bittensor/core/subtensor.py
garrett-opentensor: ok sounds good

PR #2262, opened 2024-08-22
Increase Bittensor SDK test coverage

basfroman:


bittensor/utils/__init__.py
```diff
@@ -263,70 +159,7 @@ def format_error_message(error_message: dict) -> str:
...
-                try:
-                    info_dictionary[key] = bytes.fromhex(item[2:]).decode()
-                except UnicodeDecodeError:
-                    print(f"Could not decode: {key}: {item}")
-            else:
-                info_dictionary[key] = item
-    return info_dictionary
-
-
+# Subnet 24 uses this function
```
thealligatorking, line 162: Is this needed?

basfroman: > Is this needed?

I'd like to leave this here. I marked this function because it is not used in bittensor, but we should keep it because of subnet 24. Does it make sense?

tests/unit_tests/utils/test_balance.py
```diff
@@ -507,3 +508,18 @@ def test_balance_eq_invalid_type(balance: Union[int, float]):
...
+
+
+def test_from_float():
+    """Tests from_float method call."""
+    assert Balance.from_tao(1.0) == Balance(1000000000)
+
+
+def test_from_tao():
+    """Tests from_tao method call."""
+    assert Balance.from_tao(1.0) == Balance(1000000000)
```
thealligatorking: Aren't these the same?

basfroman: Yeah. Thanks. Removed.

thealligatorking (approved):

PR #2267, opened 2024-08-24
add neuron certificate discovery

andreea-popescu-reef:
### Requirements for Adding, Changing, or Removing a Feature

subtensor migrated to version that handles certificate discovery

### Description of the Change

Adds the ability to send neuron certificate and receive other neuron's certificates. This change will enable setting up mutual tls between neurons

`serve_axon` function takes an additional optional parameter certificate
if certificate is present the `serve_axon_tls` rpc call will be called on the subtensor and the neuron's certificate will be discoverable through the subtensor. The function will behave the same other than the possible storing of certificate.

`get_neuron_certificate(netuid, neuron_uid)` will fetch from the subtensor a specific neuron's certificate.

### Alternate Designs

save the certificate as a field in the axon.
This can be implemented either:
- on subtensor level, extend the axon to have a certificate field in both bittensor library and subtensor logic. Then, the certificate will just be passed along with the axon data in current `serve_axon`, `get_neuron_info` and other function. However this would break compatibility for multiple rpc calls encoding and would need patching.
- on just bittensor library level, the bittensor library axon model would have a certificate field that would be stripped out on rpc calls to the subtensor, and similarly added as needed from separate rpc calls to the subtensor. However this obscures the axon's real data and adds complexity.

### Possible Drawbacks

none?

### Verification Process

TODO

### Release Notes

- tls certificates can now be passed via the `serve_axon` function and fetched via the `get_neuron_certificate` function


bittensor/core/chain_data/neuron_certificate.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import List
+
+from bittensor.core.chain_data.utils import from_scale_encoding, ChainDataType
+from bittensor.utils import Certificate
+
+
+# Dataclasses for chain data.
+@dataclass
+class NeuronCertificate:
+    r"""
```
thewhaleking:
```suggestion
    """
```

bittensor/core/chain_data/neuron_certificate.py
```diff
@@ -0,0 +1,20 @@
...
+    Dataclass for neuron certificate.
+    """
+
+    certificate: Certificate
+
+    @classmethod
+    def from_vec_u8(cls, vec_u8: List[int]) -> "NeuronCertificate":
+        r"""Returns a NeuronCertificate object from a ``vec_u8``."""
```
thewhaleking:
```suggestion
        """Returns a NeuronCertificate object from a ``vec_u8``."""
```

bittensor/core/subtensor.py
```diff
@@ -1170,6 +1172,37 @@ def get_neuron_for_pubkey_and_subnet(
...
+            name="NeuronCertificates",
+            block=block,
+            params=[netuid, hotkey],
+        )
+        if not hasattr(certificate, "serialize"):
+            return None
+        certificate = certificate.serialize()
+        if not certificate:
+            return None
+        return certificate.get("certificate", None)
```
thewhaleking: This is a bit confusing to follow at first glance. I prefer something like this.

```suggestion
        try:
            serialized_certificate = certificate.serialize()
            if serialized_certificate:
                return serialized_certificate.get("certificate", None)
        except AttributeError:
            return None
```

tests/unit_tests/test_subtensor.py
```diff
@@ -27,13 +27,24 @@
...
 
+def get_fake_call_params():
+    return {
+        "version": "1.0",
+        "ip": "0.0.0.0",
+        "port": 9090,
+        "ip_type": 4,
+        "netuid": 1,
+        "certificate": None,
+    }
```
thewhaleking: Any reason not to do this as a pytest parameter?

tests/unit_tests/test_subtensor.py
```diff
@@ -1440,11 +1453,52 @@ def test_do_serve_axon_is_success(subtensor, mocker):
...
+    )
+
+    subtensor.substrate.submit_extrinsic.assert_called_once_with(
+        subtensor.substrate.create_signed_extrinsic.return_value,
+        wait_for_inclusion=fake_wait_for_inclusion,
+        wait_for_finalization=fake_wait_for_finalization,
+    )
+
+    subtensor.substrate.submit_extrinsic.return_value.process_events.assert_called_once()
+    assert result == (True, None)
```
thewhaleking:
```suggestion
    assert result[0] is True
    assert result[1] is None
```

It's important to do comparisons like this because a result of `(1, None)` would also pass this assert.

thewhaleking (changes requested):
One suggestion, and a couple small nitpicks. Overall LGTM.

thewhaleking (dismissed):

thewhaleking (approved):

bittensor/core/extrinsics/serving.py
```diff
@@ -124,6 +136,7 @@ def serve_extrinsic(
         "protocol": protocol,
         "placeholder1": placeholder1,
         "placeholder2": placeholder2,
+        "certificate": certificate,
```
zyzniewski-reef, line 139: Won't this make `neuron_up_to_date` always `False`? Just not sure if that's intended behavior.

PR #2269, opened 2024-08-27
Add substrate reconnection logic

basfroman:
When the error `[Errno 32] Broken pipe` occurs after the connection has been idle, the connection check `substrate.websocket.connected` and `substrate.websocket.sock` still return `True`.
In order to get the real connection state, I used a low-level socket check via `substrate.websocket.sock.getsockopt(socket.SOL_SOCKET, socket.SO_ERROR)`.

Also, I've fixed and new tests.

### DIFF
--- a/bittensor/core/subtensor.py
+++ b/bittensor/core/subtensor.py
@@ -24,6 +24,7 @@ import argparse
 import copy
 import socket
 import sys
+from functools import wraps
 from typing import List, Dict, Union, Optional, Tuple, TypedDict, Any
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


bittensor/core/subtensor.py
```diff
@@ -187,7 +204,26 @@ def __init__(
                 "To get ahead of this change, please run a local subtensor node and point to it."
             )
 
-        # Attempt to connect to chosen endpoint. Fallback to finney if local unavailable.
+        self.log_verbose = log_verbose
+        self._get_substrate()
+
+    def __str__(self) -> str:
+        if self.network == self.chain_endpoint:
+            # Connecting to chain endpoint without network known.
```
thewhaleking, line 212: Wording.

thewhaleking (approved):
This is fine for now, but absolutely this should have better tests before release. I would also like to see the wording improved as above.

PR #2286, opened 2024-09-05
BTSDK: Move `do*` methods to related extrinsic

basfroman:
- all calls to the network are consolidated in related extrinsiсs
- improved `bittensor.utils.networking.ensure_connected` logic and description
- removed unused calls and methods from subtensor
- all extrinsics returns formatter subtensor errors


bittensor/core/extrinsics/transfer.py
```diff
@@ -15,19 +15,81 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from typing import Union, TYPE_CHECKING
+from typing import Dict, Tuple, Optional, Union, TYPE_CHECKING
```
thewhaleking, line 18: You don't need `Dict` or `Tuple`.

tests/e2e_tests/test_incentive.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-import bittensor
+# import bittensor
```
thewhaleking: Can this line just be entirely removed?

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -37,6 +105,7 @@ def commit_weights_extrinsic(
     """
     Commits a hash of the neuron's weights to the Bittensor blockchain using the provided wallet.
     This function is a wrapper around the `_do_commit_weights` method, handling user prompts and error messages.
```
thewhaleking: Update docstring.

thewhaleking (changes requested):
Only blocker is the docstring update. Others are just questions/tips.

tests/e2e_tests/test_incentive.py
basfroman: deferentially. good catch!

bittensor/core/extrinsics/commit_weights.py
basfroman: done

bittensor/core/extrinsics/transfer.py
basfroman: I need this because of the linter doesn't want to see dict and tuple instead of Dict and Tuple as type annotation. It's complaints.

thewhaleking (approved):

PR #2287, opened 2024-09-05
removed exit sys call for ConnectionRefusedError in _get_substrate

garrett-opentensor:
Removes the sys exit call when catching a ConnectionRefusedError in the _get_substrate() method of Subtensor.

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -294,7 +294,6 @@ class Subtensor:
             )
             return
 
-
         try:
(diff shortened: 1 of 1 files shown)


camfairchild (dismissed):

ibraheem-abe (dismissed):

basfroman (changes requested):
pls fix ruff formatter

basfroman (approved):

PR #2288, opened 2024-09-06
BTSDK: removed exit sys call for ConnectionRefusedError in _get_substrate

basfroman:


bittensor/core/subtensor.py
```diff
@@ -223,6 +222,11 @@ def _get_substrate(self):
...
                 )
 
+            try:
+                self.substrate.websocket.settimeout(self._connection_timeout)
+            except (AttributeError, TypeError, socket.error, OSError) as e:
+                logging.warning(f"Error setting timeout: {e}")
+
         except ConnectionRefusedError:
             logging.error(
                 f"Could not connect to {self.network} network with {self.chain_endpoint} chain endpoint. Exiting...",
```
thewhaleking: You're gonna want to change this line since we're no longer exiting.

thewhaleking (changes requested):
Shouldn't we throw an Exception?

basfroman:
> Shouldn't we throw an Exception?

I don't think so. Just warning

bittensor/core/subtensor.py
basfroman: > You're gonna want to change this line since we're no longer exiting.

log message is changed

thewhaleking:
> > Shouldn't we throw an Exception?
> 
> I don't think so. Just warning

Doesn't that just pass the buck down the line, and will bring a new exception when a request is attempted on a closed connection?

basfroman:
> Doesn't that just pass the buck down the line, and will bring a new exception when a request is attempted on a closed connection?

No, if the connection data was passed correctly, then even with a missing connection, the reconnection logic will create a new connection and set the timeout again. Otherwise, the user will receive a warning again.

garrett-opentensor (approved):

thewhaleking (approved):

PR #2308, opened 2024-09-14
fix regression where COLD_PW can no longer be passed with env vars co…

mjurbanski-reef:
## Bug

-

### Description of the Change

Apparently, in the context of python it was possible to set a env var with hyphen in it. Which makes change introduced by https://github.com/opentensor/bittensor/pull/1949 a breaking one which is not acceptable as it was introduced in minor version of bittensor, hence breaking semver contract.

This PR makes sure we support proper env variable names, while also not breaking users that used hyphens within the env var name.

See https://discord.com/channels/1120750674595024897/1242999357436071956/1284309400751964271 for more context.
 
### Alternate Designs

Ignore the regression, but that conflicts with semver versioning and further discourages keeping with the latest bittensor version.

### Possible Drawbacks

Not forcing users to migrate asap.

### Verification Process

Unit test

### Release Notes

* fix regression which broke passing passcode through env var with hyphen, e.g. `BT_COLD_PW_My-Wallet3=pw`. Please note this form is deprecated, and one should use `BT_COLD_PW_My_Wallet3=pw` instead.


alex-drocks (approved):
Looks good to me 🙏

tests/unit_tests/test_keyfile.py
```diff
@@ -615,10 +615,12 @@ def test_get_coldkey_password_from_environment(monkeypatch):
         "WALLET": "password",
         "my_wallet": "password2",
         "my-wallet": "password2",
+        "my-wallet3": "password3",
     }
 
     monkeypatch.setenv("bt_cold_pw_wallet", password_by_wallet["WALLET"])
     monkeypatch.setenv("BT_COLD_PW_My_Wallet", password_by_wallet["my_wallet"])
+    monkeypatch.setenv("BT_COLD_PW_My-Wallet3", password_by_wallet["my-wallet3"])
```
ppolewicz, line 623: won't the new warning crash this?

mjurbanski-reef: CI is passing. Test are not run with `pytest -W error` if this is what you are referring to.

Or did I misunderstand you and you mean something else?

bittensor/keyfile.py
```diff
@@ -282,7 +284,19 @@ def get_coldkey_password_from_environment(coldkey_name: str) -> Optional[str]:
...
     }
-    return envs.get(f"BT_COLD_PW_{coldkey_name.replace('-', '_').upper()}")
+    pass_ = envs.get(f"BT_COLD_PW_{coldkey_name.upper()}")
+    if "-" in coldkey_name:
+        if pass_ is None:
+            return envs.get(f"BT_COLD_PW_{coldkey_name.replace('-', '_').upper()}")
+        else:
+            warnings.warn(
+                "Cold key name contains a hyphen. "
+                "Please use an underscore instead in enviornment variable name. "
```
garrett-opentensor:
```suggestion
                "Please use an underscore instead in environment variable name. "
```
fix typo

alex-drocks (approved):

mjurbanski-reef:
no longer relevant after refactor

PR #2311, opened 2024-09-17
remove unused code (tensor.py-> class tensor), remove old tests, add new tests

basfroman:


thewhaleking (commented):
Do we really want to remove all these tests? It seems like they can just be slightly modified to work with `Tensor`

basfroman:
> Do we really want to remove all these tests? It seems like they can just be slightly modified to work with `Tensor`

Fixed

thewhaleking (approved):

PR #2326, opened 2024-09-30
add ConnectionRefusedError raising

basfroman:


bittensor/core/subtensor.py
```diff
@@ -186,6 +186,7 @@ def __init__(
 
         self.log_verbose = log_verbose
         self._connection_timeout = connection_timeout
+        self.substrate: "SubstrateInterface" = None
```
thewhaleking, line 189:
```suggestion
        self.substrate: Optional["SubstrateInterface"] = None
```

bittensor/core/subtensor.py
```diff
@@ -223,14 +225,15 @@ def _get_substrate(self):
...
+        except ConnectionRefusedError as error:
             logging.error(
                 f"Could not connect to {self.network} network with {self.chain_endpoint} chain endpoint.",
             )
             logging.info(
                 "You can check if you have connectivity by running this command: nc -vz localhost "
-                f"{self.chain_endpoint.split(':')[2]}"
+                f"{self.chain_endpoint}"
             )
+            raise ConnectionRefusedError(error.args)
```
thewhaleking, line 236:
```suggestion
            raise
```

Then you don't have to worry about the `error` var.

thewhaleking (approved):
Two suggestions. If you use the second one, also remove the `as error` from line 228

bittensor/core/subtensor.py
basfroman: There are reasons why I didn't do it

bittensor/core/subtensor.py
```diff
@@ -223,14 +225,15 @@ def _get_substrate(self):
             except (AttributeError, TypeError, socket.error, OSError) as e:
                 logging.warning(f"Error setting timeout: {e}")
```
thewhaleking, line 226: This needs to be updated.

thewhaleking (changes requested):

bittensor/core/subtensor.py
basfroman: I want to show the user not only the stack, but also the code with the message

bittensor/core/subtensor.py
basfroman: don't think so

thewhaleking (approved):

PR #2351, opened 2024-10-14
Add reconnection logic to 6.12.3

KMFODA:
### Bug

https://discord.com/channels/1120750674595024897/1283483486182969465

Currently 6.12.3 does not have the fixed reconnection logic introduced in 7.4.0. This means that when running on testnet miners and validators get the attached error:

Screenshot_2024-09-11_at_09 55 12

### Alternate Designs

As per the linked Discord conversation the following change adds this wrapper https://github.com/opentensor/bittensor/blob/staging/bittensor/utils/networking.py#L178-L199 as a decorator to all the methods where chain call goes with self.substrate.%SOME_SUBSTRATE_METHOD% in an attempt to resolve this bug.

### DIFF
--- a/bittensor/subtensor.py
+++ b/bittensor/subtensor.py
@@ -82,7 +82,7 @@ from .extrinsics.senate import (
 )
 from .extrinsics.root import root_register_extrinsic, set_root_weights_extrinsic
 from .types import AxonServeCallParams, PrometheusServeCallParams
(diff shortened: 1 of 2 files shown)


thewhaleking:
This can be a patch applied locally, but we don't backport new features into old releases unless there are security vulnerabilities.

PR #2355, opened 2024-10-16
Add commit-reveal subprocess and related utilities

opendansor:
## Summary

This PR introduces a new subprocess dedicated to automating the commit-reveal mechanism. Key utility functions are also included, along with end-to-end tests to verify the subprocess functionality.

With this update, the commit-reveal process can be fully automated. Subnet owners need only enable the `commit-reveal` flag within the hyper-parameter settings, and the subprocess will handle the commit-reveal process autonomously.

## Details

- **Long-Lived Subprocess**:  
   This subprocess is designed to operate independently of the parent process. If the validator is stopped and restarted, the subprocess will continue managing existing commit-reveal cycles on its own. It doesn’t terminate when the parent thread is closed.

- **Local WebSocket Communication**:  
   A WebSocket connection on `localhost` allows the parent and subprocess to communicate. This connection offers the flexibility to start, terminate, commit, and reveal actions independently of the parent process, even if the parent has stopped. For example, you can terminate the subprocess from the terminal:

   ```bash
   echo "terminate" | nc localhost 9949

## Benefits
This enhancement improves:

- **Reliability**:
 The commit-reveal process becomes more robust, maintaining continuity across validator restarts.
- **Control**:
 Greater flexibility in managing the subprocess independently enhances control over commit-reveal operations.


scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+    Returns:
+        None
+    """
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server.bind(('127.0.0.1', 9949))
+    server.listen(5)
+    print('Listening on port 9949...')
+    while running:
+        client_sock, addr = server.accept()
+        client_handler = threading.Thread(
```
basfroman: I'd recommended to use `ThreadPoolExecutor(max_workers=amoun_of_workers)` is `amoun_of_workers` which limits the max workers for that pool. In case of some error each iteration of the loop will create a new trend and this will quickly lead to memory overflow.

example:
```python
def start_socket_server() -> None:
    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    server.bind(('127.0.0.1', 9949))
    server.listen(5)
    print('Listening on port 9949...')
    
    with ThreadPoolExecutor(max_workers=10) as executor:  # limit of workers amount
        while running:
            try:
                client_sock, addr = server.accept()
                print(f"Accepted connection from {addr[0]}.")
                executor.submit(handle_client_connection, client_sock)
            except socket.timeout:
                print("Socket timeout, continuing to listen...")
            except Exception as e:
                print(f"Error accepting connection: {e}.")
                break
```

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+                    uids=json.loads(args[9]),
+                    weights=json.loads(args[10]),
+                    salt=json.loads(args[11]),
+                    version_key=int(args[12])
+                )
+                committed(commit)
+            elif command == 'terminate':
+                terminate_process(None, None)
+            else:
+                print("Command not recognized")
```
basfroman: Pls consider to use the next approach.

```python
command = args[0]
commands = {
    'revealed': lambda: revealed(
        args[1],
        args[2],
        args[3],
        args[4],
        int(args[5]),
        json.loads(args[6]),
        json.loads(args[7]),
        json.loads(args[8]),
        int(args[9])
    ),
    'revealed_hash': lambda: revealed_hash(args[1]),
    'committed': lambda: committed(Commit(
        wallet_hotkey_name=args[3],
        wallet_hotkey_ss58=args[4],
        wallet_name=args[1],
        wallet_path=args[2],
        commit_hash=args[7],
        netuid=int(args[8]),
        commit_block=int(args[5]),
        reveal_block=int(args[6]),
        uids=json.loads(args[9]),
        weights=json.loads(args[10]),
        salt=json.loads(args[11]),
        version_key=int(args[12])
    )),
    'terminate': lambda: terminate_process(None, None)
}

if command in commands:
    try:
        commands[command]()
    except (IndexError, ValueError, json.JSONDecodeError) as e:
        logging.error(f"Error in processing command {command}: {e}")
else:
    logging.warning(f"Command not recognized: {command}")
```
It seems to me this makes the code more readable, and if an error comes, we can more clearly and easy trace where it comes from and for which command

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+    except Exception as e:
```
basfroman: I'd add one more exception checker like 
```python
except socket.error as e:
    logging.error(f"Socket error: {e}")
except Exception as e:
    ...
```
this will clearly show an error if the error is in the socket

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+        print(f"Error: {e}")
+    finally:
+        client_socket.close()
+
+
+def start_socket_server() -> None:
+    """
+    Starts the socket server to listen for incoming connections.
+
+    Returns:
```
basfroman: you don't really need this

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+        None
+    """
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server.bind(('127.0.0.1', 9949))
+    server.listen(5)
```
basfroman: In my experience, without `server.settimeout(X)`, the server can hang and lock up at the accept() stage if no commands are received for a long time. But I'm not sure about this here.

scripts/subprocess_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    for proc in psutil.process_iter(['pid', 'name', 'cmdline']):
+        cmdline = proc.info['cmdline']
+        if cmdline and (process_name in proc.info['name'] or any(process_name in cmd for cmd in cmdline)):
+            return proc.info['pid']
+    return None
+
+
+def start_commit_reveal_subprocess(network: Optional[str] = None, sleep_interval: Optional[float] = None):
+    """Start the commit reveal subprocess if not already running."""
+    log_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "subprocess", "logs"))
```
basfroman: Pls consider to use `~/.bittensor/` directory as a base for this purposes. Users are already accustomed to the fact that we store everything there.

tests/e2e_tests/conftest.py
```diff
@@ -1,13 +1,15 @@
+import logging
```
basfroman: why didn't you choose `bittensor.logging` for this purpose?

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -144,6 +146,33 @@ def commit_weights_extrinsic(
...
+def commit_weights_process(
+    subtensor: "Subtensor",
+    wallet: "Wallet",
+    netuid: int,
+    commit_hash: str,
+    uids: list[int],
+    weights: list[int],
+    salt: list[int],
+    version_key: int = settings.version_as_int,
+):
```
basfroman, line 157: pls add detailed docstring

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +301,211 @@ def reveal_weights_extrinsic(
...
+
+
+def reveal_weights_process(
+        wallet: "Wallet",
+        netuid: int,
+        uids: list[int],
+        weights: list[int],
+        salt: list[int],
+        version_key: int = settings.version_as_int,
+):
```
basfroman, line 330: pls add detailed docstring

bittensor/core/subtensor.py
```diff
@@ -130,12 +134,14 @@ class Subtensor:
...
-        _mock: bool = False,
-        log_verbose: bool = False,
-        connection_timeout: int = 600,
+            self,
+            network: Optional[str] = None,
+            config: Optional["Config"] = None,
+            _mock: bool = False,
+            log_verbose: bool = False,
+            connection_timeout: int = 600,
+            subprocess_initialization: bool = True,
```
basfroman: Consider setting `subprocess_initialization: bool = True` argument to `False` by default. But when initializing Subtensors from a validator script, they should pass this argument to `True`. 
Documenting this would be sufficient. I don't think we want to bother the majority of users who don't need it.

bittensor/core/subtensor.py
basfroman, line 1: pls check this module with `python3 -m ruff format .` command. It's very strange to have so many indents in all methods. ruff will fix it back.

scripts/__init__.py
basfroman, line 1: Am I right in understanding that the `scripts/` package is not included in the bittensor package data and to use it, validators will need to clone the repository, not just install bittensor from pypi?

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+import os
+import shlex
+import sys
+import time
+import socket
+import threading
+from bittensor.core.subtensor import Subtensor
+from bittensor_wallet import Wallet
+from scripts import subprocess_utils as utils
+from typing import List, Any, Dict, Tuple, Optional
```
basfroman: Pls use python native `list`, `tuple`, `dict` instead of `List`, `Dict`, `Tuple` for annotations.

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+    except Exception as e:
+        print(f"Error checking if table '{table_name}' is empty: {e}")
+        return False
+
+
+def initialize_db() -> None:
+    """
+    Initializes the database by creating the 'commits' table if it does not exist.
+
+    Returns:
```
basfroman: you don't need it. the same above when functions/methods returns None

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+    Reveals the weights for a batch of commits to the subtensor network.
+
+    Args:
+        subtensor (Subtensor): The subtensor network object.
+        commits (List[Commit]): A list of commit objects to be revealed.
+
+    Returns:
+        None
+    """
+    wallet = Wallet(name=commits[0].wallet_name, path=commits[0].wallet_path, hotkey=commits[0].wallet_hotkey_name)
```
basfroman: pls check if `commits[0]` is exist before use it

scripts/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,532 @@
...
+    while running:
+        client_sock, addr = server.accept()
+        client_handler = threading.Thread(
+            target=handle_client_connection,
+            args=(client_sock,)
+        )
+        client_handler.start()
+
+
+def terminate_process(signal_number: Optional[int], frame: Optional[Any]) -> None:
```
basfroman: do you plan to use `frame` argument within this function?

scripts/subprocess_utils.py
```diff
@@ -0,0 +1,157 @@
...
+        conn.commit()
+    return
+
+
+def read_table(table_name: str, order_by: str = "") -> tuple[list, list]:
+    """
+    Reads a table from a SQLite database, returning back a column names and rows as a tuple
+    :param table_name: the table name in the database
+    :param order_by: the order of the columns in the table, optional
+    :return: ([column names], [rows])
```
basfroman: the same as above

scripts/subprocess_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def create_table(title: str, columns: list[tuple[str, str]], rows: list[list]) -> None:
+    """
+    Creates and populates the rows of a table in the SQLite database.
+
+    :param title: title of the table
+    :param columns: [(column name, column type), ...]
+    :param rows: [[element, element, ...], ...]
+    :return: None
```
basfroman: Let's be consistent in the repo. Pls use Google Documentation Style for this and other docstrings.

scripts/subprocess_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    Reads a table from a SQLite database, returning back a column names and rows as a tuple
+    :param table_name: the table name in the database
+    :param order_by: the order of the columns in the table, optional
+    :return: ([column names], [rows])
+    """
+    with DB() as (conn, cursor):
+        cursor.execute(f"PRAGMA table_info({table_name})")
+        columns_info = cursor.fetchall()
+        column_names = [info[1] for info in columns_info]
+        column_types = [info[2] for info in columns_info]
```
basfroman: Is there a chance (even 1%) that `len(info)` will be less than 3? Would be worth wrapping this logic in a try - except?

tests/e2e_tests/test_reveal_weights.py
basfroman, line 1: I didn't check the tests until `bt_decode` was updated

basfroman (changes requested):

scripts/subprocess/commit_reveal.py
opendansor: Yeah, makes sense. I will add that in.

scripts/subprocess/commit_reveal.py
opendansor: Absolutely!

scripts/subprocess_utils.py
opendansor: I understand, but I think it might make more sense to store the logs where the process lives. I am kind of split on this one.

basfroman: Honestly, I don't really understand how this can be useful, since most users won't even know where the process is located. But the entire community knows where we store all the bittensor, wallet, and cli related files. This is a historically established location. I suggest being consistent (as you suggest me many times ;) ).

scripts/subprocess_utils.py
opendansor: No harm in doing a try-except

scripts/subprocess_utils.py
opendansor: If you are the type to be concerned about a process then you know how to look them up and see what they do. In our case its extremely clear where the process originated and even what parameters where passed to it

image

scripts/__init__.py
opendansor: I am not sure about this. I will double check.

scripts/subprocess_utils.py
basfroman: I see. But let's use `~/.bittensor/` directory for storing all data `bittensor` SDK related.

opendansor: Done. Storing in ~/.bittensor/logs directory

scripts/__init__.py
opendansor: Moved it.

basfroman (changes requested):
pls fix ruff and filed tests

bittensor/core/extrinsics/set_weights.py
```diff
@@ -68,23 +70,7 @@ def do_set_weights(
...
-                "version_key": version_key,
-            },
-        )
-        # Period dictates how long the extrinsic will stay as part of waiting pool
-        extrinsic = self.substrate.create_signed_extrinsic(
-            call=call,
-            keypair=wallet.hotkey,
-            era={"period": 5},
-        )
+    def make_substrate_call_with_retry(extrinsic):
```
basfroman: pls don't use the name defined already outer scop. usually `extrinsic_` recommended in this case.

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
         tuple[bool, str]: A tuple containing a success flag and an optional response message.
     """
-    # First convert types.
-    if use_torch():
-        if isinstance(uids, list):
-            uids = torch.tensor(uids, dtype=torch.int64)
-        if isinstance(weights, list):
-            weights = torch.tensor(weights, dtype=torch.float32)
+
+    if subtensor.get_subnet_hyperparameters(
```
basfroman: since `get_subnet_hyperparameters` has Optional return, it's recommendet to use double if statement or use `walrus operator` (:=)

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
+        netuid=netuid
+    ).commit_reveal_weights_enabled:
+        # if cr is enabled, commit instead of setting the weights.
+        salt = [random.randint(0, 350) for _ in range(8)]
+
+        # Ask before moving on.
```
basfroman: let's avoid use prompt in sdk. I have already started this refactoring as I work with the current PRs. The SDK does not provide for interactive interaction.

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
+        if prompt:
+            if not Confirm.ask(
+                f"Do you want to commit weights:\n[bold white]  weights: {weights}\n"
+                f"uids: {uids}[/bold white ]?"
+            ):
+                return False, "Prompt refused."
+
+        with bt_console.status(
```
basfroman: The PR for replacing `console` with `btlogging` has already been released (https://github.com/opentensor/bittensor/pull/2377). Please apply this approach wherever new code is created across whole PR. In changed parts of the code this will be replaced after the merge.

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
-    if prompt:
-        if not Confirm.ask(
-            f"Do you want to set weights:\n[bold white]  weights: {[float(v / 65535) for v in weight_vals]}\n"
-            f"uids: {weight_uids}[/bold white ]?"
+        # Reformat and normalize.
+        weight_uids, weight_vals = weight_utils.convert_weights_and_uids_for_emit(
+            uids, weights
+        )
+
+        # Ask before moving on.
```
basfroman: the same as above

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
+        if prompt:
+            if not Confirm.ask(
+                f"Do you want to set weights:\n[bold white]  weights: {[float(v / 65535) for v in weight_vals]}\n"
+                f"uids: {weight_uids}[/bold white ]?"
+            ):
+                return False, "Prompt refused."
+
+        with bt_console.status(
```
basfroman: the same as above

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
-            return False, str(e)
+
+                if not wait_for_finalization and not wait_for_inclusion:
+                    return True, "Not waiting for finalization or inclusion."
+
+                if success is True:
+                    bt_console.print(
+                        ":white_heavy_check_mark: [green]Finalized[/green]"
+                    )
+                    logging.success(
```
basfroman: pls replace `bt_console.print` with `logging.success` in this case

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
+                        suffix="<green>Finalized: </green>",
+                    )
+                    return True, "Successfully set weights and Finalized."
+                else:
+                    error_message = format_error_message(error_message)
+                    logging.error(error_message)
+                    return False, error_message
+
+            except Exception as e:
+                bt_console.print(f":cross_mark: [red]Failed[/red]: error:{e}")
```
basfroman: pls replace `bt_console.print` -> `logging.error`

bittensor/core/subtensor.py
```diff
@@ -1818,10 +1844,10 @@ def commit_weights(
         uids: Union[NDArray[np.int64], list],
         weights: Union[NDArray[np.int64], list],
         version_key: int = settings.version_as_int,
-        wait_for_inclusion: bool = False,
+        wait_for_inclusion: bool = True,
         wait_for_finalization: bool = False,
         prompt: bool = False,
-        max_retries: int = 5,
+        max_retries: int = 1,
```
basfroman: if we have `max_retries=1` might be it good to remove thia argument and reduce the logic in line 1907?

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+        wait_for_inclusion=wait_for_inclusion,
+        wait_for_finalization=wait_for_finalization,
+    )
+
+    if success:
+        success_message = "Successfully batch revealed weights."
+        logging.info(success_message)
+        return True, success_message
+    else:
+        error_message = format_error_message(error_message)
```
basfroman: pls add `substrate` argument to the call to avoid bug after reviewed PR merged.

bittensor/core/extrinsics/transfer.py
```diff
@@ -63,15 +63,7 @@ def do_transfer(
...
-    def make_substrate_call_with_retry():
-        call = self.substrate.compose_call(
-            call_module="Balances",
-            call_function="transfer_allow_death",
-            call_params={"dest": dest, "value": transfer_balance.rao},
-        )
-        extrinsic = self.substrate.create_signed_extrinsic(
-            call=call, keypair=wallet.coldkey
-        )
+    def make_substrate_call_with_retry(extrinsic):
```
basfroman: Shadows name 'extrinsic' from outer scope. pls use `extrinsic_`

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -62,19 +65,7 @@ def do_commit_weights(
...
-            call_params={
-                "netuid": netuid,
-                "commit_hash": commit_hash,
-            },
-        )
-        extrinsic = self.substrate.create_signed_extrinsic(
-            call=call,
-            keypair=wallet.hotkey,
-        )
+    def make_substrate_call_with_retry(extrinsic):
```
basfroman: Shadows name 'extrinsic' from outer scope. pls use extrinsic_

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -144,6 +147,55 @@ def commit_weights_extrinsic(
...
+        weights (list[int]): List of weight values corresponding to each UID.
+        salt (list[int]): List of salt values for the hash function.
+        version_key (int): Version key for network compatibility (default is settings.version_as_int).
+        block (Optional[int]): Specific block number to use (default is None).
+
+    The function calculates the necessary blocks until the next epoch and the reveal block, then the subprocess will
+    wait until the appropriate time to reveal the weights.
+    """
+
+    def send_command(command):
```
basfroman: Shadows name 'command' from outer scope. pls use `command_`

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -144,6 +147,55 @@ def commit_weights_extrinsic(
...
+        client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        client.connect(("127.0.0.1", 9949))
+        client.send(command.encode())
+        client.close()
+
+    curr_block = block if block is not None else subtensor.get_current_block()
+    blocks_until_next_epoch = subtensor.blocks_until_next_epoch(netuid=netuid)
+    subnet_tempo_blocks = subtensor.get_subnet_hyperparameters(netuid=netuid).tempo
```
basfroman: potential place for bug in case if `subtensor.get_subnet_hyperparameters(netuid=netuid)` returned `None`

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -144,6 +147,55 @@ def commit_weights_extrinsic(
...
+    epoch_start_block = curr_block + blocks_until_next_epoch
+    cr_periods = subtensor.get_subnet_hyperparameters(
+        netuid=netuid
+    ).commit_reveal_weights_interval
```
basfroman: the same as above. I'd recommend to use `assert` of `if` statement

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -179,22 +231,7 @@ def do_reveal_weights(
...
-                "values": values,
-                "salt": salt,
-                "version_key": version_key,
-            },
-        )
-        extrinsic = self.substrate.create_signed_extrinsic(
-            call=call,
-            keypair=wallet.hotkey,
-        )
+    def make_substrate_call_with_retry(extrinsic):
```
basfroman: Shadows name `extrinsic` from outer scope. pls use `extrinsic_`

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+    Args:
+        wallet (bittensor_wallet.Wallet): The wallet associated with the neuron revealing the weights.
+        netuid (int): The unique identifier of the subnet.
+        uids (list[int]): List of neuron UIDs for which weights are being revealed.
+        weights (list[int]): List of weight values corresponding to each UID.
+        salt (list[int]): List of salt values corresponding to the hash function.
+        version_key (int): Version key for compatibility with the network. Defaults to `settings.version_as_int`.
+    """
+
+    def send_command(command):
```
basfroman: Shadows name 'command' from outer scope

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+        wait_for_finalization (bool, optional): Waits for the transaction to be finalized on the blockchain. Defaults to False.
+
+    Returns:
+        tuple[bool, Optional[dict]]: A tuple containing a success flag and an optional error message.
+
+    This method ensures that the weight revelation is securely recorded on the Bittensor blockchain, providing transparency and accountability for the neuron's weight distribution.
+    """
+
+    @retry(delay=1, tries=3, backoff=2, max_delay=4)
+    def make_substrate_call_with_retry(extrinsic):
```
basfroman: Shadows name 'extrinsic' from outer scope

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+        netuid (int): The unique identifier of the subnet.
+        uids (list[list[int]]): List of neuron UID lists for which weights are being revealed in batch.
+        weights (list[list[int]]): List of weight value lists corresponding to each UID list.
+        salt (list[list[int]]): List of salt value lists corresponding to the hash function for each batch.
+        version_keys (list[int]): List of version keys for compatibility with the network for each batch.
+
+    This function facilitates the batch reveal process, ensuring that the hashed weights are properly recorded and sent.
+    """
+
+    def send_command(command):
```
basfroman: Shadows name 'command' from outer scope

bittensor/utils/subprocess_utils.py
```diff
@@ -0,0 +1,364 @@
+import os
```
basfroman: Pls move this module to `bittensor/utils/subprocess` sub-package and also it would be logical to rename it as `bittensor/utils/subprocess/utils.py`
In this case it obvious since all `subprocess`-related modules are located in the same place.

bittensor/utils/subprocess_utils.py
```diff
@@ -0,0 +1,364 @@
...
+import re
+import sqlite3
+import subprocess
+import time
+from datetime import datetime
+from typing import Optional
+
+import psutil
+
+LOG_DIR = os.path.expanduser("~/.bittensor/logs")
```
basfroman: nitpick: I'd recommend to use `os.path.join(...)` instead of `"~/.bittensor/logs"`. But until we officially support Linux only, it's ok

bittensor/core/subtensor.py
```diff
@@ -75,14 +79,26 @@
...
 from bittensor.utils.registration import legacy_torch_api_compat
-from bittensor.utils.weight_utils import generate_weight_hash
+from bittensor.utils.weight_utils import (
+    generate_weight_hash,
+    convert_weights_and_uids_for_emit,
+)
 
 KEY_NONCE: dict[str, int] = {}
 
+COMMIT_REVEAL_PROCESS = "commit_reveal.py"
```
basfroman: since the const name is the same as `bittensor/utils/subprocess_utils.py:PROCESS_NAME` pls import it from there or vice versa if this is not make circular import issue.
Having 2 identical constants in 2 different places can very easily lead to confusion or bugs if changed.

bittensor/utils/subprocess_utils.py
```diff
@@ -0,0 +1,364 @@
...
+    Returns:
+        bool: True if the message is found in the logs, False otherwise.
+    """
+    try:
+        stdout_log, stderr_log = get_cr_log_files()
+    except RuntimeError as e:
+        print(str(e))
+        return False
+
+    def check_message_in_log(file_path: str, message: str) -> bool:
```
basfroman: Shadows name 'message' from outer scope. pls use `message_`

bittensor/utils/subprocess_utils.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class DB:
+    """
+    For ease of interaction with the SQLite database used for --reuse-last and --html outputs of tables
+    """
+
+    def __init__(
+        self,
+        db_path: str = os.path.expanduser("~/.bittensor/bittensor.db"),
```
basfroman: the same nitpick about os.path.join

bittensor/utils/subprocess_utils.py
```diff
@@ -0,0 +1,364 @@
...
+        conn.commit()
+        columns_ = ", ".join([" ".join(x) for x in columns])
+        creation_query = f"CREATE TABLE IF NOT EXISTS {title} ({columns_})"
+        conn.commit()
+        cursor.execute(creation_query)
+        conn.commit()
+        query = f"INSERT INTO {title} ({', '.join([x[0] for x in columns])}) VALUES ({', '.join(['?'] * len(columns))})"
+        cursor.executemany(query, rows)
+        conn.commit()
+    return
```
basfroman: nitpick: I'd return `True` in successful case of table creation. Also, you can use `assert result_of_creation is True` in your code to be make sure operation is successful.

tests/e2e_tests/conftest.py
```diff
@@ -2,84 +2,99 @@
...
+        # Start new node process
+        process = subprocess.Popen(
+            cmds, stdout=subprocess.PIPE, text=True, preexec_fn=os.setsid
+        )
+
+        # Wait for the node to start using the existing pattern match
+        pattern = re.compile(r"Imported #1")
+        timestamp = int(time.time())
+
+        def wait_for_node_start(process, pattern):
```
basfroman: Shadows names `process` and `pattern` from outer scope. pls use names with `_` after var name.

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+
+    Returns:
+        tuple[bool, str]: ``True`` if the weight revelation is successful, ``False`` otherwise. And `msg`, a string
+        describing the success or potential error.
+
+    This function provides a user-friendly interface for revealing weights in batch on the Bittensor blockchain,
+    ensuring proper error handling and user interaction when required.
+    """
+
+    if prompt and not Confirm.ask(f"Would you like to batch reveal weights?"):
```
basfroman: let's avoid prompt for sdk

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -272,3 +324,233 @@ def reveal_weights_extrinsic(
...
+        values=weights,
+        salt=salt,
+        version_keys=version_keys,
+        wait_for_inclusion=wait_for_inclusion,
+        wait_for_finalization=wait_for_finalization,
+    )
+
+    if success:
+        success_message = "Successfully batch revealed weights."
+        logging.info(success_message)
```
basfroman: logging.success ?

bittensor/core/extrinsics/set_weights.py
```diff
@@ -132,63 +134,134 @@ def set_weights_extrinsic(
...
-        if not Confirm.ask(
-            f"Do you want to set weights:\n[bold white]  weights: {[float(v / 65535) for v in weight_vals]}\n"
-            f"uids: {weight_uids}[/bold white ]?"
+        # Reformat and normalize.
+        weight_uids, weight_vals = weight_utils.convert_weights_and_uids_for_emit(
+            uids, weights
+        )
+
+        # Ask before moving on.
+        if prompt:
```
basfroman: let's avoid prompt for SDK

bittensor/utils/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,784 @@
...
+        commit_data = commit.to_dict()
+        column_names = ", ".join(commit_data.keys())
+        data = ", ".join(["?"] * len(commit_data))
+        sql = f"INSERT INTO commits ({column_names}) VALUES ({data})"
+        cursor.execute(sql, tuple(commit_data.values()))
+        conn.commit()
+    print(f"Committed data: {commit_data}")
+
+
+def get_all_commits() -> List[Commit]:
```
basfroman: if this function is used within module only, pls use `_get_all_commits` for name. The same for another functions.

bittensor/utils/subprocess/commit_reveal.py
```diff
@@ -0,0 +1,784 @@
...
+    finally:
+        client_socket.close()
+
+
+def start_socket_server():
+    """
+    Starts the socket server to listen for incoming connections.
+    """
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server.bind(("127.0.0.1", 9949))
```
basfroman: Have you considered the option when the port is busy for some reason? The user may not know about it.
As an idea: Perhaps it is worth implementing a checking mechanism before starting the process, otherwise inform the user about it.

basfroman (changes requested):
Looks pretty well. Just lest a few comments.

bittensor/core/extrinsics/set_weights.py
opendansor: Ill add a check for the optional, but this should not be a valid path. Hyperparameters have default values and if we get a None as a return then something else has already gone wrong upstream.

bittensor/core/extrinsics/set_weights.py
opendansor: I think this prompt change should be in its own pr "remove prompts from SDK", changing the pattern and mixing this PR with other objectives is not the right way to go.

bittensor/core/extrinsics/set_weights.py
opendansor: Yeah, I see this was merged a few hrs ago. I will need to rebase on your PR, and I will apply to changes I have written.

bittensor/core/extrinsics/set_weights.py
opendansor: same as above. I think there should be a separate PR to remove prompts. If you like I can write that PR after this one.

basfroman (approved):

bittensor/core/extrinsics/commit_weights.py
```diff
@@ -268,11 +310,240 @@ def reveal_weights_extrinsic(
...
+    Args:
+        wallet (bittensor_wallet.Wallet): The wallet associated with the neuron revealing the weights.
+        netuid (int): The unique identifier of the subnet.
+        uids (list[int]): List of neuron UIDs for which weights are being revealed.
+        weights (list[int]): List of weight values corresponding to each UID.
+        salt (list[int]): List of salt values corresponding to the hash function.
+        version_key (int): Version key for compatibility with the network. Defaults to `settings.version_as_int`.
+    """
+
+    def send_command(command_):
```
thewhaleking, line 348: Is there a reason this function isn't reused?

bittensor/utils/background_subprocess/commit_reveal.py
```diff
@@ -0,0 +1,790 @@
...
+    """
+    _initialize_db()
+    print(
+        f"initializing subtensor with network: {parsed_args.network} and sleep time: {parsed_args.sleep_interval} seconds"
+    )
+    subtensor = Subtensor(network=parsed_args.network, subprocess_initialization=False)
+    server_thread = threading.Thread(target=_start_socket_server)
+    server_thread.start()
+
+    counter = 0  # Initialize counter
```
thewhaleking, line 757: remove comment

bittensor/utils/background_subprocess/commit_reveal.py
```diff
@@ -0,0 +1,790 @@
...
+        help="Subtensor network address",
+    )
+    parser.add_argument(
+        "--sleep-interval",
+        type=float,
+        default=12.0,
+        help="Interval between block checks in seconds",
+    )
+    args = parser.parse_args()
+    main(args)
```
thewhaleking, line 790: Why is this using an arg parser?

bittensor/utils/background_subprocess/commit_reveal.py
```diff
@@ -0,0 +1,790 @@
...
+    _initialize_db()
+    print(
+        f"initializing subtensor with network: {parsed_args.network} and sleep time: {parsed_args.sleep_interval} seconds"
+    )
+    subtensor = Subtensor(network=parsed_args.network, subprocess_initialization=False)
+    server_thread = threading.Thread(target=_start_socket_server)
+    server_thread.start()
+
+    counter = 0  # Initialize counter
+    print("commit_reveal background_subprocess is ready")
```
thewhaleking, line 758: Do we want to use print here?

bittensor/utils/background_subprocess/commit_reveal.py
```diff
@@ -0,0 +1,790 @@
...
+
+def _start_socket_server():
+    """
+    Starts the socket server to listen for incoming connections.
+    """
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server.bind((HOST, PORT))
+    server.listen(5)
+    server.settimeout(2)  # Set timeout for any incoming requests to 2 seconds
+    print(f"Listening on port {PORT}...")
```
thewhaleking, line 713: Do we want to use print here?

bittensor/utils/background_subprocess/commit_reveal.py
```diff
@@ -0,0 +1,790 @@
...
+        # Group commits by wallet_hotkey_ss58
+        if local_commits:
+            unique_combinations = list(
+                {(commit.wallet_hotkey_ss58, commit.netuid) for commit in local_commits}
+            )
+
+            for combination in unique_combinations:
+                ss58, netuid = combination
+                try:
+                    response = subtensor.query_module(
```
thewhaleking, line 348: Is this keeping the Subtensor object (and underlying ws connection) alive for the duration of this background process?

thewhaleking (changes requested):
Most of the file changes are fine, but I think everything within `background_subprocess` needs a serious re-examination (except the SQL stuff).

thewhaleking:
By design of Constt, we're skipping from v1 to v3.

PR #2366, opened 2024-10-21
Don't strip ANSI from stdout (fixes #2365)

### LINKED ISSUE #2365: ANSI Escape Sequences do not work under PM2 in version of bittensor older than 6.9.4
### Describe the bug

When using ansi escape sequences with bittensor 6.9.4, the messages are correctly formatted, but not later (7.1.2 and 8.2.0 tested) versions. Even a simple `print()` before any bittensor functions are executed (other than the import itself) doesn't work with colorized messages in the later versions.

### To Reproduce

```
print("Test Colored: \033[0;33mTest")
import bittensor as bt
print("Test Colored: \033[0;33mTest")
```
while using `bittensor==6.9.4` yields correct colorization (while the python script is ran under PM2 process manager)
then change bittensor to be `8.2.0` or `7.1.2` (i havent tested with other versions), and see that after `import bittensor` the colorization stops working.
This behaviour will carry over to any bt.logging.info or ANY attempt to print to stdout.

### Expected behavior

All messages accept and use ANSI Escape sequences correctly

### Screenshots

image
image
image

### Environment

Python3.12, PM2 5.4.2, Bittensor 6.9.4/7.1.2/8.2.0, Ubuntu 24.04, kitty terminal over openssh

### Additional context

_No response_

vaqxai:
- Until now ANSI was stripped from stdout if bittensor was not running within a TTY (like in PM2) by colorama when it was intialized
- This caused colored output to be broken under PM2 (and other process managers which do not give a TTY do bittensor/python
- Bittensor posseses its own file logger, which makes this behavior unnecessary and unwanted.
- This patch disables that behavior.

Where ANSI Stripping takes place in colorama: https://github.com/tartley/colorama/blob/136808718af8b9583cb2eed1756ed6972eda4975/colorama/ansitowin32.py#L102

Bittensor call to colorama which indirectly breaks colors in terminal output in PM2: https://github.com/opentensor/bittensor/blob/5edcee44d8ba81898b432ea00d93480fa09d2d18/bittensor/utils/btlogging/format.py#L29

Fix for issue #2365


vaqxai:
I found out that using `strip=False` introduces arbitrary newlines, instead, `wrap=False` can be used, and everything works correctly.

Maybe this should be user-definable by some config parameter?

thewhaleking:
Sorry, what is PM2?

thewhaleking:
No response from OP after two weeks. Closing. Can be re-opened if still needed.

basfroman (approved):
I tested this. Works well for me.

PR #2382, opened 2024-11-05
Avoid using `prompt` in SDK

basfroman:
SDK is intended for use in Python scripts and does not provide for interactive things. Due to this it was decided to avoid using the prompt flag and the prompt argument.

Tests are updated.


bittensor/core/extrinsics/registration.py
```diff
@@ -399,11 +379,6 @@ def burned_register_extrinsic(
         logging.info(f"\t\tcoldkey: <blue>{neuron.coldkey}</blue>")
         return True
 
-    if prompt:
-        # Prompt user for confirmation.
-        if not Confirm.ask(f"Recycle {recycle_amount} to register on subnet:{netuid}?"):
-            return False
-
```
ibraheem-abe, line 406: Do you think it will be valuable to just log this information for the user? 
I think recycle_amount would be great to show along with subnet

bittensor/core/extrinsics/set_weights.py
```diff
@@ -151,14 +148,6 @@ def set_weights_extrinsic(
...
     )
 
-    # Ask before moving on.
-    if prompt:
-        if not Confirm.ask(
-            f"Do you want to set weights:\n[bold white]  weights: {[float(v / 65535) for v in weight_vals]}\n"
-            f"uids: {weight_uids}[/bold white ]?"
-        ):
-            return False, "Prompt refused."
-
```
ibraheem-abe, line 161: Same point here

bittensor/core/extrinsics/transfer.py
```diff
@@ -161,17 +158,6 @@ def transfer_extrinsic(
...
-    if prompt:
-        if not Confirm.ask(
-            "Do you want to transfer:[bold white]\n"
-            f"  amount: {transfer_balance}\n"
-            f"  from: {wallet.name}:{wallet.coldkey.ss58_address}\n"
-            f"  to: {dest}\n"
-            f"  for fee: {fee}[/bold white]"
-        ):
-            return False
-
```
ibraheem-abe, line 174: Same point here, maybe we can just log all this info

bittensor/core/extrinsics/registration.py
```diff
@@ -160,16 +155,6 @@ def register_extrinsic(
...
-    if prompt:
-        if not Confirm.ask(
-            "Continue Registration?\n  hotkey:     [bold white]{}[/bold white]\n  coldkey:    [bold white]{}[/bold white]\n  network:    [bold white]{}[/bold white]".format(
-                wallet.hotkey.ss58_address,
-                wallet.coldkeypub.ss58_address,
-                subtensor.network,
-            )
-        ):
-            return False
-
```
ibraheem-abe: Maybe logging this will be useful as well

ibraheem-abe (approved):
LGTM. Just a few tiny nits

bittensor/core/extrinsics/transfer.py
basfroman: done

bittensor/core/extrinsics/set_weights.py
basfroman: done

bittensor/core/extrinsics/registration.py
basfroman: done

PR #2392, opened 2024-11-05
Remove `retry` and fix tests

basfroman:
The historical idea of ​​retry was that regardless of any Errors, it repeated its attempts 3 times.
Now the SDK code is much more functional and cleaner. I catch different errors at different places.
1. `ConnectionRefusedError` error is caught during the initialization of the Subtensor
2. The connection failure error is handled by decorator `@networking.ensure_connected`
3. Only the subtensor response error remains based on the data passed by the user or based on the current state of the subtensor. In this case, if the first response was an error, then all three attempts return exactly the same result. We are just unnecessarily loading the chain with additional calls. We tested this behavior many times. This is the reason why we abandoned retry in the async impl of the Subtensor and AsyncSubstrateInterface.

Also, an important point is that SDK is a Python package that is intended for use in scripts. It is not an interactive package. All responsibility for processing the results is in the hands of the user. This should be considered as a more flexible way then in btcli. The user will be able to regulate the retries themselves if they want. If the data was wrong during the request and the response is not successful, then it is easier to process and make a request based on other data than to make 3 calls and then change the data.

### DIFF
--- a/bittensor/core/extrinsics/commit_weights.py
+++ b/bittensor/core/extrinsics/commit_weights.py
@@ -19,8 +19,6 @@
 
 from typing import Optional, TYPE_CHECKING
 
-from retry import retry
-
 from bittensor.core.extrinsics.utils import submit_extrinsic
 from bittensor.utils import format_error_message
 from bittensor.utils.btlogging import logging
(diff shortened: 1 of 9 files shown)


thewhaleking (approved):
lgtm

PR #2398, opened 2024-11-08
[Tests] AsyncSubtensor (Part 1)

basfroman:


bittensor/core/async_subtensor.py
```diff
@@ -277,7 +277,7 @@ async def get_subnets(self, block_hash: Optional[str] = None) -> list[int]:
         return (
             []
             if result is None or not hasattr(result, "records")
-            else [netuid async for netuid, exists in result if exists]
+            else [netuid for netuid, exists in result.records if exists]
```
thewhaleking: This won't work.

tests/unit_tests/test_async_subtensor.py
```diff
@@ -0,0 +1,291 @@
...
+    assert mocked_decode_account_id.mock_calls == [
+        mocker.call("0"),
+        mocker.call("1"),
+        mocker.call("2"),
+        mocker.call("3"),
+    ]
+
+
+@pytest.mark.asyncio
+async def test_encode_params(subtensor, mocker):
```
thewhaleking, line 47: I'm not sure what this actually tests. I would think it should fail for invalid types, but it doesn't.

tests/unit_tests/test_async_subtensor.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+    subtensor.get_netuids_for_hotkey = mocked_get_netuids_for_hotkey
+
+    # Call
+    result = await subtensor.is_hotkey_registered_any(
+        hotkey_ss58="hotkey", block_hash="FAKE_HASH"
+    )
+
+    # Asserts
+    assert result == (len(mocked_get_netuids_for_hotkey.return_value) > 0)
```
thewhaleking:
```suggestion
    assert result is (len(mocked_get_netuids_for_hotkey.return_value) > 0)
```

Use `is` for bool comparison

tests/unit_tests/test_async_subtensor.py
```diff
@@ -0,0 +1,291 @@
...
+    result = await subtensor.get_subnets(block_hash=fake_block_hash)
+
+    # Asserts
+    mocked_substrate_query_map.assert_called_once_with(
+        module="SubtensorModule",
+        storage_function="NetworksAdded",
+        block_hash=fake_block_hash,
+        reuse_block_hash=True,
+    )
+    assert result == result
```
thewhaleking: I don't think this is what you intended

thewhaleking (changes requested):

bittensor/core/async_subtensor.py
basfroman: It works perfect. Let's try yourself. But updated.

tests/unit_tests/test_async_subtensor.py
basfroman: As we have already discussed, this tests the actual calls internally. The error case is tested in a subsequent test.

thewhaleking (approved):

PR #2433, opened 2024-11-18
Integration tests

### LINKED ISSUE #2435: integration tests for sync SDK
We need to increase integration test coverage for the SDK.  The integration tests should handle cases where data is passed between modules in the bittensor SDK, including other packages like `bt-decode`

Important: https://github.com/opentensor/bittensor/issues/2435#issuecomment-2825858004

thewhaleking:
Integration tests, initial batch


tests/integration_tests/test_subtensor_integration.py
```diff
@@ -1,407 +1,143 @@
...
-if __name__ == "__main__":
-    unittest.main()
+from bittensor.core.chain_data.axon_info import AxonInfo
+
+from bittensor import NeuronInfo
+from bittensor.core.subtensor import Subtensor
+from tests.helpers.integration_websocket_data import WEBSOCKET_RESPONSES
+
+
+class FakeWebsocket(ClientConnection):
```
basfroman: let's consider keep this class in some helper module if we gonna use in different test modules.

basfroman (dismissed):

ibraheem-abe (dismissed):
All tests passing 👍

tests/integration_tests/test_subtensor_integration.py
thewhaleking: Good idea

ibraheem-abe (approved):

PR #2442, opened 2024-11-19
Improve reconnection logic

basfroman:


ibraheem-abe (dismissed):

bittensor/core/subtensor.py
```diff
@@ -239,11 +222,7 @@ def _get_substrate(self):
 
         except (ConnectionRefusedError, ssl.SSLError) as error:
             logging.error(
-                f"Could not connect to {self.network} network with {self.chain_endpoint} chain endpoint.",
-            )
-            logging.info(
-                "You can check if you have connectivity by running this command: nc -vz localhost "
-                f"{self.chain_endpoint}"
+                f"<red>Could not connect to</red> <blue>{self.network}</blue> <red>network with</red> <blue>{self.chain_endpoint}</blue> <red>chain endpoint.</red>",
```
ibraheem-abe, line 225: I think this can also work and reduce the no. of times we use tags:
```
logging.error(
    f"<red>Could not connect to <blue>{self.network}</blue> network with <blue>{self.chain_endpoint}</blue> chain endpoint.</red>",
)
```

PS: Will need to test this, but I usually implemented like this in btcli

bittensor/utils/networking.py
```diff
@@ -178,22 +163,49 @@ def get_formatted_ws_endpoint_url(endpoint_url: Optional[str]) -> Optional[str]:
...
+        return (
+            sock is not None
+            and sock.getsockopt(socket.SOL_SOCKET, socket.SO_ERROR) == 0
+        )
+
+    @retry(
+        exceptions=ConnectionRefusedError,
+        tries=5,
+        delay=5,
+        backoff=1,
```
ibraheem-abe, line 178: What do you think of adding a backoff of 1.5 instead? 
So instead of retries every 5 seconds, the time it waits will increase: `5, 7.5, 11,25, 16...`

I was thinking it might be good if the subtensor connection takes a bit more time to reinitialize

ibraheem-abe (approved):
Few suggestions, overall lgtm

bittensor/core/subtensor.py
basfroman: No this way doesn't work with logging, unfortunately. I've tried.

ibraheem-abe: Must be a `rich` specific thing then - thank you for checking it out!

bittensor/utils/networking.py
basfroman: What is exact reason for this action? just increasing the overall delay time?
Currently it works like 5, 10, 15, 20, 25.
What is maximum acceptable time-frame to retrying in your opinion?

ibraheem-abe: My thinking revolved around the local nodes which validators would usually run and if they are down, then it might take a while to start. 

However, after our discussion today in the standup regarding websocket subscriptions etc, I think standard increments are okay. The issue is something else.

bittensor/utils/networking.py
```diff
@@ -178,22 +163,49 @@ def get_formatted_ws_endpoint_url(endpoint_url: Optional[str]) -> Optional[str]:
...
+    )
+    def reconnect_with_retries(self):
+        """Attempt to reconnect with retries using retry library."""
+        logging.info("Attempting to reconnect to substrate...")
+        self._get_substrate()
+
+        old_level = logging.get_level()
+        logging.set_info()
+        logging.success("Connection successfully restored!")
+        logging.setLevel(old_level)
```
thewhaleking, line 189: I don't understand why we need this.

basfroman: yeah, after https://github.com/opentensor/bittensor/pull/2446 we don't need this logic.
after we merge 2446 I gonna apply btlogging.console everywhere we need it (like in this case).

thewhaleking (approved):

PR #2445, opened 2024-11-20
Async unittests for `bittensor/core/extrinsics/async_registration.py`

basfroman:


bittensor/core/extrinsics/async_registration.py
```diff
@@ -212,154 +254,17 @@ async def register_extrinsic(
                 # Exited loop because pow is no longer valid.
                 logging.error("<red>POW is stale.</red>")
                 # Try again.
-                continue
+                # continue
 
+        # TODO(roman): discuss with team (looks like this section is not reachable)
```
thewhaleking: This can definitely be reached if the POW result is valid, but registration fails (and not because it's already registered) until the result goes stale.

bittensor/core/extrinsics/async_registration.py
```diff
@@ -212,154 +254,17 @@ async def register_extrinsic(
...
                 f":satellite: <magenta>Failed registration, retrying pow ...</magenta> <blue>({attempts}/{max_allowed_attempts})</blue>"
             )
+            continue
         else:
             # Failed to register after max attempts.
             logging.error("<red>No more attempts.</red>")
             return False
-
-
-async def run_faucet_extrinsic(
```
thewhaleking, line 229: why are we removing this?

basfroman: `faucet_extrinsic` was decided to be removed from SDK after splitting 7.4.0 into 3 packages

thewhaleking (approved):

PR #2446, opened 2024-11-20
BittensorConsole class

basfroman:
blogging gets a new `console` field. It appears immediately during `LoggingMachine` instance is initialization.
`BittensorConsole` methods allow you to output messages to the terminal/console without changing the logging level of the main logging (btlogging).

Usage example:
```python
from bittensor import logging

# will be logged
logging.console.info("info message")
logging.console.error("error message")
logging.console.success("success message")
logging.console.warning("warning message")
 logging.console.critical("critical message")

# will not be logged
logging.info("test info")
```

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -132,10 +132,10 @@ class AsyncSubtensor:
                     self.network = "custom"
             else:
                 logging.info(
-                    f"Network not specified or not valid. Using default chain endpoint: <blue>{NETWORK_MAP[DEFAULTS.subtensor.network]}</blue>."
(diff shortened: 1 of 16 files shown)


thewhaleking (commented):
What's the rationale for adding this?

basfroman:
> BittensorConsole

By default, SDK has a `warning` logging level. Any messages of a lower level will not be displayed. Sometimes we need to display messages (as we did it with `rich.console`), but the user will not see it. `BittensorConsole` allows you to display messages of any level without explicitly changing them. The difference from "print" is that we can use color and emoji support, since `BittensorConsole` is based on blogging.

thewhaleking (dismissed):

thewhaleking (approved):

PR #2455, opened 2024-11-21
Use websockets for Subtensor

thewhaleking:
This changes the way Subtensor connects to the websocket server. Previously, the websocket was initialized from within SubstrateInterface. However, this was using the stdlib `websocket` package. The PyPI `websockets` package has significantly better stability, and results in no broken pipes in our testing.

Furthermore, because we're bumping the websockets requirement version to 14.1+, we're transitioning the async websockets client in AsyncSubstrate to use this updated client.

### DIFF
--- a/bittensor/core/subtensor.py
+++ b/bittensor/core/subtensor.py
@@ -148,7 +148,7 @@ class Subtensor:
             _mock (bool): If set to ``True``, uses a mocked connection for testing purposes. Default is ``False``.
             log_verbose (bool): Whether to enable verbose logging. If set to ``True``, detailed log information about the connection and network operations will be provided. Default is ``True``.
             connection_timeout (int): The maximum time in seconds to keep the connection alive. Default is ``600``.
-            websocket: websockets sync (threading) client object connected to the network.
+            websocket (websockets.sync.client.ClientConnection): websockets sync (threading) client object connected to the network.
 
         This initialization sets up the connection to the specified Bittensor network, allowing for various blockchain operations such as neuron registration, stake management, and setting weights.
         """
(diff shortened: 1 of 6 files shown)


bittensor/core/subtensor.py
```diff
@@ -147,6 +148,7 @@ def __init__(
             _mock (bool): If set to ``True``, uses a mocked connection for testing purposes. Default is ``False``.
             log_verbose (bool): Whether to enable verbose logging. If set to ``True``, detailed log information about the connection and network operations will be provided. Default is ``True``.
             connection_timeout (int): The maximum time in seconds to keep the connection alive. Default is ``600``.
+            websocket: websockets sync (threading) client object connected to the network.
```
basfroman:
```suggestion
            websocket (websockets.sync.client.ClientConnection): websockets sync (threading) client object connected to the network.
```

bittensor/utils/async_substrate_interface.py
```diff
@@ -622,7 +622,7 @@ def __init__(
         # TODO allow setting max concurrent connections and rpc subscriptions per connection
         # TODO reconnection logic
         self.ws_url = ws_url
-        self.ws: Optional[websockets.WebSocketClientProtocol] = None
+        self.ws: Optional[ClientConnection] = None
```
basfroman:
```suggestion
        self.ws: Optional["ClientConnection"] = None
```

bittensor/utils/async_substrate_interface.py
```diff
@@ -21,6 +19,8 @@
     BlockNotFound,
 )
 from substrateinterface.storage import StorageKey
+from websockets.asyncio.client import ClientConnection, connect
```
basfroman: you can move this import `from websockets.asyncio.client import ClientConnection` to `if TYPE_CHECKING:` section to reduce import load. the same in `subtensor.py`

basfroman (changes requested):

basfroman (approved):

PR #2468, opened 2024-11-25
Update unit tests websocket

thewhaleking:
Updates unit tests to use the mocked websocket that is now used for subtensor instantiation.


basfroman:
I've tested this changes and them are solving the real-chain-call problem partially. We still have 39 unit-tests broken without the connection.
image
But I believe those tests are old one and doesn't related with websockt update. 
Anyway we need to create the issue for to fix this behavior.

basfroman (approved):

PR #2476, opened 2024-11-25
fix: increase stacklevel in LoggingMachine log calls

### LINKED ISSUE #2449: [SDK] Replace `loggingmachine.py` to real module name during logging
blogging always uses the name `loggingmachine.py` as "filename" in the formatter.

LoggingMachine logic needs to be improved to be able to pass a logger created in a separate module to use the module name in logging.

example:
```python
bt.logging.on()
2024-11-20 12:06:25.015 |       INFO       | bittensor:loggingmachine.py:472 | Logging enabled.
2024-11-20 12:06:25.015 |       INFO       | bittensor:loggingmachine.py:327 | Enabling default logging.
```

zyzniewski-reef:
### Bug

Fixes https://github.com/opentensor/bittensor/issues/2449

### Description of the Change

"loggingmachine.py" is printed in logs because LoggingMachine does not unstack it's internal calls to wrapped logger

### Alternate Designs

Adding LoggerAdapter

### Possible Drawbacks

None

### Verification Process

Ran unittests.
Checked logs are formatted correctly.

### Release Notes

- Fixed caller's module filename in logs

### DIFF
--- a/bittensor/utils/btlogging/loggingmachine.py
+++ b/bittensor/utils/btlogging/loggingmachine.py
@@ -443,10 +443,18 @@ class LoggingMachine(StateMachine, Logger):
         """
         return self.current_state_value == "Trace"
 
-    def trace(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
+    def trace(
+        self,
+        msg="",
+        prefix="",
+        suffix="",
+        *args,
+        stacklevel=2 if sys.version_info >= (3, 11) else 1, # logger.trace is not an internal method
+        **kwargs,
+    ):
         """Wraps trace message with prefix and suffix."""
         msg = _concat_message(msg, prefix, suffix)
-        self._logger.trace(msg, *args, **kwargs, stacklevel=stacklevel + 2)
+        self._logger.trace(msg, *args, **kwargs, stacklevel=stacklevel + 1)
 
     def debug(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
         """Wraps debug message with prefix and suffix."""
@@ -458,10 +466,18 @@ class LoggingMachine(StateMachine, Logger):
         msg = _concat_message(msg, prefix, suffix)
         self._logger.info(msg, *args, **kwargs, stacklevel=stacklevel + 1)
 
-    def success(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
+    def success(
+        self,
+        msg="",
+        prefix="",
+        suffix="",
+        *args,
+        stacklevel=2 if sys.version_info >= (3, 11) else 1, # logger.success is not an internal method
+        **kwargs,
+    ):
         """Wraps success message with prefix and suffix."""
         msg = _concat_message(msg, prefix, suffix)
-        self._logger.success(msg, *args, **kwargs, stacklevel=stacklevel + 2)
+        self._logger.success(msg, *args, **kwargs, stacklevel=stacklevel + 1)
 
     def warning(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
         """Wraps warning message with prefix and suffix."""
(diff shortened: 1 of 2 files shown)


basfroman:
Thank you for this contribution!
Could you please add the test which is testing exact PR's behavior? 
You can log all of types of logging (debug, info, etc) in the test and then capture and parse the log to be make sure that test module name is in the log line.

zyzniewski-reef:
ref: https://github.com/python/cpython/issues/97941

basfroman (changes requested):
plz fix ruff formatter issue

bittensor/utils/btlogging/loggingmachine.py
```diff
@@ -443,45 +443,65 @@ def __trace_on__(self) -> bool:
...
 
-    def trace(self, msg="", prefix="", suffix="", *args, **kwargs):
+    def trace(
+        self,
+        msg="",
+        prefix="",
+        suffix="",
+        *args,
+        stacklevel=2  # logger.trace is not an internal method
+        if sys.version_info >= (3, 11)
```
basfroman: There are several options to pass the value here, but not like this:
1. define a constant at the beginning of the module and pass it as a value;
2. pass a default value as None, then determine the value inside the function if it is None.

basfroman (changes requested):

bittensor/utils/btlogging/loggingmachine.py
zyzniewski-reef: thanks for pointing that out!

bittensor/utils/btlogging/loggingmachine.py
```diff
@@ -47,6 +47,12 @@
 from .helpers import all_loggers
 from bittensor.utils.btlogging.console import BittensorConsole
 
+# https://github.com/python/cpython/issues/97941
+if sys.version_info >= (3, 11):
+    CUSTOM_LOGGER_METHOD_STACKLEVEL = 2
+else:
+    CUSTOM_LOGGER_METHOD_STACKLEVEL = 1
```
basfroman:
```suggestion
CUSTOM_LOGGER_METHOD_STACK_LEVEL = 2 if sys.version_info >= (3, 11) else 1
```

bittensor/utils/btlogging/loggingmachine.py
```diff
@@ -447,45 +453,55 @@ def __trace_on__(self) -> bool:
...
-    def trace(self, msg="", prefix="", suffix="", *args, **kwargs):
+    def trace(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
         """Wraps trace message with prefix and suffix."""
         msg = _concat_message(msg, prefix, suffix)
-        self._logger.trace(msg, *args, **kwargs)
+        self._logger.trace(
+            msg,
+            *args,
+            **kwargs,
+            stacklevel=stacklevel + CUSTOM_LOGGER_METHOD_STACKLEVEL,
```
basfroman:
```suggestion
            stacklevel=stacklevel + CUSTOM_LOGGER_METHOD_STACK_LEVEL,
```

basfroman (changes requested):
Pls accept changes and we can move forward with approval

bittensor/utils/btlogging/loggingmachine.py
```diff
@@ -447,45 +450,55 @@ def __trace_on__(self) -> bool:
...
-    def success(self, msg="", prefix="", suffix="", *args, **kwargs):
+    def success(self, msg="", prefix="", suffix="", *args, stacklevel=1, **kwargs):
         """Wraps success message with prefix and suffix."""
         msg = _concat_message(msg, prefix, suffix)
-        self._logger.success(msg, *args, **kwargs)
+        self._logger.success(
+            msg,
+            *args,
+            **kwargs,
+            stacklevel=stacklevel + CUSTOM_LOGGER_METHOD_STACKLEVEL,
```
basfroman:
```suggestion
            stacklevel=stacklevel + CUSTOM_LOGGER_METHOD_STACK_LEVEL,
```

basfroman (changes requested):

basfroman (approved):

thewhaleking (approved):

ibraheem-abe (approved):

PR #2497, opened 2024-11-28
Fix submit_extrinsic timeout

thewhaleking:
`submit_extrinsic` is broken, because exceptions in threads cannot be caught. This fixes that issue, and adds a unit test. Also adds a TODO because I don't think our reconnection logic is solid.

### DIFF
--- a/bittensor/core/extrinsics/utils.py
+++ b/bittensor/core/extrinsics/utils.py
@@ -1,5 +1,7 @@
 """Module with helper functions for extrinsics."""
 
+from concurrent.futures import ThreadPoolExecutor
+import os
 import threading
 from typing import TYPE_CHECKING
 
@@ -12,14 +14,7 @@ if TYPE_CHECKING:
     from substrateinterface import SubstrateInterface, ExtrinsicReceipt
     from scalecodec.types import GenericExtrinsic
 
-
-class _ThreadingTimeoutException(Exception):
-    """
-    Exception raised for timeout. Different from TimeoutException because this also triggers
-    a websocket failure. This exception should only be used with `threading` timer..
-    """
-
-    pass
+EXTRINSIC_SUBMISSION_TIMEOUT = os.getenv("EXTRINSIC_SUBMISSION_TIMEOUT", 200)
 
 
 def submit_extrinsic(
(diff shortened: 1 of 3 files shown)


bittensor/core/extrinsics/utils.py
```diff
@@ -50,55 +45,53 @@ def submit_extrinsic(
...
-            block_hash = substrate.get_block_hash(block_num)
-            try:
-                response = substrate.retrieve_extrinsic_by_hash(
-                    block_hash, f"0x{extrinsic_hash.hex()}"
+        future = executor.submit(submit)
+        if not event.wait(timeout):
+            logging.error("Timed out waiting for extrinsic submission.")
+            after_timeout_block = substrate.get_block()
+
+            for block_num in range(
```
camfairchild, line 82: Not a blocker, but we should rethink this method of getting the block hash.

thewhaleking: I'm open to suggestions. Why is that?

bittensor/core/extrinsics/utils.py
```diff
@@ -12,14 +14,7 @@
...
 
-
-class _ThreadingTimeoutException(Exception):
-    """
-    Exception raised for timeout. Different from TimeoutException because this also triggers
-    a websocket failure. This exception should only be used with `threading` timer..
-    """
-
-    pass
+EXTRINSIC_SUBMISSION_TIMEOUT = int(os.getenv("EXTRINSIC_SUBMISSION_TIMEOUT", 200))
```
camfairchild:
```suggestion
_EXTRINSIC_SUBMISSION_TIMEOUT_default = 200
try:
    EXTRINSIC_SUBMISSION_TIMEOUT = int(os.getenv("EXTRINSIC_SUBMISSION_TIMEOUT", _EXTRINSIC_SUBMISSION_TIMEOUT_default))
except ValueError:
    EXTRINSIC_SUBMISSION_TIMEOUT = _EXTRINSIC_SUBMISSION_TIMEOUT_default
```
Need to handle the int parse here or we crash on import for bad env var

camfairchild: See coment

camfairchild (changes requested):

camfairchild (approved):

PR #2531, opened 2024-12-10
Fixes E2E test chain buffer issues on devnet

ibraheem-abe:
- Adds a background thread to keep reading the lines even after the pattern is found to avoid buffer getting full.


thewhaleking (dismissed):

ibraheem-abe:
Working on devnet - switching back to testnet

thewhaleking (approved):

PR #2534, opened 2024-12-12
Feat/use tx pool for set weights

camfairchild:
This PR addresses an issue with `set_weights` calls where submitting transactions can error if there is already a `set_weights` transaction in the pool. This happens because the nonce and the call method match with a transaction already in the pool. 

We usually grab the next account nonce using the chain's state, and we increment it, therefore submitting the same nonce to the pool before the existing transaction has been included.

We can fix this by using the RPC method `account_nextIndex` which takes into account all the transactions in the pool and gives you the next nonce that the account should use to not conflict with any pending transactions in the pool.


bittensor/utils/async_substrate_interface.py
```diff
@@ -1704,6 +1705,24 @@ def make_payload(id_: str, method: str, params: list) -> dict:
             "payload": {"jsonrpc": "2.0", "method": method, "params": params},
         }
 
+    @lru_cache(maxsize=512)  # RPC methods are unlikely to change often
```
thewhaleking: `lru_cache` won't work here, as it will cache an awaitable, rather than the result. Either remove this or use something like [async-lru](https://github.com/aio-libs/async-lru).

thewhaleking (changes requested):

tests/e2e_tests/test_set_weights.py
```diff
@@ -0,0 +1,158 @@
...
+            local_chain,
+            alice_wallet,
+            "sudo_set_commit_reveal_weights_enabled",
+            False,
+            netuid,
+        ), "Unable to enable commit reveal on the subnet"
+
+        assert subtensor.get_subnet_hyperparameters(
+            netuid=netuid,
+        ).commit_reveal_weights_enabled, "Failed to enable commit/reveal"
```
thewhaleking: I'm confused here. You're setting "sudo_set_commit_reveal_weights_enabled" to False, and then checking to see that it is enabled.

thewhaleking (changes requested):

thewhaleking (approved):

ibraheem-abe (approved):

PR #2555, opened 2024-12-21
set_weights: prevent exception when not registered

heeres:
- The return value of get_uid_for_hotkey_on_subnet was not checked. If the hotkey is not registered, this call will return None, resulting in an exception later on.


ibraheem-abe:
Hey @heeres,

Thank you for your contribution! Can you please update the branch so we can merge this in.

heeres:
Seems to be fixed by commit 3497cd6

Issue #2475, opened 2024-11-25
Get rid of py-substrate-interface


thewhaleking:
### Is your feature request related to a problem? Please describe.

We already rewrote it for async. Let's just do it for sync too.

### Describe the solution you'd like

Rewrite py-substrate-interface in the vein of async substrate.

### Describe alternatives you've considered

Continue using py-substrate-interface

### Additional context

Const asked.

thewhaleking:
Closed by https://github.com/opentensor/async-substrate-interface

PR #2565, opened 2025-01-03
[SDK] Get rid of py-substrate-interface (DO NOT MERGE)

basfroman:
### Context:
The ultimate goal is to focus on async development and no longer depend on the **outdated** `py-substrate-interface` library.

At the same time, at least temporarily, we need to maintain backward compatibility with the current implementation of Bittensor SDK. After merging this PR, we will not have to support 2 versions of the code. It will be enough to implement new functions in the asynchronous version and use a wrapper for a sync calls.

### Advantages:
- Only asynchronous code needs to be supported
- Easy to understand codebase for community developers
- Possibility of simple deprecation marking for sync method/function calls (without affecting the async ones)
- Support for backward compatibility

### AC:
- [x] Implement `coroutine executor`
- [x] Match list of methods of Subtensor and AsyncSubtensor classes
- [x] Keep backwards compatibility for sync codebase
- [x] Use wrapper calls for Subtensor and sync extrinsics
- [x] Add `bittensor.core.chain_data.WeightCommitInfo` class for CRv3 data
- [x] Separate the Sabtenso and AsyncSabtensor modules into logical blocks and place the methods in alphabetical order (_for the convenience of developers using SDK_)
- [x] Implement an **async** `Metagraph` with the ability to use it in synchronous code
- Update return object for following methods (_make backwards compatible with sync methods_):
  - [x] query_constant
  - [x] query_map
  - [x] query_map_subtensor
  - [x] query_module;
  - [x] query_runtime_api
  - [x] query_subtensor
  - [x] state_call	
- [x] Add test to compare sync and async Subtensor methods
- Add/Fix/edit/update existed tests:
  - [x] Unit tests
  - [x] Integration tests
  - [x] e2e tests
 - [ ] Test the functionality on a **testnet** with a full cycle of operations
 
### Related links:
- Github [issue](https://github.com/opentensor/bittensor/issues/2475)
- [Spreadsheet](https://docs.google.com/spreadsheets/d/1RF129dANa_erLtZGWwNN5bFLyzsyrnTSt0vmG69DoXE/edit?usp=sharing) with updated methods and calls

##### NOTE: This description could be updated


bittensor/core/extrinsics/asyncex/unstaking.py
```diff
@@ -127,8 +126,6 @@ async def __do_remove_stake_single(
 
     if success:
         return True
-    else:
-        raise StakeError(format_error_message(err_msg))
```
thewhaleking, line 131: Why is this gone?

bittensor/utils/substrate_interface.py
```diff
@@ -3915,37 +3960,37 @@ def sync_method(*args, **kwargs):
...
+    #     module_name: str,
+    #     constant_name: str,
+    #     block_hash: Optional[str] = None,
+    #     reuse_block_hash: bool = False,
+    # ) -> Optional["ScaleType"]:
+    #     return self.event_loop.run_until_complete(
+    #         self._async_instance.get_constant(
+    #             module_name, constant_name, block_hash, reuse_block_hash
+    #         )
+    #     )
```
thewhaleking: Why are these lines commented out?

bittensor/core/subtensor.py
```diff
@@ -1,584 +1,88 @@
...
+            network=network,
+            config=config,
+            log_verbose=log_verbose,
+            event_loop=self.event_loop,
+        )
 
-        result = self.query_subtensor(param_name, block, [netuid])
-        if result is None or not hasattr(result, "value"):
-            return None
+        self.substrate = SubstrateInterface(substrate=self.async_subtensor.substrate)
```
thewhaleking: This is going to double-initialise AsyncSubstrateInterface. We should not do this.

bittensor/core/extrinsics/asyncex/unstaking.py
basfroman: I'm doing this in `def _do_unstake` in line 82

bittensor/utils/substrate_interface.py
basfroman: Ah, it's not finished yet. I tested how it works without implemented methods in the skeleton. Still deciding whether we should add all the `query_*` methods or keep them dynamic. We'll get back to that.

bittensor/utils/substrate_interface.py
```diff
@@ -3637,7 +3681,7 @@ def concat_hash_len(key_hasher: str) -> int:
                     except Exception as _:
                         if not ignore_decoding_errors:
                             raise
-                        item_value = None
+                        item_value = []
```
thewhaleking: Not sure why this change.

thewhaleking (commented):
Left one comment that I'm not so sure about. Other than that, LGTM.

thewhaleking (approved):

PR #2592, opened 2025-01-21
perf: don't use 2 threads to create FastAPI server

zyzniewski-reef:
### Description of the Change

Looks like `FastAPIThreadedServer.start` is launching 2 threads to start FastAPI/uvicorn server with no real reason.

### Quantitative Performance Benefits

No performance tests were run - it's a simple design change.

### Possible Drawbacks

None?

### Verification Process

Created unittest. Ran Axon and checked FastAPI is listening in accepting requests.

### Release Notes

- Improve Axon's resource management

### DIFF
--- a/bittensor/core/axon.py
+++ b/bittensor/core/axon.py
@@ -142,14 +142,6 @@ class FastAPIThreadedServer(uvicorn.Server):
             self.should_exit = True
             thread.join()
 
-    def _wrapper_run(self):
-        """
-        A wrapper method for the :func:`run_in_thread` context manager. This method is used internally by the ``start`` method to initiate the server's execution in a separate thread.
-        """
-        with self.run_in_thread():
-            while not self.should_exit:
-                time.sleep(1e-3)
-
     def start(self):
         """
         Starts the FastAPI server in a separate thread if it is not already running. This method sets up the server to handle HTTP requests concurrently, enabling the Axon server to efficiently manage incoming network requests.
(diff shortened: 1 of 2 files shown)


bittensor/core/axon.py
```diff
@@ -119,37 +118,6 @@ class FastAPIThreadedServer(uvicorn.Server):
     should_exit: bool = False
     is_running: bool = False
 
-    def install_signal_handlers(self):
```
zyzniewski-reef, line 122: `install_signal_handlers` not needed anymore - uvicorn changed the way it handles signals

tests/unit_tests/test_axon.py
```diff
@@ -785,3 +788,44 @@ async def forward_fn(synapse: streaming_synapse_cls):
...
+            app,
+        ),
+    )
+    server.start()
+
+    server_started.wait()
+
+    assert server.is_running is True
+
+    client = httpx.Client(
```
thewhaleking: HTTPX is not a dependency. Instead use aiothttp.

bittensor/core/axon.py
```diff
@@ -119,37 +118,6 @@ class FastAPIThreadedServer(uvicorn.Server):
...
-        """
-        thread = threading.Thread(target=self.run, daemon=True)
-        thread.start()
-        try:
-            while not self.started:
-                time.sleep(1e-3)
-            yield
-        finally:
-            self.should_exit = True
-            thread.join()
```
thewhaleking: Generally speaking, we can't entirely remove public methods/functions in a minor release. Instead deprecate these with `warnings`, and basically just change them to do nothing.

thewhaleking (changes requested):

tests/unit_tests/test_axon.py
zyzniewski-reef: https://github.com/opentensor/bittensor/blob/7ccb07dba4a8db738d15a050b4f13158e1fba05b/requirements/dev.txt#L16
but sure I can switch to use `aiohttp`

thewhaleking (approved):
This is a great catch!

PR #2609, opened 2025-01-29
feat: use bt_decode

zyzniewski-reef:
requires:
- https://github.com/opentensor/async-substrate-interface/pull/15
- https://github.com/opentensor/subtensor/pull/832


bittensor/core/chain_data/info_base.py
```diff
@@ -0,0 +1,27 @@
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import Any, TypeVar
+
+import munch
```
thewhaleking, line 5: We really don't want to do munch

thewhaleking (changes requested):

PR #2613, opened 2025-01-30
fix integration metagraph test

basfroman:
we need to improve this test. but it's fine for now


bittensor/core/async_subtensor.py
```diff
@@ -1084,7 +1084,7 @@ async def get_delegate_identities(
             )
 
             all_delegates_details = {}
-            async for ss58_address, identity in identities_info:
+            for ss58_address, identity in identities_info:
```
thewhaleking: This is important

thewhaleking (changes requested):

bittensor/core/async_subtensor.py
basfroman: Exactly so. In this code, the use of `async` loop is redundant which complicates testing and that's it. Inside this loop there are sync function calls.

thewhaleking: Why is it redundant?

thewhaleking: Take this snippet for example:

```python
async def main():
    async with AsyncSubtensor("test") as subtensor:
        print(await subtensor.get_delegate_identities())
```

It won't work, because `AsyncQueryMapResult` has no `__next__` method.
(with the current `async-substrate-interface`

basfroman: Ah, yeah. This is `AsyncQueryMapResult` with async iterator instead of `QueryMapResult` with sync one.
Make sense. I checked all calls inside `for` loop, but not the iterated object. Let's bring it back and replace mocked object with `AsyncMock`

thewhaleking (approved):

PR #2645, opened 2025-02-07
fix: typos in documentation files

maximevtush:


thewhaleking:
PRs should be based off staging. This is fixed in staging, and `numpy.arrange` is not a function.

PR #2658, opened 2025-02-11
add name and symbol fields to metagraph

basfroman:


thewhaleking (approved):
Tested with Numpy and Torch for sync.

ibraheem-abe (approved):

PR #2666, opened 2025-02-12
Changes the logging level for unexpected header keys to trace

thewhaleking:
It was recently changed to warning, but we've received complaints that this fills up logs.

### DIFF
--- a/bittensor/core/synapse.py
+++ b/bittensor/core/synapse.py
@@ -808,7 +808,8 @@ class Synapse(BaseModel):
                     logging.error(f"Error while parsing 'input_obj' header {key}: {e}")
                     continue
             else:
-                logging.warning(f"Unexpected header key encountered: {key}")
+                # setting this to warning fills up logs unnecessarily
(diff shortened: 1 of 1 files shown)


bittensor/core/synapse.py
```diff
@@ -808,7 +808,8 @@ def parse_headers_to_inputs(cls, headers: dict) -> dict:
                     logging.error(f"Error while parsing 'input_obj' header {key}: {e}")
                     continue
             else:
-                logging.warning(f"Unexpected header key encountered: {key}")
+                # setting this to warning fills up logs unnecessarily
+                logging.trace(f"Unexpected header key encountered: {key}")
```
basfroman, line 812: Is there any reason don't use debug level?

thewhaleking: > Is there any reason don't use debug level?

Trace is lower, and a number of people have expressed annoyance at `debug` level logs also filling things up.

basfroman (approved):

PR #2676, opened 2025-02-18
Add compatibility for read-only systems

Arthurdw:
### Bug

Read-only systems are not able to utilise the bittensor package. Even if they dont need the functionalities which require io.

Why is this a bug? This wasn't an issue in previous versions due to code differences.

### Description of the Change

This adds a env variable READ_ONLY, which can be set to 1 and as such fixes the above mentioned issues.

### Alternate Designs

Use cli params.

### Possible Drawbacks

None?

### Verification Process

Not applicable.

### Release Notes

- Allow read-only systems to utilize the bittensor package by setting `READ_ONLY=1` in the environment.


thewhaleking:
~~Can you base this on staging? Looks good to me.~~

Nevermind, I can.

thewhaleking (approved):

PR #2702, opened 2025-02-25
Dendrite `__del__` method fix

thewhaleking:
Adds a try:except catch for the `__del__` method, as `self.close_session()` only works if not run from within an already-running event loop.

It's already explained throughout the module to use async context managers. The problem arises when the object is attempt to be garbage-collected, and thus the `__del__` method is called.

### DIFF
--- a/bittensor/core/dendrite.py
+++ b/bittensor/core/dendrite.py
@@ -877,7 +877,10 @@ class DendriteMixin:
             # ... some operations ...
             del dendrite  # This will implicitly invoke the __del__ method and close the session.
         """
-        self.close_session()
+        try:
+            self.close_session()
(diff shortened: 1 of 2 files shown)


bittensor/core/dendrite.py
```diff
@@ -877,7 +877,10 @@ def __del__(self):
             # ... some operations ...
             del dendrite  # This will implicitly invoke the __del__ method and close the session.
         """
-        self.close_session()
+        try:
+            self.close_session()
+        except RuntimeError:
+            pass
```
basfroman: I'd add debug log

basfroman (dismissed):

basfroman (approved):

Issue #2715, opened 2025-03-03
Update delegate identities function


ibraheem-abe:
Currently it is fetching data from the old identities storage function - we need to migrate it to use the new V2 identities (chainidentities).

zyzniewski-reef:
What shall we do about fetching [delegates.json](https://github.com/opentensor/bittensor-delegates/commits/main/public/delegates.json) - it seems still used

PR #2718, opened 2025-03-04
fix(2715): use ChainIdentity for identities

zyzniewski-reef:
https://github.com/opentensor/bittensor/issues/2715


ibraheem-abe (approved):

ibraheem-abe:
This also goes away from fetching github delegates. In the next step, we should also deprecate this from the CLI

PR #2731, opened 2025-03-07
Feat: set_delegate_take

zyzniewski-reef:
https://github.com/opentensor/bittensor/issues/2716


bittensor/core/async_subtensor.py
```diff
@@ -1111,7 +1115,7 @@ async def get_delegate_take(
         block: Optional[int] = None,
         block_hash: Optional[str] = None,
         reuse_block: bool = False,
-    ) -> Optional[float]:
+    ) -> float:
```
zyzniewski-reef, line 1118: `get_delegate_take` always returns value even for non-existing hotkey (InitialDefaultDelegateTake)

bittensor/core/async_subtensor.py
```diff
@@ -3267,6 +3276,71 @@ async def root_set_weights(
...
+    async def set_delegate_take(
+        self,
+        wallet: "Wallet",
+        hotkey_ss58: str,
+        take: float,
+        wait_for_inclusion=True,
+        wait_for_finalization=True,
+    ) -> None:
+        """
+        Sets the delegate 'take' percentage for a nueron identified by its hotkey.
```
thewhaleking:
```suggestion
        Sets the delegate 'take' percentage for a neuron identified by its hotkey.
```

thewhaleking (changes requested):

bittensor/core/subtensor.py
```diff
@@ -2076,6 +2142,7 @@ def sign_and_send_extrinsic(
             wait_for_inclusion (bool): whether to wait until the extrinsic call is included on the chain
             wait_for_finalization (bool): whether to wait until the extrinsic call is finalized on the chain
             sign_with: the wallet's keypair to use for the signing. Options are "coldkey", "hotkey", "coldkeypub"
+            raise_error: doesn't return error but raises SubstrateRequestException exception
```
```suggestion
            raise_error: raises relevant exception rather than returning `False` if unsuccessful.
```

bittensor/core/async_subtensor.py
```diff
@@ -2758,6 +2760,7 @@ async def sign_and_send_extrinsic(
             wait_for_inclusion (bool): whether to wait until the extrinsic call is included on the chain
             wait_for_finalization (bool): whether to wait until the extrinsic call is finalized on the chain
             sign_with: the wallet's keypair to use for the signing. Options are "coldkey", "hotkey", "coldkeypub"
+            raise_error: doesn't return error but raises SubstrateRequestException exception
```
thewhaleking:
```suggestion
            raise_error: raises relevant exception rather than returning `False` if unsuccessful.
```

thewhaleking (changes requested):
Few suggestions. Overall looks good, but I think we should adjust that the new methods should take `raise_error` bool args, with default `False` to keep consistency with the rest of the codebase. Otherwise you have people writing different styles of code to handle two types of situations (and the unlock keys should be done in the same way, though they should also take a `raise_error` arg). I agree with your thinking that we should raise errors rather than return success bool for an SDK, but we must also keep consistency in code style throughout. When we make a plan for 10.0, we'll add deprecation warnings to all `raise_error=False` situations.

thewhaleking (approved):

PR #2732, opened 2025-03-10
Fix: burned_register supports root subnet (netuid=0 param)

zyzniewski-reef:
https://github.com/opentensor/bittensor/issues/2716


bittensor/core/async_subtensor.py
```diff
@@ -2901,6 +2901,14 @@ async def burned_register(
             bool: ``True`` if the registration is successful, False otherwise.
         """
         async with self:
```
thewhaleking, line 2917: I know this is unrelated to this PR, but why do we do this?

bittensor/core/extrinsics/asyncex/root.py
```diff
@@ -62,6 +63,38 @@ async def root_register_extrinsic(
...
+                param_name="Burn",
+                netuid=netuid,
+                block_hash=block_hash,
+            ),
+            subtensor.get_balance(
+                wallet.coldkeypub.ss58_address,
+                block_hash=block_hash,
+            ),
+        )
+    except TypeError as e:
```
thewhaleking: This is actually a fairly difficult one. `asyncio.gather` returns errors in an odd non-standard way, where doing it like this can usually catch fine, but sometimes will not. You'll instead want to use `asyncio.gather(*cos, return_exceptions=True)` and then check to see if the returns are the exception type. It's not great, but it's the spec.

thewhaleking (changes requested):
Overall LGTM, just fix the gather catch

bittensor/core/extrinsics/asyncex/root.py
zyzniewski-reef: I don't really understand. Can you provide an example of "non-catchable" exception thrown from `asyncio.gather`?

zyzniewski-reef: I think `TypeError` very likely can be safely removed anyway (I've copied it from existing `root_register` [method code](https://github.com/opentensor/bittensor/pull/2732/files#diff-87f2770377552e515f60ce0d2e31a3a9bb129bffa240ab2133e3233d3e22f6c3L3210)) as it seems to be a leftover from this: https://github.com/opentensor/bittensor/commit/835dfdb876226f80cd7b17ec4b1b688c8037c95b#diff-87f2770377552e515f60ce0d2e31a3a9bb129bffa240ab2133e3233d3e22f6c3R1141-R1145

bittensor/core/async_subtensor.py
zyzniewski-reef: probably to make sure Subtensor/Substrate is properly initialized?

ibraheem-abe (approved):

thewhaleking (approved):

PR #2739, opened 2025-03-12
Tests: increse test_incentive timeout + fix sudo_set_weights_set_rate_limit

zyzniewski-reef:


tests/e2e_tests/test_set_weights.py
```diff
@@ -84,15 +83,19 @@ async def test_set_weights_uses_next_nonce(local_chain, subtensor, alice_wallet)
...
-        assert sudo_set_hyperparameter_values(
+        status, error = sudo_set_admin_utils(
             local_chain,
             alice_wallet,
             call_function="sudo_set_weights_set_rate_limit",
             call_params={"netuid": netuid, "weights_set_rate_limit": "0"},
             return_error_message=True,
         )
 
+        assert error is None
```
thewhaleking, line 93: Why would error ever be `None` for these? Doesn't this fn return (bool, str)?

zyzniewski-reef: error_message is actually a `dict` - I've updated function's signature and docs

thewhaleking (dismissed):

thewhaleking (approved):

PR #2746, opened 2025-03-13
Fix E2E test_incentive by waiting till start of the very next epoch

zyzniewski-reef:
I've run test_incentive around 20 times on CI/CD and it seems stable now.
I think the problem was down to [time.sleep(5)](https://github.com/opentensor/subnet-template/blob/main/validator.py#L160) in Validator which is bigger than epoch's duration for fast-blocks (10 blocks * 0.25s) so it's prone to skip (oversleep) one.
Now E2E waits for the first successful set_weights and checks results immediately


zyzniewski-reef:
I've run `test_incentive` around 20 times on CI/CD and it seems stable now.
I think the problem is down to [time.sleep(5)](https://github.com/opentensor/subnet-template/blob/main/validator.py#L160) in Validator which is bigger than epoch's duration for fast-blocks (10 blocks * 0.25s) so it's prone to skip (oversleep) one.
Now E2E tests for the first successful set_weights and checks results immediately

ibraheem-abe (approved):

PR #2764, opened 2025-03-21
Improve logic in Balance magic methods

basfroman:


bittensor/utils/balance.py
```diff
@@ -6,6 +6,18 @@
...
+def _check_currencies(self, other):
+    """Checks that Balance objects have the same netuids to perform arithmetic operations."""
+    if self.netuid != other.netuid:
+        warnings.simplefilter("default", DeprecationWarning)
+        warnings.warn(
+            "Balance objects must have the same netuid (Alpha currency) to perform arithmetic operations. "
+            f"First balance is `{self}`.  Second balance is `{other}`. ",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
```
ibraheem-abe: Suggestion to briefly show how to create new balance object with correct netuid in the deprecation warning

ibraheem-abe (dismissed):
LGTM overall

thewhaleking (commented):
A bit confused here because it seems it still does the arithmetic, but just displays a warning. Is this what we want?

basfroman:
> A bit confused here because it seems it still does the arithmetic, but just displays a warning. Is this what we want?

Exactly. Right now we can't brake users' scripts. That would cause hate. We decided to display a warning. Perhaps by version 10 we'll decide to change this logic. But we have to let users know that the operation is being performed with an error if the units don't match.

thewhaleking (dismissed):

ibraheem-abe (approved):

thewhaleking (approved):

PR #2775, opened 2025-03-26
AsyncSubtensor E2E tests

zyzniewski-reef:
All E2E tasks are asyncio-compatible and we wrap (sync) Subtensor and simply pass all calls directly to it and return in blocking mode

requires:
- https://github.com/opentensor/bittensor/pull/2771
- https://github.com/opentensor/bittensor/pull/2772


tests/e2e_tests/test_set_weights.py
```diff
@@ -9,8 +9,17 @@
...
     wait_epoch,
 )
+from tests.e2e_tests.utils.e2e_test_utils import SyncSubtensor
 
 
+@pytest.mark.parametrize(
+    "subtensor",
+    [
+        SyncSubtensor,
+        # TODO skip AsyncSubtensor as Nonce doesn't work properly
```
thewhaleking, line 20: What do you mean Nonce doesn't work properly?

basfroman:
Implemented in SDKv10

PR #2777, opened 2025-03-27
split process_weights_for_netuid

andreea-popescu-reef:
### Description of the Change

Refactor `process_weights_for_netuid` so that the subtensor calls (for getting `min_allowed_weights` and `max_weight_limit` params and optionally the metagraph) remain in the original function body while the rest of the logic moves to a new `process_weights` function. This allows users to use directly `process_weights` without extra subtensor calls. No functional changes.

### Alternate Designs

N/A

### Possible Drawbacks

N/A

### Verification Process

checked locally

### Release Notes

Introduced `process_weights` for direct usage of `process_weights_for_netuid` without subtensor calls.

### DIFF
--- a/bittensor/utils/weight_utils.py
+++ b/bittensor/utils/weight_utils.py
@@ -245,7 +245,6 @@ def process_weights_for_netuid(
     """
 
     logging.debug("process_weights_for_netuid()")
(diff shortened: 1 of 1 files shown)


thewhaleking (approved):
Approved but should be merged through #2798

PR #2782, opened 2025-04-01
add py3.13 to tests in CI

basfroman:


basfroman:
remove 3.13 for test. bittensor  requere bt-wallet with py 3.13 support -> numpy has a problem with py 3.13 during compilation

PR #2781, opened 2025-03-29
Add drand-commitments

basfroman:
Added Subtensor methods (async/sync):
- `get_revealed_commitment` get revealed commitments based on netuid and hotkey
- `get_all_revealed_commitments` get all reveled commitments in subnet based in netuid
- `set_reveal_commitment` set commitments to specific subnet with `blocks_until_reveal` (number of blocks from now after which the data will be revealed)
- a few helper functions
- e2e test for mentioned methods

PR #2787, opened 2025-04-04
Improvement and fix for https://github.com/opentensor/bittensor/pull/2781

basfroman:
Added Subtensor methods (async/sync):
- `get_revealed_commitment` get revealed commitments based on netuid and uid
- `get_revealed_commitment_by_hotkey` get revealed commitments based on netuid and hotkey
- `get_all_revealed_commitments` get all reveled commitments in subnet based in netuid
- `set_reveal_commitment` set commitments to specific subnet with `blocks_until_reveal` (number of blocks from now after which the data will be revealed)
- a few helper functions
- e2e test for mentioned methods

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -1,7 +1,7 @@
 import asyncio
 import copy
-from datetime import datetime, timezone
 import ssl
+from datetime import datetime, timezone
 from functools import partial
 from typing import Optional, Any, Union, Iterable, TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


bittensor/core/async_subtensor.py
```diff
@@ -1092,15 +1096,50 @@ async def get_revealed_commitment(
             block_hash=block_hash,
             reuse_block=reuse_block,
         )
-        return decode_revealed_commitment(query)
+        if query is None:
+            return None
+        return tuple(decode_revealed_commitment(pair) for pair in query)
```
thewhaleking, line 1101: Is there a reason to use a tuple here rather than the conventional list?

bittensor/core/subtensor.py
```diff
@@ -787,6 +788,7 @@ def get_commitment(self, netuid: int, uid: int, block: Optional[int] = None) ->
             return ""
 
         metadata = cast(dict, get_metadata(self, netuid, hotkey, block))
+        print(metadata)
```
```suggestion
```

bittensor/core/async_subtensor.py
basfroman: yes, object should be immutable

thewhaleking (approved):

PR #2804, opened 2025-04-09
9.3.0: Bumps version and changelog

ibraheem-abe:
- Also bumps btwallet


basfroman (approved):

thewhaleking (approved):
Why do we have so many places for "version"?

PR #2829, opened 2025-04-15
[WIP] Fallback Chains

### LINKED ISSUE #2533: Retry with list of chains

Rhef mentioned that it would be great if Subtensor could take a list of chain endpoints or networks, which would simply switch to the next one if there's an issue with the chain as fallbacks.

Accept chain and fallbacks.

_No response_

### Additional context

Will be interesting to implement.

thewhaleking:
Resolves #2533 

Requires https://github.com/opentensor/async-substrate-interface/pull/96

Currently Proof of concept for fallback networks. Not tested or working with Async. Not perfect with sync.


basfroman:
`RetrySubstrate` class should be implemented in `async-substrate-interface` package as it is a wrapper for `Async/SubstrateInterface` classes.
Also, I'd implemented new class like
```py
from dataclasses import dataclass

@dataclass()
class RetrySubstrateArgs:
    main_endpoint: str  # initial endpoint
    fallback_chains: list[str]  # endpoints for reconnection
    retry_attempts: int  # amount of connections before fail (if 0 then infinite)
```

`Subtensor` class will accept some new argument. For example:
```py
safety_substrate: Optional["RetrySubstrateArgs"] = None`
```
for `__init__()` and from this will decide something like
```py
if safety_substrate and is_instance(safety_substrate, RetrySubstrateArgs):
  self.substrate = RetrySubstrate(..., safety_substrate: "RetrySubstrateArgs" = safety_substrate)
else:
  self.substrate = SubstrateInterface(...)
```
Since the use of `safety_substrate` is optional, it is OK to have a combined argument instead of a primitive. In this case, the user decides which option suits him better, SDK does not contain any non-native logic, everything is implemented cleanly and in Python style. Also, SDK does not need to constantly monitor changes of `Async/SubstrateInterface classes`.

Let's move logic `RetrySubstrate` from SDK to `async-substrate-interface`. 
Only `1` argument and `1` if condition for its processing will be added to the `Async/Subtensor` class. (small changes)

thewhaleking:
> `RetrySubstrate` class should be implemented in `async-substrate-interface` package as it is a wrapper for `Async/SubstrateInterface` classes. Also, I'd implemented new class like
> 
> ```python
> from dataclasses import dataclass
> 
> @dataclass()
> class RetrySubstrateArgs:
>     main_endpoint: str  # initial endpoint
>     fallback_chains: list[str]  # endpoints for reconnection
>     retry_attempts: int  # amount of connections before fail (if 0 then infinite)
> ```
> 
> `Subtensor` class will accept some new argument. For example:
> 
> ```python
> safety_substrate: Optional["RetrySubstrateArgs"] = None`
> ```
> 
> for `__init__()` and from this will decide something like
> 
> ```python
> if safety_substrate and is_instance(safety_substrate, RetrySubstrateArgs):
>   self.substrate = RetrySubstrate(..., safety_substrate: "RetrySubstrateArgs" = safety_substrate)
> else:
>   self.substrate = SubstrateInterface(...)
> ```
> 
> Since the use of `safety_substrate` is optional, it is OK to have a combined argument instead of a primitive. In this case, the user decides which option suits him better, SDK does not contain any non-native logic, everything is implemented cleanly and in Python style. Also, SDK does not need to constantly monitor changes of `Async/SubstrateInterface classes`.
> 
> Let's move logic `RetrySubstrate` from SDK to `async-substrate-interface`. Only `1` argument and `1` if condition for its processing will be added to the `Async/Subtensor` class. (small changes)

Will move majority of that logic to https://github.com/opentensor/async-substrate-interface/pull/100

thewhaleking:
Moved the entire thing to https://github.com/opentensor/async-substrate-interface/pull/100

PR #2830, opened 2025-04-16
Fix and improve e2e tests after `start call` new limit in subtensor

basfroman:
### Includes:
- main issue with start_call related fixes
- updated `set-output` in GH workflow
- added e2e tests retry mechanism with 2 attempts for each test

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -753,7 +753,7 @@ class AsyncSubtensor(SubtensorMixin):
         return_val = (
             False
             if result is None
-            # not a genesis key of fast blocks
+            # not the default key (0x0)
             else result != "5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM"
         )
         return return_val
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -753,6 +753,7 @@ class AsyncSubtensor(SubtensorMixin):
         return_val = (
             False
             if result is None
+            # not a genesis key of fast blocks
             else result != "5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM"
         )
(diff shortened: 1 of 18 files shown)


bittensor/core/async_subtensor.py
```diff
@@ -753,6 +753,7 @@ async def does_hotkey_exist(
         return_val = (
             False
             if result is None
+            # not a genesis key of fast blocks
```
camfairchild:
```suggestion
            # not the default key (0x0)
```

bittensor/core/subtensor.py
```diff
@@ -517,6 +517,7 @@ def does_hotkey_exist(self, hotkey_ss58: str, block: Optional[int] = None) -> bo
         return_val = (
             False
             if result is None
+            # not a genesis key of fast blocks
```
camfairchild:
```suggestion
            # not the default key (0x0)
```

tests/e2e_tests/test_commitment.py
```diff
@@ -7,34 +7,44 @@
...
-def test_commitment(local_chain, subtensor, alice_wallet):
+def test_commitment(local_chain, subtensor, alice_wallet, dave_wallet):
+    dave_subnet_netuid = 2
+    assert subtensor.register_subnet(dave_wallet)
+    assert subtensor.subnet_exists(dave_subnet_netuid), (
+        "Subnet wasn't created successfully"
+    )
+
+    subtensor.wait_for_block(subtensor.block + 20)
+    status, message = subtensor.start_call(dave_wallet, dave_subnet_netuid, True, True)
```
camfairchild: Why not make a test helper for register subnet and start call

basfroman: > Why not make a test helper for register subnet and start call

We've discussed it already. I want to add this feature in an elegant way later

basfroman: Well, it's added

camfairchild (approved):

PR #2862, opened 2025-05-06
Introduce `SubtensorApi` interface for SDK

basfroman:
Depends: https://github.com/opentensor/async-substrate-interface/pull/100 (until this PR is merged, most tests fail)

# `SubtensorApi` – Usage Guide

The `SubtensorApi` is a high-level, unified interface to interact with the Bittensor blockchain. It wraps both the synchronous and asynchronous `Subtensor` implementations, providing modular access to chain subsystems like wallets, delegates, neurons, and more.

---

## 1. Basic Synchronous Usage (default)

```python
import bittensor as bt

sub = bt.SubtensorApi()

print(sub.block)  # Get current block number
print(sub.delegates.get_delegate_identities())
sub.chain.tx_rate_limit()
```

---

## 2. Asynchronous Usage

```python
import bittensor as bt
import asyncio

async def main():
    sub = bt.SubtensorApi(async_subtensor=True)
    async with sub:
        print(await sub.block)
        print(await sub.delegates.get_delegate_identities())
        await sub.chain.tx_rate_limit()

asyncio.run(main())
```

---

## 3. Legacy Method Support

You can enable all legacy methods from the original `Subtensor` class directly on this interface:

```python
import bittensor as bt

sub = bt.SubtensorApi(legacy_methods=True)
print(sub.bonds(0))  # Classic method usage
```

---

## 4. Retry and Fallback RPC Nodes

Enable redundancy and resilience with fallback chains and retry logic:

```python
import bittensor as bt

sub = bt.SubtensorApi(
    "local",
    fallback_chains=[
        "wss://fallback1.taonetwork.com:9944",
        "wss://lite.sub.latent.to:443",
    ],
    retry_forever=True,
)
print(sub.block)
```

---

## 5. Mock Mode for Testing

Use `mock=True` to simulate the interface without connecting to the blockchain:

```python
import bittensor as bt

sub = bt.SubtensorApi(mock=True)
print(sub)  # Output: "<Network: None, Chain: Mock, Sync version>"
```

---

## 6. Subsystem Overview

All methods are grouped into logical modules for better organization and readability.   Some methods may belong to more than one group if they span multiple functional areas. This does not compromise the internal logic — rather, it enhances discoverability and cohesion. Method equivalence between `SubtensorApi` and the original `Subtensor` is automatically verified by test coverage on every pull request (PR).

The following properties are modular access points for specialized subsystems:

| Property      | Description                          |
|---------------|--------------------------------------|
| `chain`       | Blockchain interaction methods       |
| `commitments` | Commitment and reveal logic          |
| `delegates`   | Delegate management tools            |
| `extrinsics`  | Transaction construction and signing |
| `metagraphs`  | Metagraph data and operations        |
| `neurons`     | Neuron-level APIs                    |
| `queries`     | General query endpoints              |
| `stakes`      | Staking operations                   |
| `subnets`     | Subnet access and management         |
| `wallets`     | Wallet creation, import/export       |

---

## 7. Custom Configuration Support

You can pass a pre-built `Config` object:

```python
import argparse
import bittensor as bt
parser = argparse.ArgumentParser('Miner')
bt.SubtensorApi.add_args(parser)
config = bt.config(parser)
sub = bt.SubtensorApi(config=config)

print(sub)
```

---

## 8. Using as a Context Manager (Synchronous)

```python
import bittensor as bt

with bt.SubtensorApi() as sub:
    print(sub.block)
```

---

## 9. Using as a Context Manager (Asynchronous)

```python
import bittensor as bt
import asyncio

async def main():
    async with bt.SubtensorApi(async_subtensor=True) as sub:
        print(await sub.block)

asyncio.run(main())
```


bittensor/core/async_subtensor.py
```diff
@@ -283,6 +283,38 @@ async def get_hyperparameter(
...
+            _mock: Whether this is a mock instance. Mainly just for use in testing.
+
+        Returns:
+            the instance of the SubstrateInterface or RetrySyncSubstrate class.
+        """
+        if fallback_chains or retry_forever:
+            return RetryAsyncSubstrate(
+                url=self.chain_endpoint,
+                fallback_chains=fallback_chains,
+                retry_forever=retry_forever,
```
thewhaleking, line 308:
```suggestion
                ss58_format=SS58_FORMAT,
                type_registry=TYPE_REGISTRY,
                retry_forever=retry_forever,
                use_remote_preset=True,
                chain_name="Bittensor",
```

bittensor/core/subtensor.py
```diff
@@ -163,11 +163,41 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            retry_forever (bool): Whether to retry forever on connection errors. Defaults to `False`.
+            _mock: Whether this is a mock instance. Mainly just for use in testing.
+
+        Returns:
+            the instance of the SubstrateInterface or RetrySyncSubstrate class.
+        """
+        if fallback_chains or retry_forever:
+            return RetrySyncSubstrate(
+                url=self.chain_endpoint,
+                fallback_chains=fallback_chains,
```
thewhaleking, line 192:
```suggestion
                url=self.chain_endpoint,
                ss58_format=SS58_FORMAT,
                type_registry=TYPE_REGISTRY,
                use_remote_preset=True,
                chain_name="Bittensor",
                fallback_chains=fallback_chains,
```

bittensor/core/subtensor_api/__init__.py
```diff
@@ -0,0 +1,220 @@
...
+
+if TYPE_CHECKING:
+    from bittensor.core.config import Config
+
+
+class SubtensorApi:
+    """Subtensor API class.
+
+    Arguments:
+        network: The network to connect to. Defaults to `None` -> `finney`.
```
thewhaleking:
```suggestion
        network: The network to connect to. Defaults to `None` -> "finney".
```
Backticks should be used to denote valid Python.

bittensor/core/subtensor_api/__init__.py
```diff
@@ -0,0 +1,220 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    async def __aenter__(self):
+        return await self._subtensor.__aenter__()
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        await self.substrate.close()
```
thewhaleking: This will break since it's valid to use `async with SubtensorApi()` because of this with either Subtensor or AsyncSubtensor, however only AsyncSubtensor has this. Maybe check the `self.is_async` flag and raise a `NotImplementedError` if called without it.

bittensor/core/subtensor_api/chain.py
```diff
@@ -0,0 +1,19 @@
+from typing import Union
+from bittensor.core.subtensor import Subtensor as _Subtensor
+from bittensor.core.async_subtensor import AsyncSubtensor as _AsyncSubtensor
```
thewhaleking, line 3: Not a requirement for this PR, but you might consider defining 
```python
SubtensorLike = typing.Type[typing.Union[Subtensor, AsyncSubtensor]]
```

in utils.py or something, and then just import this from there.

thewhaleking (changes requested):
Small changes, otherwise g2g

bittensor/core/subtensor_api/__init__.py
basfroman: I don't want to create a foolproof
If call synchronous and asynchronous context manager as a given, there will be no errors

bittensor/core/subtensor_api/chain.py
basfroman: I avoid any public objects as much as possible in this module.

bittensor/core/subtensor_api/__init__.py
basfroman: You know, yeah, let's add it :)

thewhaleking (approved):

PR #2865, opened 2025-05-08
Fix `get_next_epoch_start_block`

basfroman:


camfairchild (approved):

surcyf123:
@basfroman if you use LastMechansimStepBlock instead of BlocksSinceLastStep, you don't have to do current_block - X on every step.

basfroman:
> @basfroman if you use LastMechansimStepBlock instead of BlocksSinceLastStep, you don't have to do current_block - X on every step.

I'll take a look it tomorrow. Thank you for this tip.

PR #2880, opened 2025-05-15
Rename argument `fallback_chains` to `fallback_endpoints`

basfroman:
The name of the `fallback_chains` argument in its description confuses users. They perceive it as a list of network names, not endpoints.
The current changes bring clarity and transparency.

### DIFF
--- a/bittensor/core/subtensor.py
+++ b/bittensor/core/subtensor.py
@@ -190,7 +190,7 @@ class Subtensor(SubtensorMixin):
                 type_registry=TYPE_REGISTRY,
                 use_remote_preset=True,
                 chain_name="Bittensor",
-                fallback_endpoints=fallback_endpoints,
+                fallback_chains=fallback_endpoints,
                 retry_forever=retry_forever,
                 _mock=_mock,
             )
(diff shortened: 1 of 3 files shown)


ibraheem-abe (dismissed):

bittensor/core/subtensor.py
```diff
@@ -167,28 +169,28 @@ def close(self):
...
-        if fallback_chains or retry_forever:
+        if fallback_endpoints or retry_forever:
             return RetrySyncSubstrate(
                 url=self.chain_endpoint,
                 ss58_format=SS58_FORMAT,
                 type_registry=TYPE_REGISTRY,
                 use_remote_preset=True,
                 chain_name="Bittensor",
-                fallback_chains=fallback_chains,
+                fallback_endpoints=fallback_endpoints,
```
```suggestion
                fallback_chains=fallback_endpoints,
```

thewhaleking (changes requested):
Well, two things. The first is the suggestion.
The second is that we can't rename it without a major version change.

thewhaleking (approved):

PR #2886, opened 2025-05-20
Add defaults for endpoint and network from local env

basfroman:
Adds:
- local env variable`BT_NETWORK`
- local env variable `BT_CHAIN_ENDPOINT`

Logic:
Just add local env variable `BT_NETWORK` with known settings.NETWORKS (`finney`, `test`, `archive`, `local`, `subvortex`, `latent-lite`) or `BT_CHAIN_ENDPOINT` with custom endpoint.

### DIFF
--- a/bittensor/core/settings.py
+++ b/bittensor/core/settings.py
@@ -121,8 +121,8 @@ DEFAULTS = munchify(
             "maxsize": int(_BT_PRIORITY_MAXSIZE) if _BT_PRIORITY_MAXSIZE else 10,
         },
         "subtensor": {
-            "chain_endpoint": DEFAULT_ENDPOINT,
-            "network": DEFAULT_NETWORK,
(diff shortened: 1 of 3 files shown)


ibraheem-abe (approved):

bittensor/core/settings.py
```diff
@@ -121,8 +121,8 @@
             "maxsize": int(_BT_PRIORITY_MAXSIZE) if _BT_PRIORITY_MAXSIZE else 10,
         },
         "subtensor": {
-            "chain_endpoint": DEFAULT_ENDPOINT,
-            "network": DEFAULT_NETWORK,
+            "chain_endpoint": os.getenv("BT_CHAIN_ENDPOINT") or DEFAULT_ENDPOINT,
```
thewhaleking, line 124: We use `BT_SUBTENSOR_CHAIN_ENDPOINT` on line 32. The logic should be thought-out a bit more, because that is used for "local".

PR #2909, opened 2025-06-06
Retry archive node support

thewhaleking:
Companion to https://github.com/opentensor/async-substrate-interface/issues/114, which is released


basfroman (changes requested):
`SubtensorApi` needs to have new `archive_endpoints` parameter too.

thewhaleking:
> `SubtensorApi` needs to have new `archive_endpoints` parameter too.

Good catch

basfroman (approved):

PR #2918, opened 2025-06-12
[LINT] Improved type hints

Arthurdw:
### Description of the Change

Improved the developer type hints correctness, this pr should not contain any logic changes.

### Alternate Designs

n/a

### Possible Drawbacks

none

### Verification Process

Less lsp warnings :D

### Branch Acknowledgement

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -3338,8 +3338,8 @@ class AsyncSubtensor(SubtensorMixin):
             datetime object for the timestamp of the block
         """
         res = await self.query_module(
-            "timestamp",
-            "now",
+            "Timestamp",
+            "Now",
             block=block,
             block_hash=block_hash,
             reuse_block=reuse_block,
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -3039,7 +3039,7 @@ class AsyncSubtensor(SubtensorMixin):
             data_type="TimelockEncrypted",
             data=data_,
             period=period,
-        ), reveal_round,
+        ), reveal_round
 
     async def subnet(
         self,
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -3032,17 +3032,14 @@ class AsyncSubtensor(SubtensorMixin):
         # increase reveal_round in return + 1 because we want to fetch data from the chain after that round was revealed
         # and stored.
         data_ = {"encrypted": encrypted, "reveal_round": reveal_round}
-        return (
-            await publish_metadata(
-                subtensor=self,
-                wallet=wallet,
-                netuid=netuid,
-                data_type="TimelockEncrypted",
-                data=data_,
-                period=period,
-            ),
-            reveal_round,
-        )
+        return await publish_metadata(
+            subtensor=self,
+            wallet=wallet,
+            netuid=netuid,
+            data_type="TimelockEncrypted",
+            data=data_,
+            period=period,
+        ), reveal_round,
 
     async def subnet(
         self,
@@ -3224,7 +3221,6 @@ class AsyncSubtensor(SubtensorMixin):
             return None
 
         current_block = await self.substrate.get_block()
-        assert current_block is not None, "Failed to retrieve current block"
         current_block_hash = current_block.get("header", {}).get("hash")
 
         if block is not None:
@@ -3331,8 +3327,6 @@ class AsyncSubtensor(SubtensorMixin):
             block_hash=block_hash,
             reuse_block=reuse_block,
         )
-        assert unix is not None, "Failed to retrieve timestamp"
-        assert isinstance(unix.value, int), "Timestamp value is not an integer"
         return datetime.fromtimestamp(unix.value / 1000, tz=timezone.utc)
 
     async def get_subnet_owner_hotkey(
@@ -3353,9 +3347,6 @@ class AsyncSubtensor(SubtensorMixin):
         """
         hotkey = await self.query_subtensor(
             name="SubnetOwnerHotkey", params=[netuid], block=block
-        )
-        assert hotkey is None or isinstance(hotkey, str), (
-            f"Expected str or None, got {type(hotkey).__name__}"
         )
         return hotkey
 
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -3352,9 +3352,9 @@ class AsyncSubtensor(SubtensorMixin):
         hotkey = await self.query_subtensor(
             name="SubnetOwnerHotkey", params=[netuid], block=block
         )
-        assert hotkey is None or isinstance(
-            hotkey, str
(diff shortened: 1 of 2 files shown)


bittensor/core/async_subtensor.py
```diff
@@ -3341,9 +3349,12 @@ async def get_subnet_owner_hotkey(
         Returns:
             The hotkey of the subnet owner if available; None otherwise.
         """
-        return await self.query_subtensor(
+        hotkey = await self.query_subtensor(
             name="SubnetOwnerHotkey", params=[netuid], block=block
         )
+        assert hotkey is not None, "Failed to retrieve subnet owner hotkey"
+        assert isinstance(hotkey.value, str), "Subnet owner hotkey is not a string"
+        return hotkey.value
```
Arthurdw: Not 100% about the value part here as this was not in the original code, but it should be there for the value right?

Arthurdw: Okey the tests pointed out this concern.

Arthurdw: Resolved in https://github.com/opentensor/bittensor/pull/2918/commits/cb830169a9fe0360b65ae3445e52635b02971fe4

bittensor/core/async_subtensor.py
```diff
@@ -3032,19 +3032,22 @@ async def set_reveal_commitment(
...
+            await publish_metadata(
+                subtensor=self,
+                wallet=wallet,
+                netuid=netuid,
+                data_type="TimelockEncrypted",
+                data=data_,
+                period=period,
+            ),
+            reveal_round,
+        )
```
basfroman: I'm not actually against it, ruff accepts both options without objection. 
The option we have now is considered more Pythonic.
What is the essence of this change from your point of view?

PS: **Avoid extra parentheses...**

bittensor/core/async_subtensor.py
```diff
@@ -3316,16 +3324,16 @@ async def get_timestamp(
...
-        return datetime.fromtimestamp(unix / 1000, tz=timezone.utc)
+        unix = await self.query_module(
+            "Timestamp",
+            "Now",
+            block=block,
+            block_hash=block_hash,
+            reuse_block=reuse_block,
+        )
+        assert unix is not None, "Failed to retrieve timestamp"
+        assert isinstance(unix.value, int), "Timestamp value is not an integer"
```
basfroman: We do not use assertions in the code, because SDK is not a responsible tool. The user himself catches errors and decides what to do.

bittensor/core/async_subtensor.py
```diff
@@ -3218,7 +3224,9 @@ async def handler(block_data: dict):
             return None
 
         current_block = await self.substrate.get_block()
+        assert current_block is not None, "Failed to retrieve current block"
```
basfroman: We do not use assertions in the code, because SDK is not a responsible tool. The user himself catches errors and decides what to do.

bittensor/core/async_subtensor.py
```diff
@@ -3343,9 +3351,13 @@ async def get_subnet_owner_hotkey(
         Returns:
             The hotkey of the subnet owner if available; None otherwise.
         """
-        return await self.query_subtensor(
+        hotkey = await self.query_subtensor(
             name="SubnetOwnerHotkey", params=[netuid], block=block
         )
+        assert hotkey is None or isinstance(hotkey, str), (
+            f"Expected str or None, got {type(hotkey).__name__}"
+        )
```
basfroman: We do not use assertions in the code, because SDK is not a responsible tool. The user himself catches errors and decides what to do.

bittensor/core/chain_data/info_base.py
```diff
@@ -3,6 +3,7 @@
 
 from bittensor.core.errors import SubstrateRequestException
 
+# NOTE: once pyhton 3.10+ is required, we can use `typing.Self` instead of this
```
basfroman: good point (but we still support python v3.9)

basfroman (changes requested):

bittensor/core/async_subtensor.py
Arthurdw: This was auto formatted by the `black` fmt; don't have any preference either way. :)

bittensor/core/chain_data/info_base.py
Arthurdw: Yes, but just added it as a future quick fix note. As it greatly improves the lsp compatibility.

bittensor/core/async_subtensor.py
Arthurdw: Not quite sure what to make from this, if we don't assert we are not guaranteed that the type is in fact what we expect and promise to return right?

Here these assertions are not user input exceptions, but these assertions are what should always be the case, if it is not the case then something gravely has gone wrong and our promised implementation is not doing what we are saying that we are going to do. (if you get my gist)

bittensor/core/async_subtensor.py
Arthurdw: However, an argument for this formatting could be that it more clearly indicates the tuple return because of the response being multi-line.

basfroman: Please see the requirements. we use `ruff` instead of `black`. 
Please also undo these changes.

bittensor/core/async_subtensor.py
basfroman: If something does not match or is changed in the chain, the error will be AssertionError instead of the real one.
Pls remove all `assert`s in your pr. The same above.

Arthurdw: Altough that I do believe that a assertion error with a more descriptive error message would be more useful than receiving can not use .get of None; I have removed all the assertions.

bittensor/core/async_subtensor.py
```diff
@@ -3360,9 +3362,10 @@ async def get_subnet_owner_hotkey(
         Returns:
             The hotkey of the subnet owner if available; None otherwise.
         """
-        return await self.query_subtensor(
+        hotkey = await self.query_subtensor(
             name="SubnetOwnerHotkey", params=[netuid], block=block
         )
+        return hotkey
```
```suggestion
        return await self.query_subtensor(
            name="SubnetOwnerHotkey", params=[netuid], block=block
        )
```

bittensor/core/async_subtensor.py
```diff
@@ -3333,16 +3337,14 @@ async def get_timestamp(
...
-            )
-        ).value
-        return datetime.fromtimestamp(unix / 1000, tz=timezone.utc)
+        unix = await self.query_module(
+            "Timestamp",
+            "Now",
+            block=block,
+            block_hash=block_hash,
+            reuse_block=reuse_block,
+        )
```
basfroman: if you removed `.value` from this part, then it is no longer `unix =`. It is now just a `response =`.
You should either put `.value` back or add an intermediate line like:
```py
unix = response.value
```
and remove `unix.value` below in a part `unix.value / 1000`

basfroman (changes requested):

thewhaleking (approved):

basfroman (approved):

PR #2937, opened 2025-06-26
Improve subnet method logic

basfroman:
- fixed the case if decoded is None


bittensor/core/async_subtensor.py
```diff
@@ -3091,6 +3091,7 @@ async def subnet(
 
         if isinstance(decoded := query.decode(), dict):
             return DynamicInfo.from_dict(decoded)
+        return None
```
thewhaleking, line 3094: Does this actually change the logic, or just makes it explicit?

thewhaleking (approved):

bittensor/core/async_subtensor.py
basfroman: Just explicit :)

PR #2952, opened 2025-07-08
New logic to get price for `DynamicInfo`

basfroman:
Subtensors have new methods:
- `get_subnet_price`
- `get_subnet_prices`

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -2350,8 +2350,9 @@ class AsyncSubtensor(SubtensorMixin):
             current_price = current_sqrt_price * current_sqrt_price
             current_price_in_tao = Balance.from_rao(int(current_price * 1e9))
             prices.update({id_: current_price_in_tao})
+
         # SN0 price is always 1 TAO
-        prices[0] = Balance.from_tao(1)
+        prices.update({0: Balance.from_tao(1)})
         return prices
 
     async def get_unstake_fee(
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -651,8 +651,13 @@ class AsyncSubtensor(SubtensorMixin):
             "get_all_dynamic_info",
             block_hash=block_hash,
         )
-        subnets = DynamicInfo.list_from_dicts(query.decode())
-        return subnets
+
+        subnet_prices = await self.get_subnet_prices()
+        decoded = query.decode()
+
+        for sn in decoded:
+            sn.update({"price": subnet_prices.get(sn["netuid"], Balance.from_tao(0))})
+        return DynamicInfo.list_from_dicts(decoded)
 
     async def blocks_since_last_step(
         self,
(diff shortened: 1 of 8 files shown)


bittensor/core/async_subtensor.py
```diff
@@ -651,8 +651,13 @@ async def all_subnets(
             "get_all_dynamic_info",
             block_hash=block_hash,
         )
-        subnets = DynamicInfo.list_from_dicts(query.decode())
-        return subnets
+
+        subnet_prices = await self.get_subnet_prices()
+        decoded = query.decode()
```
thewhaleking, line 659: Does this need to be decoded?

bittensor/core/async_subtensor.py
```diff
@@ -651,8 +651,13 @@ async def all_subnets(
             "get_all_dynamic_info",
             block_hash=block_hash,
         )
-        subnets = DynamicInfo.list_from_dicts(query.decode())
-        return subnets
+
+        subnet_prices = await self.get_subnet_prices()
```
thewhaleking: Would recommend doing this in a gather

bittensor/core/async_subtensor.py
```diff
@@ -902,8 +907,13 @@ async def get_all_subnets_info(
         )
         if not result:
             return []
-        else:
-            return SubnetInfo.list_from_dicts(result)
+
+        subnets_prices = await self.get_subnet_prices()
```
thewhaleking, line 914: Same questions as above

bittensor/core/chain_data/dynamic_info.py
```diff
@@ -113,7 +106,7 @@ def _from_dict(cls, decoded: dict) -> "DynamicInfo":
             tao_in=tao_in,
             k=tao_in.rao * alpha_in.rao,
             is_dynamic=is_dynamic,
-            price=price,
+            price=decoded.get("price", None),
```
thewhaleking: Price should be a Balance type

bittensor/core/async_subtensor.py
basfroman: same answer

bittensor/core/async_subtensor.py
basfroman: no

bittensor/core/async_subtensor.py
basfroman: good catch

bittensor/core/chain_data/dynamic_info.py
basfroman: The balance is Optional. I don't want to create another convention that if the price = 0 then it is not transferred or something else. There is no price, so it is not passed. 
The user may want to use the subnet data without a price in his/her script independently.

basfroman: But we raise the error in case price isn't None or Balance

ibraheem-abe (dismissed):

ibraheem-abe (approved):

PR #2957, opened 2025-07-09
missed await

thewhaleking:


thewhaleking:
@basfroman updating this in another PR

PR #2978, opened 2025-07-18
Fix crv3 tests after devnet-ready get `CRV3WeightCommitsV2`

basfroman:
broken test in subtensor pr https://github.com/opentensor/subtensor/actions/runs/16350309250/job/46196105509?pr=1862

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -109,6 +109,7 @@ from bittensor.utils.balance import (
     fixed_to_float,
     check_and_convert_to_balance,
 )
+from bittensor.utils import deprecated_message
 from bittensor.utils.btlogging import logging
 from bittensor.utils.liquidity import (
     calculate_fees,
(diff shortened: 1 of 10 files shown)


thewhaleking (commented):
This should probably be a new method. Otherwise this looks like a breaking change to me.

basfroman:
> This should probably be a new method. Otherwise this looks like a breaking change to me.

good point just about old blocks

bittensor/core/async_subtensor.py
```diff
@@ -1678,7 +1678,7 @@ async def get_current_weight_commit_info(
         block: Optional[int] = None,
         block_hash: Optional[str] = None,
         reuse_block: bool = False,
-    ) -> list:
+    ) -> list[tuple[str, str, int]]:
         """
```
thewhaleking, line 1683: Make sure to mark this with a deprecated flag

thewhaleking (dismissed):
Good, but make sure you deprecate the `get_current_weight_commit_info` method in subtensor and async_subtensor. 

Also, we probably shouldn't merge this yet, as we should merge https://github.com/opentensor/bittensor/pull/2979 and release the patch today.

thewhaleking (approved):

PR #2993, opened 2025-07-24
Transfers improvements

thewhaleking:
Mirrors https://github.com/opentensor/btcli/pull/562

Uses different extrinsics depending on what is selected.

Transfer All uses "Balances.transfer_all", with param `keep_alive` toggled depending on `keep_alive` arg.

Non-Transfer-All with `keep_alive=False` uses "Balances.transfer_allow_death"

Non-Transfer-All with `keep_alive=True` uses "Balances.transfer_keep_alive"

### DIFF
--- a/bittensor/core/extrinsics/asyncex/transfer.py
+++ b/bittensor/core/extrinsics/asyncex/transfer.py
@@ -6,6 +6,7 @@ from bittensor.utils import (
     get_explorer_url_for_network,
     is_valid_bittensor_address_or_public_key,
     unlock_key,
+    get_transfer_fn_params,
 )
 from bittensor.utils.balance import Balance
 from bittensor.utils.btlogging import logging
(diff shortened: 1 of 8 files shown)


bittensor/core/extrinsics/asyncex/transfer.py
```diff
@@ -19,7 +19,8 @@ async def _do_transfer(
     subtensor: "AsyncSubtensor",
     wallet: "Wallet",
     destination: str,
-    amount: "Balance",
+    amount: Optional[Balance],
+    keep_alive: bool = True,
```
basfroman: Move `keep_alive: bool = True` at the end of the argument list. 
This could potentially break user scripts.
Let's make it nice in version 10. I wrote it down.

basfroman (changes requested):
Left a comment.
Also, it would be good to add an e2e test using the new flag in `tests/e2e_tests/test_transfer.py`

bittensor/core/extrinsics/asyncex/transfer.py
thewhaleking: `_do_transfer` is a private function though.  But I can move it.

basfroman: > `_do_transfer` is a private function though. But I can move it.

Agree. But we know that SNs still uses `_do_*` methods and extrinsics in their scripts.

basfroman (dismissed):

basfroman (approved):

PR #3008, opened 2025-08-08
`LoggingMachine` initialization updated to explicitly call both parent constructors

### LINKED ISSUE #3003: Cannot add handler to btlogging
### Describe the bug

If I use `logging.addHandler`, it breaks:
```py
    logging.addHandler(log.FileHandler("/tmp/test.log"))
/opt/homebrew/Cellar/python@3.12/3.12.10_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py:1708: in addHandler
    if not (hdlr in self.handlers):
E   AttributeError: 'LoggingMachine' object has no attribute 'handlers'. Did you mean: '_handlers'?
```

### To Reproduce

```py
from bittensor import logging
import logging as log

logging.addHandler(log.SteamHandler()`
```

### Expected behavior

It does not break

### Screenshots

_No response_

### Environment

macOS

### Additional context

_No response_

basfroman:
Issue: https://github.com/opentensor/bittensor/issues/3003

### Note:
2 hours ago [`uv=0.8.7` released](https://pypi.org/project/uv/#history). This version doesn't support Python 3.9. We have to pin 0.8.6 while we support Python 3.9

### DIFF
--- a/.github/workflows/flake8-and-mypy.yml
+++ b/.github/workflows/flake8-and-mypy.yml
@@ -42,7 +42,8 @@ jobs:
           python -m venv venv
           source venv/bin/activate
           python -m pip install --upgrade pip
-          python -m pip install uv
+          # uv==0.8.6 is the last version that supports python 3.9
+          python -m pip install uv==0.8.6
           python -m uv sync --extra dev --active
 
       - name: Flake8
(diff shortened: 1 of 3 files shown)


thewhaleking (approved):
I'm surprised the pip resolver breaks for uv greater than this

PR #3021, opened 2025-08-20
[e2e] [testing] use close_in_value for a balance check

camfairchild:
Adds a `CLOSE_IN_VALUE` to a test to make the balance check less strict (we are adding tx fees to `register_network`).

Also removes unneeded balance comparisons (seems to test just the wrapping in `Balance`)


basfroman:
implemented

PR #3043, opened 2025-09-03
Async Get_balances at a specific block returns current block

### LINKED ISSUE #3038: Async Get_balances at a specific block returns current block
### Describe the bug

subtensor.get_balance(mycoldkey, block=block) returns a different value to subtensor.get_balances(*(mycoldkey), block=block)

### To Reproduce

run this snippet the values should be the same 
```
import asyncio

# block and coldkey we are interested in

mycoldkey = '5F6a4sLjgH6B3wfioD2s2xvv64UcFUcrK3UVAHzfNLYNiT4R'
mysubtensor = 'wss://archive.minersunion.ai'
async def main():
    async with AsyncSubtensor(mysubtensor) as subtensor:
        myaddresses=[mycoldkey]
        block = 6309620 
        balance_data_block = await subtensor.get_balance(mycoldkey, block=block)
        print(f'balance_data_block: {balance_data_block}')
        balanceS_data_block = await subtensor.get_balances(*(myaddresses), block=block)
        print(f'balanceS_data_block: {balanceS_data_block[mycoldkey]}')

if __name__ == "__main__":
    asyncio.run(main())
```

### Expected behavior

Both calls should return the same value

### Screenshots

_No response_

### Environment

bittensor 9.10.0

### Additional context

_No response_

basfroman:
Fix for https://github.com/opentensor/bittensor/issues/3038

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -1235,8 +1235,6 @@ class AsyncSubtensor(SubtensorMixin):
         """
         if reuse_block:
             block_hash = self.substrate.last_block_hash
-        elif not block_hash:
-            block_hash = await self.get_block_hash(block)
         else:
             block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
         calls = [
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -1236,7 +1236,7 @@ class AsyncSubtensor(SubtensorMixin):
         if reuse_block:
             block_hash = self.substrate.last_block_hash
         elif not block_hash:
-            block_hash = await self.get_block_hash()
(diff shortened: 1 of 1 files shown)


ibraheem-abe (approved):

bittensor/core/async_subtensor.py
```diff
@@ -1235,8 +1235,6 @@ async def get_balances(
         """
         if reuse_block:
             block_hash = self.substrate.last_block_hash
-        elif not block_hash:
-            block_hash = await self.get_block_hash()
```
camfairchild, line 1239: How about
```
elif block_hash is None and block is None:
```
That way they can specify blockhash without an extra network call.

Also using `is None` instead of checking for a false-y value because `block` can be `0`

camfairchild (changes requested):

PR #3044, opened 2025-09-03
Fix bug if `block==0`

basfroman:
This is not super critical and only works with archived data, but theoretically if someone passes block=0 then the current block is used instead of the real 0.

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -343,7 +343,7 @@ class AsyncSubtensor(SubtensorMixin):
         # Return the appropriate value.
         if block_hash:
             return block_hash
-        if block:
+        if block is not None:
             return await self.get_block_hash(block)
         return None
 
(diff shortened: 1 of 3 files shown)


camfairchild (approved):

bittensor/core/async_subtensor.py
```diff
@@ -1235,6 +1235,9 @@ async def get_balances(
         """
         if reuse_block:
             block_hash = self.substrate.last_block_hash
+        elif block_hash is None and block is None:
+            # Neither block nor block_hash provided, default to head
+            block_hash = await self.get_block_hash()
         else:
```
camfairchild, line 1241:
```suggestion
         elif block_hash is None:
```
So caller can specify blockhash

camfairchild (changes requested):

bittensor/core/async_subtensor.py
basfroman: > So caller can specify blockhash

this logic is covered in `determine_block_hash` method

```py
        # Return the appropriate value.
        if block_hash:
            return block_hash
```

camfairchild (approved):

PR #3053, opened 2025-09-09
[v10] Unify extrinsic return values by introducing an `ExtrinsicResponse` class

basfroman:
A new class `ExtrinsicResponse` has been implemented that completely inherits the behavior of extrinsic responses in previous SDK versions, but at the same time expands the capabilities, contains new data that SDK users will use for more detailed verification or checking of responses.

```py
@dataclass
class ExtrinsicResponse:
    """
    A standardized response container for handling the extrinsic results submissions and related operations in the SDK.

    This class is designed to give developers a consistent way to represent the outcome of an extrinsic call — whether
    it succeeded or failed — along with useful metadata for debugging, logging, or higher-level business logic.

    The object also implements tuple-like behavior:
      * Iteration yields ``(success, message)``.
      * Indexing is supported: ``response[0] -> success``, ``response[1] -> message``.
      * ``len(response)`` returns 2.

    Attributes:
        success: Indicates if the extrinsic execution was successful.
        message: A status or informational message returned from the execution (e.g., "Successfully registered subnet").
        error: Captures the underlying exception if the extrinsic failed, otherwise `None`.
        data: Arbitrary data returned from the extrinsic, such as decoded events, or extra context.
        extrinsic_function: The name of the SDK extrinsic function that was executed (e.g. "register_subnet_extrinsic").
        extrinsic: The raw extrinsic object used in the call, if available.

    Example:
        import bittensor as bt

        subtensor = bt.SubtensorApi("local")
        wallet = bt.Wallet("alice")

        response = subtensor.subnets.register_subnet(alice_wallet)
        print(response)

        ExtrinsicResponse:
            success: True
            message: Successfully registered subnet
            error: None
            extrinsic_function: register_subnet_extrinsic
            extrinsic: {'account_id': '0xd43593c715fdd31c...

        success, message = response
        print(success, message)

        True Successfully registered subnet

        print(response[0])
        True
        print(response[1])
        'Successfully registered subnet'
    """

    success: bool = True
    message: str = None
    error: Optional[Exception] = None
    data: Optional[Any] = None
    extrinsic_function: Optional[str] = None
    extrinsic: Optional[GenericExtrinsic] = None

    def __iter__(self):
        yield self.success
        yield self.message

    def __str__(self):
        return str(
            f"ExtrinsicResponse:"
            f"\n\tsuccess: {self.success}"
            f"\n\tmessage: {self.message}"
            f"\n\terror: {self.error}"
            f"\n\textrinsic_function: {self.extrinsic_function}"
            f"\n\textrinsic: {self.extrinsic}"
        )

    def __repr__(self):
        return repr((self.success, self.message))

    def __eq__(self, other: Any) -> bool:
        if isinstance(other, tuple):
            return (self.success, self.message) == other
        if isinstance(other, ExtrinsicResponse):
            return (
                self.success == other.success
                and self.message == other.message
                and self.error == other.error
                and self.extrinsic_function == other.extrinsic_function
                and self.extrinsic == other.extrinsic
            )
        return super().__eq__(other)

    def __getitem__(self, index: int) -> Any:
        if index == 0:
            return self.success
        elif index == 1:
            return self.message
        else:
            raise IndexError(
                "ExtrinsicResponse only supports indices 0 (success) and 1 (message)."
            )

    def __len__(self):
        return 2

```

### DIFF
--- a/bittensor/core/extrinsics/asyncex/registration.py
+++ b/bittensor/core/extrinsics/asyncex/registration.py
@@ -287,7 +287,7 @@ async def register_extrinsic(
     if not torch:
         log_no_torch_error()
         return ExtrinsicResponse(
-            False, "No torch installed.", extrinsic_function=get_function_name()
+            False, "Torch is not installed.", extrinsic_function=get_function_name()
         )
 
     # Attempt rolling registration.
--- a/bittensor/core/extrinsics/asyncex/registration.py
+++ b/bittensor/core/extrinsics/asyncex/registration.py
@@ -10,7 +10,8 @@ import asyncio
 from typing import Optional, Union, TYPE_CHECKING
 
 from bittensor.core.extrinsics.asyncex.utils import get_extrinsic_fee
-from bittensor.utils import unlock_key
+from bittensor.core.types import ExtrinsicResponse
+from bittensor.utils import unlock_key, get_function_name
 from bittensor.utils.btlogging import logging
 from bittensor.utils.registration import log_no_torch_error, create_pow_async, torch
 
@@ -27,7 +28,7 @@ async def burned_register_extrinsic(
     raise_error: bool = False,
     wait_for_inclusion: bool = True,
     wait_for_finalization: bool = True,
-) -> bool:
+) -> ExtrinsicResponse:
     """Registers the wallet to chain by recycling TAO.
 
     Parameters:
@@ -42,18 +43,24 @@ async def burned_register_extrinsic(
         wait_for_finalization: Whether to wait for the finalization of the transaction.
 
     Returns:
-        success: True if the extrinsic was successful. Otherwise, False.
+        ExtrinsicResponse: The result object of the extrinsic execution.
     """
     block_hash = await subtensor.substrate.get_chain_head()
     if not await subtensor.subnet_exists(netuid, block_hash=block_hash):
         logging.error(
             f":cross_mark: [red]Failed error:[/red] subnet [blue]{netuid}[/blue] does not exist."
         )
-        return False
+        return ExtrinsicResponse(
+            False,
+            f"Subnet #{netuid} does not exist",
+            extrinsic_function=get_function_name(),
+        )
 
     if not (unlock := unlock_key(wallet)).success:
         logging.error(unlock.message)
-        return False
+        return ExtrinsicResponse(
+            False, unlock.message, extrinsic_function=get_function_name()
+        )
 
     logging.info(
         f":satellite: [magenta]Checking Account on subnet[/magenta] [blue]{netuid}[/blue][magenta] ...[/magenta]"
@@ -71,12 +78,15 @@ async def burned_register_extrinsic(
     )
 
     if not neuron.is_null:
-        logging.info(":white_heavy_check_mark: [green]Already Registered[/green]")
+        message = "Already registered."
+        logging.info(f":white_heavy_check_mark: [green]{message}[/green]")
         logging.info(f"\t\tuid: [blue]{neuron.uid}[/blue]")
         logging.info(f"\t\tnetuid: [blue]{neuron.netuid}[/blue]")
         logging.info(f"\t\thotkey: [blue]{neuron.hotkey}[/blue]")
         logging.info(f"\t\tcoldkey: [blue]{neuron.coldkey}[/blue]")
-        return True
+        return ExtrinsicResponse(
+            message=message, extrinsic_function=get_function_name()
+        )
 
     logging.debug(":satellite: [magenta]Recycling TAO for Registration...[/magenta]")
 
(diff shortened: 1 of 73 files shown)


bittensor/core/extrinsics/asyncex/registration.py
```diff
@@ -271,7 +286,9 @@ async def register_extrinsic(
 
     if not torch:
         log_no_torch_error()
-        return False
+        return ExtrinsicResponse(
+            False, "No torch installed.", extrinsic_function=get_function_name()
```
ibraheem-abe: Torch is not installed

bittensor/core/extrinsics/asyncex/registration.py
```diff
@@ -125,12 +136,16 @@ async def burned_register_extrinsic(
...
+        return response
 
     # neuron not found, try again
-    logging.error(":cross_mark: [red]Unknown error. Neuron not found.[/red]")
-    return False
+    message = "Unknown error. Neuron not found."
+    logging.error(f":cross_mark: [red]{message}[/red]")
+    response.success = False
+    response.message = message
+    return response
```
ibraheem-abe, line 148: Do u think we should also check the uid of the registered hotkey and display it in response.data['uid']?
Like we do in root

bittensor/core/extrinsics/asyncex/staking.py
```diff
@@ -178,15 +184,15 @@ async def add_stake_extrinsic(
...
-    if safe_staking and "Custom error: 8" in message:
+    if safe_staking and "Custom error: 8" in response.message:
         logging.error(
             ":cross_mark: [red]Failed[/red]: Price exceeded tolerance limit. Either increase price tolerance or enable partial staking."
         )
     else:
-        logging.error(f":cross_mark: [red]Failed: {message}.[/red]")
-    return False
+        logging.error(f":cross_mark: [red]Failed: {response.message}.[/red]")
+    return response
```
ibraheem-abe, line 195: Thoughts on adding stake and balance in extrinsic.response?

bittensor/core/extrinsics/asyncex/transfer.py
```diff
@@ -138,7 +145,7 @@ async def transfer_extrinsic(
         logging.info(
             f"Balance: [blue]{account_balance}[/blue] :arrow_right: [green]{new_balance}[/green]"
         )
-        return True
+        return response
```
ibraheem-abe, line 148: What do u think of adding the new account balance in the extrinsic response?

bittensor/core/extrinsics/asyncex/unstaking.py
```diff
@@ -72,118 +74,110 @@ async def unstake_extrinsic(
...
+                block_hash=new_block_hash,
+            ),
+        )
+        logging.info(
+            f"Balance: [blue]{old_balance}[/blue] :arrow_right: [green]{new_balance}[/green]"
+        )
+        logging.info(
+            f"Stake: [blue]{old_stake}[/blue] :arrow_right: [green]{new_stake}[/green]"
+        )
+        return response
```
ibraheem-abe, line 172: Thoughts on adding new stake and new balance in the response?

migration.md
```diff
@@ -42,11 +42,12 @@
     ```
     </details>
   
-2. Unify extrinsic return values by introducing an ExtrinsicResponse class. Extrinsics currently return either a boolean or a tuple. 
+2. ✅ Unify extrinsic return values by introducing an ExtrinsicResponse class. Extrinsics currently return either a boolean or a tuple.
```
ibraheem-abe, line 45: Since this is a big one, do u think an example will suffice? 
How it extrinsics previously worked vs what they return now etc.

ibraheem-abe (commented):
Overall looks good. 

One qq: Few important calls (like committing weights etc) had exception handling which we removed. 

Do u think end behaviour will stay the same?

ibraheem-abe (approved):

migration.md
basfroman: The thing is that all the behavior that was before is coverad in the current implementation. But yes, there is a sufficient amount of new data/fields. I thought to reflect this in the documentation that will be created based on migration.md. The point is that the data that is provided in ExtrinsicResponse will most likely be expanded from this PR. For example: I am going to add a new field `.extrinsic_fee` that will be calculated in case of successful execution of the extrinsic and added to the ExtrinsicResponse instance. Do you think it is worth adding this right now?

bittensor/core/extrinsics/asyncex/registration.py
basfroman: I would really not like to burden the logic with additional calculations. Ultimately, the user gets a message that the neuron will not be found. This means that the neuron is not registered. Also, if the user wants to determine whether the uid of the registered hotkey exists, he will determine this using the corresponding subtensor methods. Does it make sense?

bittensor/core/extrinsics/asyncex/staking.py
basfroman: I'll do it in another PR for all extrinsics

migration.md
basfroman: well, I'll do it in another PR for all extrinsics

bittensor/core/extrinsics/asyncex/unstaking.py
basfroman: I'll do it in another PR for all extrinsics

bittensor/core/extrinsics/asyncex/transfer.py
basfroman: I'll do it in another PR for all extrinsics

bittensor/core/extrinsics/asyncex/registration.py
ibraheem-abe: Yes agreed. 

But since we will be passing data in the extrinsicresponse (in root registration), why not do it here too for normal subnets :) 
Will be super useful for the user to just do response.data["uid"] if it was successful in their scripts etc. 

But open to skip this if its too much

basfroman:
> Overall looks good.
> 
> One qq: Few important calls (like committing weights etc) had exception handling which we removed.
> 
> Do u think end behaviour will stay the same?

Yes. The same. But I want to add wrapper `try... except...` for all extrinsics and raise an error if`raise_error=True` is passed and extrinsic or some logic inside (e.g. unlock wallet) raised the error. 
This is in my TODO for one of the next PRs.

bittensor/core/extrinsics/asyncex/registration.py
basfroman: Done

bittensor/core/extrinsics/asyncex/registration.py
basfroman: Let me come back to this issue in the next PR. I made a note. Again, we should try to get rid of additional calculations. But the idea is great

PR #3061, opened 2025-09-19
deps: update munch to prevent deprecation warning

Arthurdw:
### Bug

Just resolves the deprecation issue:
```
.venv/lib/python3.13/site-packages/munch/__init__.py:24
  /home/arthur/Documents/tao/oms/.venv/lib/python3.13/site-packages/munch/__init__.py:24: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_r
esources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
    import pkg_resources
```

### Description of the Change

Bumps the munch dependency version, allowing projects that use this as a dependency to not have to worry.

### Alternate Designs

N/A

### Possible Drawbacks

As far as I've seen there are no incompatibilities.

### Release Notes

- deps: update munch to latest version to resolve deprecation warnings

### Branch Acknowledgement
- [x] I am acknowledging that I am opening this branch against `staging`


uv.lock
basfroman, line 1: pls remove this file from pr

pyproject.toml
```diff
@@ -95,4 +94,4 @@ classifiers = [
 
 [tool.setuptools]
 package-dir = {"bittensor" = "bittensor"}
-script-files = ["bittensor/utils/certifi.sh"]
\ No newline at end of file
+script-files = ["bittensor/utils/certifi.sh"]
```
basfroman, line 97: Could you clarify the reason for adding this line?

We generally avoid exposing internal helper scripts (like certifi.sh) in the global bin/ directory, unless there’s a strong need to make them directly executable.

Also, this script can already be executed using:
```bash
python -m bittensor certifi
```
which keeps the API cleaner and avoids polluting the user’s path with project internals.

basfroman (changes requested):

pyproject.toml
Arthurdw: Was also in the original file, my file save just removed a invisible char.
image

uv.lock
Arthurdw: my bad, removed

pyproject.toml
basfroman: My bad. Good!

basfroman:
Hey @Arthurdw, pls merge staging into your PR and I'll run tests.

basfroman (approved):

PR #3079, opened 2025-10-04
[v10] Development framework

basfroman:
### Summary

This PR introduces a new Development framework for Bittensor SDK.

The framework provides a clean orchestration layer for testing and automating subnet lifecycle operations.  
It abstracts low-level Subtensor extrinsics into clear, reproducible steps and now serves as the foundation for all existing E2E tests.

As a result, tests are easier to write, understand, and maintain with clearer structure, better logs, and more consistent behavior during execution.

---

### Motivation

Until now, E2E tests relied on direct Subtensor API calls, which led to repeated code, inconsistent patterns, and limited debugging insight.  
This framework consolidates those patterns into a single, reusable system built around the `TestSubnet` class.

`TestSubnet` standardizes subnet operations  such as registration, activation, neuron addition, and hyperparameter updates under one API.  
It introduces synchronous and asynchronous execution modes, unified extrinsic responses, and automatic `NETUID` handling across test runs.

The result is a much more reliable, maintainable, and developer-friendly way to test the Bittensor protocol, both for SDK devs and the broader community.

---

### Key Advantages

- Unified, declarative syntax for subnet orchestration.
- Supports both **single-step** and **multi-step** execution flows.
- Works in synchronous (`execute_steps`) and asynchronous (`async_execute_steps`) modes.
- Automatically injects the correct `NETUID` for all operations within the same test instance.
- Uses the standardized `ExtrinsicResponse` type for richer debugging including status, message, fees, receipts, and etc.
- Produces structured, colorized logs that make on-chain activity and errors easy to trace.
- Reduces code duplication and improves readability across all test suites.
- **Community-ready:** can also be used by subnet operators and researchers to deploy, configure, and test subnet behavior in real or simulated environments.

---

### Result

All E2E test suites are now built on this framework.  
The codebase is simpler, easier to maintain, and much more transparent during debugging and analysis.

For full technical documentation, see [`bittensor/extras/dev_framework/README.md`](https://github.com/opentensor/bittensor/blob/feat/roman/e2e-tests-framework/bittensor/extras/dev_framework/README.md).


bittensor/core/extrinsics/asyncex/utils.py
```diff
@@ -67,6 +68,7 @@ async def sudo_call_extrinsic(
         raise_error: Raises a relevant exception rather than returning `False` if unsuccessful.
         wait_for_inclusion: Whether to wait for the inclusion of the transaction.
         wait_for_finalization: Whether to wait for the finalization of the transaction.
+        root: Whether to use the root account for not sudo call in some subtensor pallet.
```
thewhaleking: This is very confusing.

bittensor/core/extrinsics/asyncex/utils.py
```diff
@@ -79,19 +81,20 @@ async def sudo_call_extrinsic(
...
+            call_module=call_module,
+            call_function=call_function,
+            call_params=call_params,
         )
+        if not root:
+            call = await subtensor.substrate.compose_call(
+                call_module="Sudo",
+                call_function="sudo",
+                call_params={"call": call},
+            )
```
thewhaleking: Shouldn't this get applied is root is True?

bittensor/core/extrinsics/utils.py
```diff
@@ -84,6 +84,7 @@ def sudo_call_extrinsic(
     raise_error: bool = False,
     wait_for_inclusion: bool = True,
     wait_for_finalization: bool = True,
+    root: Optional[bool] = None,
```
thewhaleking: Why is this an optional bool rather than a bool?

bittensor/core/extrinsics/utils.py
```diff
@@ -102,6 +103,7 @@ def sudo_call_extrinsic(
         raise_error: Raises a relevant exception rather than returning `False` if unsuccessful.
         wait_for_inclusion: Whether to wait for the inclusion of the transaction.
         wait_for_finalization: Whether to wait for the finalization of the transaction.
+        root: Whether to use the root account for not sudo call in some subtensor pallet.
```
thewhaleking: Confusing wording.

bittensor/core/extrinsics/utils.py
```diff
@@ -114,19 +116,20 @@ def sudo_call_extrinsic(
...
+            call_function=call_function,
+            call_params=call_params,
         )
+        if not root:
+            call = subtensor.substrate.compose_call(
+                call_module="Sudo",
+                call_function="sudo",
+                call_params={"call": call},
+            )
+
```
thewhaleking: Shouldn't this be applied if root is True?

bittensor/extras/dev_framework/calls/non_sudo_calls.py
thewhaleking, line 1: Should probably be git ignored.

bittensor/extras/dev_framework/calls/pallets.py
thewhaleking, line 1: Should probably be gitignored

bittensor/extras/dev_framework/calls/sudo_calls.py
thewhaleking, line 1: Should probably be gitignored

bittensor/extras/dev_framework/subnet.py
```diff
@@ -0,0 +1,537 @@
...
+
+    def execute_steps(self, steps: list[Union[STEPS, tuple]]):
+        """Executes a multiple steps synchronously."""
+        for step in steps:
+            self.execute_one(step)
+
+    async def async_execute_steps(self, steps: list[Union[STEPS, tuple]]):
+        """Executes a multiple steps asynchronously."""
+        for step in steps:
+            await self.async_execute_one(step)
```
thewhaleking, line 68: Why don't we just make it so that `execute_steps` and `async_execute_steps` take a `*steps` arg, so we can just pass in however many we want.

bittensor/extras/dev_framework/subnet.py
```diff
@@ -0,0 +1,537 @@
...
+            )
+            response = self.set_hyperparameter(
+                sudo_or_owner_wallet=sudo_or_owner_wallet,
+                call_function=call_function,
+                call_module=call_module,
+                call_params=call_params,
+                sudo_call=sudo_call,
+            )
+        if not response:
+            raise NotImplementedError(
```
thewhaleking, line 130: Not sure this is the best exception

bittensor/extras/dev_framework/calls/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # rewrite pallets.py
+    with open(pallets_file_path, "w") as f:
+        f.write(pallets_text)
+
+    print(f"Module {pallets_file_path} has been recreated successfully.")
+
+
+if __name__ == "__main__":
+    recreate_calls_subpackage()
```
thewhaleking, line 119: Would be great if we could run like `python -m bittensor generate-calls` or something to run this too.

tests/e2e_tests/conftest.py
```diff
@@ -325,3 +328,23 @@ def log_test_start_and_end(request):
     logging.console.info(f"🏁[green]Testing[/green] [yellow]{test_name}[/yellow]")
     yield
     logging.console.success(f"✅ [green]Finished[/green] [yellow]{test_name}[/yellow]")
+
+
+@pytest_asyncio.fixture(scope="session")
+def event_loop():
+    """Create an instance of the default event loop for each test case and close all alive tasks at the end."""
+    loop = asyncio.get_event_loop()
```
thewhaleking, line 336: Do we want to `get_event_loop` or `new_event_loop` here? `get_event_loop` can be a bit confusing because it will get a running loop if it's running, or create it if it's not.

thewhaleking (commented):
I really like this, but I think the structure could use some work. Namely having the generated code be gitignored. If you're a user generating methods, you are going to run into merge conflicts when you just want to `git pull`.

tests/e2e_tests/conftest.py
basfroman: we don't. this fixture solve the issue with closing alive tasks after test done. works well in this impl. also this used for e2e tests only.

bittensor/extras/dev_framework/calls/__init__.py
basfroman: It is better not to take this logic outside the subpackage

bittensor/extras/dev_framework/subnet.py
basfroman: We're currently passing as many steps as we need. But I want to explicitly pass a list. This provides a clear separation between execute_steps and execute_one.

bittensor/extras/dev_framework/calls/sudo_calls.py
basfroman: I understand why you suggested this. But since this is an explicit subpackage of the extension, even if for some reason the lists become outdated, we will update them. Otherwise, we and our users would have to constantly update them. I suggest keeping it this way. Three files don't carry much load, but they allow us to use the out-of-the-box logic.

bittensor/core/extrinsics/utils.py
basfroman: nope.

bittensor/core/extrinsics/asyncex/utils.py
basfroman: nope

bittensor/core/extrinsics/utils.py
basfroman: good catch. forgot to update. done

bittensor/core/extrinsics/utils.py
basfroman: fixed

bittensor/core/extrinsics/asyncex/utils.py
basfroman: updated

bittensor/core/extrinsics/asyncex/utils.py
basfroman: fixed

bittensor/extras/dev_framework/subnet.py
basfroman: This step hasn't been implemented. What could be simpler? We can expand this in the future if we want. Right now, it's fine.

thewhaleking (approved):

PR #3080, opened 2025-10-05
[v10] `serving` extrinsic module refactoring

basfroman:
- `bittensor.core.extrinsic.serving.get_metadata` functions moved to `subtensor.get_metadata` method
- `bittensor.core.extrinsic.serving.get_last_bonds_reset` function moved to `subtensor.get_last_bonds_reset` method


bittensor/core/async_subtensor.py
```diff
@@ -1977,6 +2006,26 @@ async def get_minimum_required_stake(self):
 
         return Balance.from_rao(getattr(result, "value", 0))
 
+    async def get_metadata(
```
thewhaleking: Very confusing name. Maybe `commitment_metadata` or something else. But `metadata` is very confusing. Most people are going to read this as fetching the Metadata (V14/V15) from the chain. I know this is a legacy method name (and I think I even named it lol), but it should be renamed for V10.

basfroman: Very good point. The later at night I send a PR, the more reviews it requires 😂

thewhaleking (approved):

ibraheem-abe (approved):

bittensor/core/async_subtensor.py
```diff
@@ -1274,6 +1272,37 @@ async def get_balances(
...
         return results
 
+    async def get_commitment_metadata(
+        self,
+        netuid: int,
+        hotkey_ss58: str,
+        block: Optional[int] = None,
+        block_hash: Optional[str] = None,
+        reuse_block: bool = False,
+    ) -> Union[str, dict]:
```
thewhaleking, line 1287: We can use `str | dict` now! But not necessary.

bittensor/core/async_subtensor.py
```diff
@@ -1274,6 +1272,37 @@ async def get_balances(
...
+            netuid: The unique subnet identifier.
+            hotkey_ss58: The hotkey ss58 address.
+            block: The blockchain block number for the query.
+            block_hash: The hash of the block at which to check the hotkey ownership.
+            reuse_block: Whether to reuse the last-used blockchain hash.
+
+        Returns:
+            The raw commitment metadata from specific subnet for given hotkey.
+        """
+        async with self.substrate:
```
thewhaleking: There's a slight problem with doing `async with self.substrate` in individual calls, as when the context closes, `AsyncSubstrateInterface.__aexit__` gets called. When this gets called, it begins the shutdown timer to close the connection. I don't think we want to do that.

(Yes, I am aware that I was the one who added this logic in the original `get_metadata` function)

thewhaleking (changes requested):

bittensor/core/async_subtensor.py
basfroman: I don't want to replace this across the entire codebase right now. And changing it in one place inconsistently is not an option. Maybe in a future refactoring if there's time before the release.

bittensor/core/async_subtensor.py
basfroman: done

bittensor/core/async_subtensor.py
thewhaleking: You don't have to. Just reminding you!

thewhaleking (approved):

PR #3084, opened 2025-10-06
Update get_neuron_for_pubkey_and_subnet to use block_hash instead of block and block_hash

### LINKED ISSUE #3083: AsyncSubtensor  get_neuron_for_pubkey_and_subnet passes block and block_hash to future calls
### Describe the bug

When using an async subtensor and calling https://github.com/opentensor/bittensor/blob/master/bittensor/core/async_subtensor.py#L2519 with the `block` parameter the block_hash is received and passed to the following method https://github.com/opentensor/bittensor/blob/master/bittensor/core/async_subtensor.py#L2530.
This eventually calls https://github.com/opentensor/bittensor/blob/master/bittensor/core/async_subtensor.py#L3818 with both block and block_hash.

### To Reproduce

```
import asyncio

async def main():
    async with AsyncSubtensor("wss://archive.chain.opentensor.ai:443") as subtensor:
        try:
            result = await subtensor.get_neuron_for_pubkey_and_subnet(
                hotkey_ss58="5CnvuQgNLBd2tmhykivRfnJZBYc5D2NZEBYxjFm2fc5Msn1o",
                netuid=10,
                block=6602498,
            )
            print(result.active)
        except Exception as e:
            print(e)
        return


if __name__ == "__main__":
    asyncio.run(main())
````

### Expected behavior

Function should work correctly when block is passed

### Screenshots

_No response_

### Environment

Ubuntu

### Additional context

_No response_

yuma-doc:
block_hash is retrieved above and adding both blockhash and block to the neron_for_uid call results in you getting "Only one of ``block``, ``block_hash``, or ``reuse_block`` can be specified." when the determine_block_hash call occurs later

fixes #3083

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -2530,7 +2530,6 @@ class AsyncSubtensor(SubtensorMixin):
             return await self.neuron_for_uid(
                 uid=uid,
                 netuid=netuid,
-                block=block,
                 block_hash=block_hash,
                 reuse_block=reuse_block,
             )


Swamination-Yuma:
Appears to force an extra block/hash/reuse value through neuron_for_uid() if specifying a uid, resulting in the error.

thewhaleking (approved):

PR #3085, opened 2025-10-07
[v10] Fix `argparser` issue

basfroman:
This PR refactors the Config class to dynamically respect the `BT_NO_PARSE_CLI_ARGS` environment variable, ensuring CLI argument parsing is enabled by default and can be explicitly disabled when needed.
This change prevents SDK crashes when it is used within external scripts that define their own CLI arguments overlapping with the SDK’s.
- The environment variable is recognized when set to any of the following case-insensitive values: 1, true, yes, or on.
- By default, `BT_NO_PARSE_CLI_ARGS` is not set, and CLI argument parsing remains active.
- The local environment variable `BT_NETWORK` has been renamed to `BT_SUBTENSOR_NETWORK` for consistency with other Subtensor-related settings.

@MichaelTrestman — please note that the documentation needs to be updated to reflect:
- the new environment variable `BT_NO_PARSE_CLI_ARGS`, and
- the renamed variable `BT_SUBTENSOR_NETWORK`.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -2,18 +2,12 @@
 
 ## Extrinsics and related
 1. ✅ Standardize parameter order across all extrinsics and related calls. Pass extrinsic-specific arguments first (e.g., wallet, hotkey, netuid, amount), followed by optional general flags (e.g., wait_for_inclusion, wait_for_finalization)
-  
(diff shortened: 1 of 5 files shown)


thewhaleking (approved):
Nice!

PR #3087, opened 2025-10-07
Improves `determine_block_hash` method of `AsyncSubtensor`

### LINKED ISSUE #3083: AsyncSubtensor  get_neuron_for_pubkey_and_subnet passes block and block_hash to future calls

thewhaleking:
Issue #3083 raised the issue of `block_hash` and `block` being passed to `determine_block_hash` when supplying a `block` to `get_neuron_for_pubkey_and_subnet`. 

The short fix for this was #3084, but improving on this is better.

Because `AsyncSubtensor.get_block_hash` is cached, there's really no performance loss in checking the block hash if both the block and block_hash are supplied to `determine_block_hash`.

With this PR, if both a block and block_hash are passed to `determine_block_hash`, it will verify that the actual block hash maps to the corresponding supplied block hash for the given block, raising a ValueError if not.

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -318,31 +318,41 @@ class AsyncSubtensor(SubtensorMixin):
         for blockchain queries.
 
         Arguments:
-            block: The block number to get the hash for. Do not specify if using block_hash or reuse_block.
-            block_hash: The hash of the blockchain block. Do not specify if using block or reuse_block.
-            reuse_block: Whether to reuse the last-used block hash. Do not set if using block or reuse_block.
+            block: The block number to get the hash for. If specifying along with `block_hash`, the hash of `block` will
+                be checked and compared with the supplied block hash, raising a ValueError if the two do not match.
+            block_hash: The hash of the blockchain block. If specifying along with `block`, the hash of `block` will be
+                checked and compared with the supplied block hash, raising a ValueError if the two do not match.
+            reuse_block: Whether to reuse the last-used block hash. Do not set if using block or block_hash.
 
         Returns:
             Optional[str]: The block hash if one can be determined, None otherwise.
 
         Raises:
-            ValueError: If more than one of block, block_hash, or reuse_block is specified.
+            ValueError: If reuse_block is set, while also supplying a block/block_hash, or if supplying a block and
+                block_hash, but the hash of the block does not match the supplied block hash.
 
         Example:
             # Get hash for specific block
             block_hash = await subtensor.determine_block_hash(block=1000000)
 
             # Use provided block hash
-            hash = await subtensor.determine_block_hash(block_hash="0x1234...")
+            hash_ = await subtensor.determine_block_hash(block_hash="0x1234...")
 
             # Reuse last block hash
-            hash = await subtensor.determine_block_hash(reuse_block=True)
+            hash_ = await subtensor.determine_block_hash(reuse_block=True)
         """
-        # Ensure that only one of the parameters is specified.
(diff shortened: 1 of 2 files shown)


bittensor/core/async_subtensor.py
```diff
@@ -318,31 +318,41 @@ async def determine_block_hash(
...
+            ValueError: If reuse_block is set, while also supplying a block/block_hash, or if supplying a block and
+                block_hash, but the hash of the block does not match the supplied block hash.
 
         Example:
             # Get hash for specific block
             block_hash = await subtensor.determine_block_hash(block=1000000)
 
             # Use provided block hash
-            hash = await subtensor.determine_block_hash(block_hash="0x1234...")
+            hash_ = await subtensor.determine_block_hash(block_hash="0x1234...")
```
```suggestion
            block_hash = await subtensor.determine_block_hash(block_hash="0x1234...")
```

bittensor/core/async_subtensor.py
```diff
@@ -318,31 +318,41 @@ async def determine_block_hash(
...
+        if block and block_hash:
+            bh = await self.get_block_hash(block)
+            if bh != block_hash:
+                raise ValueError(
+                    "You have supplied a `block_hash` and a `block`, but the block does not map to the same hash as "
+                    f"the one you supplied. You supplied `block_hash={block_hash}` for `block={block}`, but this block"
+                    f"maps to the block hash {bh}."
+                )
+            else:
+                return bh
```
```suggestion
            block_hash_ = await self.get_block_hash(block)
            if block_hash_ != block_hash:
                raise ValueError(
                    "You have supplied a `block_hash` and a `block`, but the block does not map to the same hash as "
                    f"the one you supplied. You supplied `block_hash={block_hash}` for `block={block}`, but this block"
                    f"maps to the block hash {block_hash_}."
                )
            else:
                return block_hash_
```

bittensor/core/async_subtensor.py
```diff
@@ -318,31 +318,41 @@ async def determine_block_hash(
...
-            hash = await subtensor.determine_block_hash(reuse_block=True)
-        """
-        # Ensure that only one of the parameters is specified.
-        if sum(bool(x) for x in [block, block_hash, reuse_block]) > 1:
-            raise ValueError(
-                "Only one of ``block``, ``block_hash``, or ``reuse_block`` can be specified."
-            )
+            hash_ = await subtensor.determine_block_hash(reuse_block=True)
+        """
+        if reuse_block and any([block_hash, block_hash]):
```
basfroman: are you comparing the same variable?

basfroman (changes requested):

basfroman (approved):

PR #3093, opened 2025-10-10
[v10] All TODOs resolved

basfroman:
## Metagraph changes:
- all methods with `mechid` parameter has reordered list of parameters.
- async `_initialize_subtensor` method no longer kill the subtensor instance after use.

Added:
- e2e tests to check weights and bonds


bittensor/core/async_subtensor.py
```diff
@@ -279,24 +279,7 @@ async def initialize(self):
...
-                f" Verify your chain and network settings: {self}"
-            )
-            raise ConnectionError
-        except (ConnectionRefusedError, ssl.SSLError) as error:
-            logging.error(
-                f"[red]Error[/red]: Connection refused when connecting to substrate. "
-                f"Verify your chain and network settings: {self}. Error: {error}"
-            )
-            raise ConnectionError
+        await self.initialize()
```
thewhaleking: This still needs to return self. Otherwise you can't do `async with AsyncSubtensor() as subtensor`

thewhaleking (changes requested):
Great. Just one small change needed.

bittensor/core/async_subtensor.py
basfroman: right. forgot. fixed

thewhaleking (approved):

Teri68 (approved):

Teri68 (approved):

Teri68 (approved):

PR #3133, opened 2025-11-05
fix: simplify AxonInfo.__eq__ method and add type check

DeVikingMark:
Simplifies the __eq__ method by returning comparison result directly instead of using if-return True/False pattern. 

Adds isinstance check to prevent AttributeError when comparing with non-AxonInfo objects.


basfroman:
hey @DeVikingMark, could you pls read https://github.com/opentensor/bittensor/blob/master/contrib/CONTRIBUTING.md ?

thewhaleking (approved):

basfroman (approved):

PR #3137, opened 2025-11-07
[e2e] fix: e2e test for metagraph expected not matching

camfairchild:
For https://github.com/opentensor/subtensor/pull/2187  
We are correcting an issuance with the emission calculation for subnets that are not root selling.
I'm not sure how this was passing before but it's not now.

### DIFF
--- a/tests/e2e_tests/test_metagraph.py
+++ b/tests/e2e_tests/test_metagraph.py
@@ -217,12 +217,12 @@ def test_metagraph_info(subtensor, alice_wallet, bob_wallet):
         blocks_since_last_step=1,
         subnet_emission=Balance(0),
         alpha_in=Balance.from_tao(10).set_unit(1),
-        alpha_out=Balance.from_tao(block * alpha_per_block).set_unit(1),  # it's been 2 blocks
+        alpha_out=Balance.from_tao(1 + block * alpha_per_block).set_unit(1),  # it's been 2 blocks
         tao_in=Balance.from_tao(10),
-        alpha_out_emission=Balance.from_tao(0).set_unit(1),  # 1 alpha per block
+        alpha_out_emission=Balance.from_tao(block).set_unit(1),  # 1 alpha per block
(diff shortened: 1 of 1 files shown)


tests/e2e_tests/test_metagraph.py
```diff
@@ -209,18 +211,18 @@ def test_metagraph_info(subtensor, alice_wallet, bob_wallet):
...
         owner_coldkey="5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM",
-        block=1,
+        block=block,
         tempo=100,
         last_step=0,
         blocks_since_last_step=1,
         subnet_emission=Balance(0),
         alpha_in=Balance.from_tao(10).set_unit(1),
-        alpha_out=Balance.from_tao(1).set_unit(1),
+        alpha_out=Balance.from_tao((1 + block) * alpha_per_block).set_unit(1),  # it's been 2 blocks
```
thewhaleking: Why has it been 2 blocks if we're checking at block 1?

tests/e2e_tests/test_metagraph.py
```diff
@@ -199,7 +199,9 @@ def test_metagraph_info(subtensor, alice_wallet, bob_wallet):
     alice_subnet_netuid = subtensor.get_total_subnets()  # 2
     subtensor.register_subnet(alice_wallet, True, True)
 
-    metagraph_info = subtensor.get_metagraph_info(netuid=1, block=1)
+    block = 1
```
thewhaleking, line 202: Does this need to be `block = registration block + 1`? Otherwise it's just always using the first block, which is happens before the subnet is even created. But even then we're basically checking a blank slate.

basfroman: in this part of the test we compare first blocks

basfroman (dismissed):

thewhaleking (dismissed):

PR #3139, opened 2025-11-10
[v10] Adds block backwards compatibility for fetching metagraph.

### LINKED ISSUE #3138: Needed: Compatibility for older blocks
### Describe the bug

A number of calls (usually runtime calls, but not always) only work for current runtime versions, a key example being `Subtensor.metagraph`. 

Because of the change in the runtime call from https://github.com/opentensor/bittensor/pull/3056, it is no longer possible to use `Subtensor.metagraph` for calls before that runtime version. But even before that, there was a fairly recent change that made it impossible to make calls older (when we changed to the selective_metagraph runtime call.

We need backwards compatibility (not in terms of API) but in terms of ability to fetch info on older blocks on the chain. 

### To Reproduce

```py
with Subtensor(ARCHIVE_ENTRYPOINT) as subtensor:
    print(subtensor.metagraph(netuid=0, block=6500000))
```

produces:
```py
ValueError: Runtime API Call 'SubnetInfoRuntimeApi.get_selective_mechagraph' not found in registry
```

### Expected behavior

I get the metagraph.

### Screenshots

_No response_

### Environment

any

### Additional context

First step here should be dealing with runtime calls. Queries should be second.

Because runtime versions are sequential, it's fairly simple to define the behavior for a given runtime version for a given call, and then route the logic based on that.

thewhaleking:
Resolves the metagraph portion of #3138 

Demonstrated with:
```py
import asyncio
from bittensor.core.subtensor import Subtensor
from bittensor.core.settings import ARCHIVE_ENTRYPOINT

async def main():
    async with AsyncSubtensor(ARCHIVE_ENTRYPOINT) as subtensor:
        print(await subtensor.metagraph(netuid=0, block=6500000))
        print(await subtensor.metagraph(netuid=0, block=5500000))
        print(await subtensor.metagraph(netuid=0))


def main_sync():
    with Subtensor(ARCHIVE_ENTRYPOINT) as subtensor:
        print(subtensor.metagraph(netuid=0, block=6500000))
        print(subtensor.metagraph(netuid=0, block=5500000))
        print(subtensor.metagraph(netuid=0))

if __name__ == '__main__':
    asyncio.run(main())
    main_sync()
```

### What this does not cover:
Legacy (think 2023) Runtime API methods were defined through dicts. This is still an issue that needs resolved upstream in [async-substrate-interface](https://github.com/opentensor/async-substrate-interface/pull/194). However, once that is resolved, no additional work will need done here, but until then the legacy runtime call will not work.


bittensor/core/async_subtensor.py
```diff
@@ -2673,9 +2673,17 @@ async def get_mechanism_emission_split(
             whole numbers). Returns None if emission is evenly split or if the data is unavailable.
         """
         block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
+        module = "SubtensorModule"
+        storage_function = "MechanismEmissionSplit"
+        if not await self.substrate.get_metadata_storage_function(
+            module,
+            storage_function,
+            block_hash=block_hash,
```
```suggestion
            module_name=module,
            storage_name=storage_function,
            block_hash=block_hash,
```

bittensor/core/async_subtensor.py
```diff
@@ -2703,14 +2711,57 @@ async def get_mechanism_count(
             The number of mechanisms for the given subnet.
         """
         block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
+        module = "SubtensorModule"
+        storage_function = "MechanismCountCurrent"
+        if not await self.substrate.get_metadata_storage_function(
+            module,
+            storage_function,
+            block_hash=block_hash,
```
```suggestion
            module_name=module,
            storage_name=storage_function,
            block_hash=block_hash,
```

basfroman (changes requested):
Good work on adding backwards compatibility for metagraph!

Current approach:
If we're only targeting `get_metagraph_info`, the current solution is fine. However, if we plan to apply this pattern to other methods, we should consider a more scalable approach.

I consider these options:
1. `RuntimeAPIVersionAdapter` (or something like this) in SDK (short-term)
 - create a centralized adapter with method mapping and fallback logic.
 - avoids duplicating logic in each method.
 - can be implemented quickly without changing dependencies.

2. Move logic to `ASI` (preferred long-term)
 - SDK (or any other clients) shouldn't need to worry about which method to use.
 - the low-level library should handle version detection and return the right data for a given block.
 - better separation of concerns: SDK focuses on business logic, version handling belongs in the substrate layer.
 - reusable for other clients.

thewhaleking:
> 2. Move logic to `ASI` (preferred long-term)
>  - SDK (or any other clients) shouldn't need to worry about which method to use.
>  - the low-level library should handle version detection and return the right data for a given block.
>  - better separation of concerns: SDK focuses on business logic, version handling belongs in the substrate layer.
>  - reusable for other clients.

The logic cannot live in ASI. ASI is a general-purpose Substrate library, and thus does have any concept of which methods retrieve which data.

thewhaleking:
The failing tests are failing upstream. Seems a new devnet-ready release broke them.

basfroman (approved):
Looks great! Please resolve the conflicts and let's merge it.

basfroman:
> The failing tests are failing upstream. Seems a new devnet-ready release broke them.

Yes, tomorrow in the next PR these tests will be fixed.

Issue #3128, opened 2025-10-31
Logging directory and file handler behavior inconsistent with runtime overrides


basfroman:
When `BT_LOGGING_LOGGING_DIR` or `BT_LOGGING_RECORD_LOG` are set, updating logging_dir at runtime (e.g., inside a validator) doesn’t redirect output — existing handlers continue logging to the original path. The SDK should correctly reinitialize or respect the new directory when updated.

Issue #3111, opened 2025-10-23
`btlogging` disables another loggers by default


basfroman:
Investigation + possibly improvement

PR #3143, opened 2025-11-12
[v10] Improve btlogging

basfroman:
- [x] fix for https://github.com/opentensor/bittensor/issues/3128
- [x] Reduce log level column width from 16 to 8
- [x] logging shows the actual calling module
- [x] extending logic regarding https://github.com/opentensor/bittensor/issues/3111 

Added logic for enabling/disabling third-party loggers. This is accomplished using the local environment variable `BT_LOGGING_ENABLE_THIRD_PARTY_LOGGERS`. @MichaelTrestman please add new local env variable to the documentation.
  
**Note:** Regardless of whether third-party loggers are enabled or disabled, the SDK logger is a primary logger and filters logging messages from third-party loggers by the SDK logging level.

### DIFF
--- a/bittensor/utils/btlogging/loggingmachine.py
+++ b/bittensor/utils/btlogging/loggingmachine.py
@@ -17,7 +17,7 @@ from typing import NamedTuple, Union
 from statemachine import State, StateMachine
 
 from bittensor.core.config import Config
-from bittensor.core.settings import READ_ONLY
+from bittensor.core.settings import DEFAULTS
 from bittensor.utils.btlogging.console import BittensorConsole
 from .defines import (
     BITTENSOR_LOGGER_NAME,
@@ -53,6 +53,7 @@ class LoggingConfig(NamedTuple):
     info: bool
     record_log: bool
     logging_dir: str
+    enable_third_party_loggers: bool
 
 
 class LoggingMachine(StateMachine, Logger):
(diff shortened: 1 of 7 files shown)


bittensor/utils/btlogging/loggingmachine.py
```diff
@@ -12,12 +12,12 @@
 import sys
 from logging import Logger
 from logging.handlers import QueueHandler, QueueListener, RotatingFileHandler
-from typing import NamedTuple
+from typing import NamedTuple, Union
```
thewhaleking, line 15: Because we now only support 3.10+, you don't actually need to use `typing.Union` anymore

You can just write `dict | Config`

tests/unit_tests/test_logging.py
```diff
@@ -234,3 +234,96 @@ def test_logger_level(logging_machine, caplog):
...
+        info=False,
+        record_log=True,
+        logging_dir=str(initial_log_dir),
+    )
+    lm = LoggingMachine(initial_config)
+
+    # Verify initial file handler exists and points to initial directory
+    file_handler = None
+    for handler in lm._handlers:
+        if isinstance(handler, RotatingFileHandler):
```
thewhaleking, line 288: Why check RotatingFileHandler rather than FileHandler,  which is RotatingFileHandler's base class?

ibraheem-abe (approved):

ibraheem-abe (approved):

PR #3148, opened 2025-11-18
[v10] Fix async extrinsic receipt coroutine reuse error 

basfroman:
### Problem
Accessing `ExtrinsicResponse.extrinsic_receipt` fields in async code could raise `RuntimeError: cannot reuse already awaited coroutine` if the same coroutines were awaited multiple times. Users experiencing this behavior now.

### Solution:
All async fields from `AsyncExtrinsicReceipt` are awaited once via `asyncio.gather` in `async_apply_extrinsic_receipt` and stored in a synchronous `SDKExtrinsicReceipt`. After that, all fields are accessed synchronously.

### Benefits:
- Users no longer hit RuntimeError when accessing `extrinsic_receipt` fields
- Consistent ExtrinsicResponse interface — extrinsic_receipt is always synchronous, whether using Subtensor or AsyncSubtensor

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -5284,7 +5284,7 @@ class AsyncSubtensor(SubtensorMixin):
                 wait_for_inclusion=wait_for_inclusion,
                 wait_for_finalization=wait_for_finalization,
             )
-            await extrinsic_response.async_apply_extrinsic_receipt(response)
+
             # We only wait here if we expect finalization.
             if not wait_for_finalization and not wait_for_inclusion:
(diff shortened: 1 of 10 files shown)


thewhaleking:
Which fields are causing issues? Each of the fields of `AsyncExtrinsicReceipt` are coroutine functions rather than coroutine objects, and thus should be infinitely awaitable.

basfroman:
> Which fields are causing issues? Each of the fields of `AsyncExtrinsicReceipt` are coroutine functions rather than coroutine objects, and thus should be infinitely awaitable.

Yes, you're right - fields of `AsyncExtrinsicReceipt`. But if the user needs to access the awaitable field more than once, it will cause an error. The user expects the response object has reusable data.

PR #3155, opened 2025-11-25
[v10] fix: resolve python 3.14 syntax warnings and update deps

Arthurdw:
- Move return statements out of finally blocks in axon.py and dendrite.py (SyntaxWarning: 'return' in a 'finally' block is deprecated in 3.14, will be an error in 3.16)
- Relax aiohttp constraint from ~=3.9 to >=3.9,<4.0

### Bug

Python 3.14 raises SyntaxWarning for `return` statements inside `finally` blocks, which will become errors in Python 3.16.

### Description of the Change

- Move return statements out of finally blocks in `axon.py:1211` and `dendrite.py:392,601`
- Relax aiohttp constraint from `~=3.9` to `>=3.9,<4.0` to allow newer compatible versions

### Alternate Designs

N/A - this is the standard fix for the deprecation warning.

### Possible Drawbacks

None - the behavior is identical, just restructured to avoid the deprecated pattern.

### Verification Process

Ran test suite on Python 3.14 and confirmed warnings are resolved.

### Release Notes

Fixed Python 3.14 deprecation warnings for return statements in finally blocks.

### Branch Acknowledgement
- [x] I am acknowledging that I am opening this branch against `SDKv10` (for sdkv10 not v9, so not staging. this is correct right?)

### DIFF
--- a/bittensor/core/axon.py
+++ b/bittensor/core/axon.py
@@ -1207,8 +1207,8 @@ class AxonMiddleware(BaseHTTPMiddleware):
                     f"axon     | --> | {response.headers.get('content-length', -1)} B | {synapse.name} | None | None | 200 | Success "
                 )
 
-            # Return the response to the requester.
-            return response
+        # Return the response to the requester.
+        return response
 
     async def preprocess(self, request: "Request") -> "Synapse":
         """
--- a/bittensor/core/dendrite.py
+++ b/bittensor/core/dendrite.py
@@ -389,7 +389,8 @@ class DendriteMixin:
             use_new_loop = True
         finally:
             self.close_session(using_new_loop=use_new_loop)
-            return result  # type: ignore
+
+        return result  # type: ignore
 
     async def forward(
         self,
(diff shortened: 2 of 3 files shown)


Arthurdw:
Uh, how come only 6 checks run? Is this expected/? I remember there being way more checks?

thewhaleking:
> Uh, how come only 6 checks run? Is this expected/? I remember there being way more checks?

SDKv10 is a temporary "staging" branch, so we didn't update the workflows to automatically run all tests on PRs opened against it. It requires certain labels to run.

thewhaleking (approved):

basfroman:
> > Uh, how come only 6 checks run? Is this expected/? I remember there being way more checks?
> 
> SDKv10 is a temporary "staging" branch, so we didn't update the workflows to automatically run all tests on PRs opened against it. It requires certain labels to run.

This isn't accurate. All workflows are configured correctly from the scratch, tests and checks run automatically on every PR to this branch. Since I previously created PRs from an account in our organization, I didn't account for the fact that GitHub Actions repository variables aren't available for PRs from forks. This is a GitHub security restriction, but still my bad too.

Thanks for your efforts, but unfortunately, PR #3156 doesn't fix this for a few reasons: the inherit flag is set by default, so adding it only clarifies intent. Even with inherit, PRs from forks don't get access to GitHub Actions repository variables.

I've created PR #3158 to address this. To be more accurate, I created a PR from the fork under an account that is not part of the organization. Now, instead of storing the list of supported Python versions in GitHub Settings, I now use a JSON file and import it in the workflows for use.

@Arthurdw thank you! Your fork PR helped us discover this issue on early stage! 
Pls update you PR with updated (after #3158 merged) SDKv10 branch and I'll merge it. After you pull the SDKv10 changes, you have to see all tests and checkers in your PR.

PR #3156, opened 2025-11-25
Fix: run e2e single

thewhaleking:
Ensures the e2e test runner inherits the vars from GH Actions vars so that the _run-e2e-single.yaml has the needed context

See failures here: https://github.com/opentensor/bittensor/actions/runs/19667665405

for PR #3155


basfroman:
Unfortunately, this PR doesn't related with the issue in PR https://github.com/opentensor/bittensor/pull/3155. 
The tests ran and passed because the PR was created under gh profile within the organization, not from a fork.

PR #3191, opened 2025-12-05
Feat/new mev

camfairchild:
Replaces the MEV shield design with a simpler one to just submit a signed extrinsic payload to the node, encrypted.


basfroman:
implemented here https://github.com/opentensor/bittensor/pull/3192

PR #3203, opened 2025-12-09
[v10] `Delegate` set claim type support

basfroman:
## Validator Claim Type Delegation

This PR implements validator claim type delegation, allowing validators to set their claim behavior (Swap/Keep) per subnet. Stakers with `Delegated` claim type will inherit the validator's claim type when claiming root emissions.

### New Methods
- `subtensor.get_validator_claim_type(hotkey_ss58, netuid)` - Get validator's claim type for a subnet
- `subtensor.set_validator_claim_type(wallet, hotkey_ss58, netuid, new_claim_type)` - Set validator's claim type
- `set_validator_claim_type_extrinsic` - Low-level extrinsic calls (sync/async)

### NOTE: Validator can't set `Delegated` claim type

Think of this logic like this:
- The validator is the **chef** who chooses the menu (Swap/Keep) for their restaurant (subnet)
- The staker with `Delegated` is the **guest** who says _"I'll have whatever the chef recommends"_
- The staker with `Swap/Keep` is the **guest** who orders a specific dish regardless of chef's recommendation
- If the chef says _"as the guest decides"_ and the guest says _"as the chef decides"_, the decision is undecided. (This is why the validator can't set Delegated.)

### Example
- Validator Alice sets `Keep` on subnet 2
- Staker Charlie has `Delegated` claim type (default)
- When Charlie claims root emissions, they inherit Alice's `Keep` behavior → alpha stays on subnet 2, not swapped to TAO

### Testing
- All tests pass successfully on a locally compiled node.

Subtensor base PR: https://github.com/opentensor/subtensor/pull/2224

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -96,6 +96,7 @@ from bittensor.core.extrinsics.asyncex.root import (
     claim_root_extrinsic,
     root_register_extrinsic,
     set_root_claim_type_extrinsic,
+    set_validator_claim_type_extrinsic,
 )
 from bittensor.core.extrinsics.asyncex.serving import (
     publish_metadata_extrinsic,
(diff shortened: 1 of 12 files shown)


basfroman:
Subtensor based PR is closed

PR #3212, opened 2025-12-12
Refactor AsyncSubtensor and Dendrite to offload signing to ProcessPoolExecutor 

### LINKED ISSUE #3211: Optimization Offload CPU-bound Keypair.sign to ProcessPoolExecutor to prevent Event Loop blocking in AsyncSubtensor
While AsyncSubtensor and SDK v10 have significantly improved the ability to handle concurrent network operations, a critical performance bottleneck remains regarding cryptographic operations.

Python's asyncio event loop is single-threaded. Consequently, CPU bound tasks executed within an async function will block the entire loop until they complete. The most frequent and computationally expensive CPU bound task in the Bittensor protocol is signing transactions and messages (for example, Keypair.sign).

## The Scenario
When a validator operating on a large subnet (for example, 1024 UIDs) attempts to broadcast queries or set weights:

1. The code iterates through the list of UIDs.  
2. For every request or extrinsic, it calls dendrite.keypair.sign() or wallet.hotkey.sign().  
3. These signing operations are currently synchronous.  

**Result:** The main event loop freezes while the CPU performs Elliptic Curve arithmetic.

## Impact
During this freeze period:

- **WebSocket Heartbeats Fail:** KeepAlive pings to the Substrate node are not sent or processed, which leads to `1006 Connection Closed` errors.  
- **Network I/O Paused:** Incoming responses from miners are buffered at the OS level but not processed by the application, causing artificial latency spikes.  
- **Throughput Cap:** The validator cannot effectively use available bandwidth because it is bottlenecked by the single core signing speed.

## Proposed Solution
We should refactor the signing logic within `AsyncSubtensor` and `Dendrite` to offload these CPU-bound tasks to a separate process, bypassing the GIL and keeping the event loop responsive.

This can be achieved using `asyncio.loop.run_in_executor` with a `ProcessPoolExecutor`.

### Recommended Implementation Pattern

```python
import asyncio
from concurrent.futures import ProcessPoolExecutor
from functools import partial

# In the initialization of the class (e.g., AsyncSubtensor or a global config)
self._executor = ProcessPoolExecutor()

async def sign_async(self, data: bytes) -> str:
    loop = asyncio.get_running_loop()
    # Offload the heavy synchronous 'sign' method to a separate process
    signature = await loop.run_in_executor(
        self._executor, 
        partial(self.keypair.sign, data=data)
    )
    return signature

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by Dairus01**
## Description
This PR addresses critical Event Loop blocking caused by CPU-intensive cryptographic signing operations in the Bittensor SDK. Specifically, the `Keypair.sign` method (Elliptic Curve math) was being called synchronously on the main asyncio thread in both [AsyncSubtensor](cci:2://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:185:0-9647:9) and [Dendrite](cci:2://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:907:0-911:44), causing heartbeats to be missed and significant latency during high-volume operations (e.g., signing extrinsics for hundreds of miners).

## Changes
- **[bittensor/core/async_subtensor.py](cci:7://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:0:0-0:0)**: 
    - Implemented [create_signed_extrinsic](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:5986:4-6029:24) to offload payload signing to a `ProcessPoolExecutor`.
    - Added [sign_extrinsic_worker](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:170:0-174:40) helper to handle signing in a separate process (pickling-safe).
    - Updated [sign_and_send_extrinsic](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:5963:4-6078:37) to use the async signing flow.
- **[bittensor/core/dendrite.py](cci:7://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:0:0-0:0)**: 
    - Updated [preprocess_synapse_for_request](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:710:4-754:22) to be [async](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/async_subtensor.py:9650:0-9686:14) and offload header signing to a `ProcessPoolExecutor`.
    - Added [sign_message_worker](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:43:0-53:40) helper.
    - Updated [call](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:560:4-626:64), [call_stream](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:628:4-708:29), and [forward](cci:1://file:///c:/Users/user/Desktop/bittensor/bittensor/core/dendrite.py:418:4-558:98) methods to await the async preprocessing.
- **Executor Management**: Added initialization and graceful shutdown (via `close/aclose`) of the `ProcessPoolExecutor` in both classes.

## Related Issues
Fixes #3211

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING, Any, Iterable, Literal, Optional, Union, cast
 
 import asyncstdlib as a
 import scalecodec
+from concurrent.futures import ProcessPoolExecutor
+from functools import partial
 from async_substrate_interface import AsyncSubstrateInterface
 from async_substrate_interface.substrate_addons import RetryAsyncSubstrate
 from async_substrate_interface.utils.storage import StorageKey
@@ -163,6 +165,16 @@ from bittensor.utils.liquidity import (
     tick_to_price,
 )
 
+
+
+
+def sign_extrinsic_worker(private_key: str, payload: bytes) -> str:
+    from bittensor_wallet import Keypair
+    # We use the private key to recreate the keypair for signing
+    kp = Keypair(private_key=private_key)
+    return f"0x{kp.sign(payload).hex()}"
+
+
 if TYPE_CHECKING:
     from async_substrate_interface import AsyncQueryMapResult
     from async_substrate_interface.types import ScaleObj
@@ -253,6 +265,8 @@ class AsyncSubtensor(SubtensorMixin):
             logging.info(
                 f"Connected to {self.network} network and {self.chain_endpoint}."
             )
+        
+        self._executor = ProcessPoolExecutor(max_workers=1)
 
     async def close(self):
         """Closes the connection to the blockchain.
@@ -278,6 +292,10 @@ class AsyncSubtensor(SubtensorMixin):
         """
         if self.substrate:
             await self.substrate.close()
+
+        if self._executor:
+            self._executor.shutdown(wait=True)
+            self._executor = None
 
     async def initialize(self):
         """Establishes connection to the blockchain.
@@ -328,7 +346,12 @@ class AsyncSubtensor(SubtensorMixin):
         return await self.initialize()
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
-        await self.substrate.close()
+        if self.substrate:
+            await self.substrate.close()
+        
+        if self._executor:
+            self._executor.shutdown(wait=True)
+            self._executor = None
 
     # Helpers ==========================================================================================================
 
@@ -5961,6 +5984,51 @@ class AsyncSubtensor(SubtensorMixin):
             block_hash=block_hash,
         )
 
+    async def create_signed_extrinsic(
+        self,
+        call: "GenericCall",
+        keypair: "Keypair",
+        nonce: Optional[int] = None,
+        era: Optional[dict] = None,
+        tip: int = 0,
+        tip_asset_id: Optional[int] = None,
+    ):
+        """
+        Creates a signed extrinsic, offloading the signing process to a separate process to avoid blocking the event loop.
+        """
+        if nonce is None:
+            nonce = await self.substrate.get_account_next_index(keypair.ss58_address)
+
+        # Generate the signature payload.
+        # This uses the underlying substrate interface to prepare the bytes to be signed.
+        signature_payload = await self.substrate.generate_signature_payload(
+            call=call,
+            era=era,
+            nonce=nonce,
+            tip=tip,
+            tip_asset_id=tip_asset_id
+        )
+
+        # Offload the signing of the payload to an executor.
+        loop = asyncio.get_running_loop()
+        # signature_payload.data should contain the bytes to sign.
+        signature = await loop.run_in_executor(
+            self._executor,
+            partial(sign_extrinsic_worker, keypair.private_key, signature_payload.data)
+        )
+
+        # Create the extrinsic with the attached signature.
+        extrinsic = await self.substrate.create_extrinsic(
+            call=call,
+            nonce=nonce,
+            era=era,
+            tip=tip,
+            tip_asset_id=tip_asset_id,
+            signature=signature,
+            keypair=keypair,  # Keypair may be needed for address/type info
+        )
+        return extrinsic
+
     async def sign_and_send_extrinsic(
         self,
         call: "GenericCall",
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

**Comment by Dairus01** [comment]
@thewhaleking 
Please I'd love to hear your thoughts on this PR

**Comment by thewhaleking** [comment]
> @thewhaleking 
> 
> Please I'd love to hear your thoughts on this PR

 Will review and reply on Monday when I'm back

**Comment by basfroman** [comment]
Hi @Dairus01 

I'll close all PRs from Gittensor miners if you guys are so disrespectful:
1. PRs for `master`, not `staging`
2. A bunch of failed test cases

Conclusion: you didn't read https://github.com/opentensor/bittensor/blob/master/contrib/CONTRIBUTING.md

With all due respect to your work, I'd like to see the fixes when I return to this on Monday. 
If you want us to spend time on your PR, please respect our time too.

**Review by basfroman — CHANGES_REQUESTED** [review]
https://github.com/opentensor/bittensor/pull/3212#issuecomment-3648200191

**Comment by Dairus01** [comment]
I have changed the branch from master to staging, I will work on the checks, once again thanks for your time I will make sure that I get everything fixed in the next two hours

**Comment by Dairus01** [comment]
I have work on the issues I would love to know the update

File: bittensor/core/async_subtensor.py [hunk 1]
```diff
@@ -163,6 +165,15 @@
     tick_to_price,
 )
 
+
+def sign_extrinsic_worker(private_key: str, payload: bytes) -> str:
+    from bittensor_wallet import Keypair
```
basfroman, line 170: import should be on the top of the module

bittensor/core/async_subtensor.py
```diff
@@ -163,6 +165,15 @@
     tick_to_price,
 )
 
+
+def sign_extrinsic_worker(private_key: str, payload: bytes) -> str:
```
basfroman, line 169: function should have proper docstring with the detailed parameters description

basfroman, line 169: `sign_extrinsic_worker` is duplicating of `sign_message_worker`
It would be great to move this into `bittensor/utils/registration` as a single function.

bittensor/core/async_subtensor.py
```diff
@@ -5961,6 +5983,47 @@ async def compose_call(
...
+        extrinsic = await self.substrate.create_extrinsic(
+            call=call,
+            nonce=nonce,
+            era=era,
+            tip=tip,
+            tip_asset_id=tip_asset_id,
+            signature=signature,
+            keypair=keypair,  # Keypair may be needed for address/type info
+        )
+        return extrinsic
```
basfroman, line 6025: you don't need this function in subtensor modules. all you want to do:
- create `signature`
- pass that `signature` to `async_substrate_interface.sync_substrate.SubstrateInterface.create_signed_extrinsic`
- you will get `signed extrinsic`

bittensor/core/dendrite.py
```diff
@@ -39,6 +41,19 @@ def event_loop_is_running():
...
+    from bittensor_wallet import Keypair
+
+    # Re-instantiate the keypair from the private key
+    # ss58_address is required but can be dummy if only signing?
+    # Actually bittensor_wallet Keypair usually takes ss58_address.
+    # If private_key is provided, it might derive it.
+    # Let's try passing None or empty string if allowed, or check Keypair usage.
+    # Safe bet: pass private key. Keypair usually derives public from private.
+    kp = Keypair(private_key=private_key_hex)
+    return f"0x{kp.sign(message).hex()}"
```
basfroman, line 54: read above about DRY

basfroman (changes requested):
Before submitting a PR, always perform the following steps in your virtual environment:
- `make check` - formatting and linters
- `pytest tests/unit_tests` - run unit tests
- `pytest tests/integration_tests` - run integration tests
- would be nice `pytest tests/e2e_tests` - to check that all e2e still aren't broken with your changes

Please fix all comments, linters and format, broken test issues and we can move forward

basfroman:
**PR is redundant and should be closed**
`ProcessPoolExecutor` usage is redundant for this case. `asyncio.to_thread()` (python 3.9+) is sufficient. They are simpler, faster, and solve the event loop blocking problem. But `asyncio.to_thread` should be applied to ASI in `AsyncSubstrateInterface.create_signed_extrinsic` method, but not in SDK.

PR #3215, opened 2025-12-16
Add insufficient balance transfer tests

Kikeskr:
Problem:
Transfer behavior when wallet balance is insufficient was not explicitly covered
by unit tests.

Solution:
Added sync and async tests ensuring transfers fail safely when the wallet balance
is insufficient.

Testing:
pytest tests/unit_tests/test_transfer.py

### DIFF
--- a/tests/e2e_tests/test_transfer.py
+++ b/tests/e2e_tests/test_transfer.py
@@ -10,6 +10,32 @@ if typing.TYPE_CHECKING:
     from bittensor.extras import SubtensorApi
 
 
+def test_transfer_insufficient_balance(subtensor, alice_wallet):
+    """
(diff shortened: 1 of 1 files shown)


basfroman (changes requested):
Hey @Kikeskr , thank you for contribution.

But I have to ask you to re-write the test logic, bc of 2 issues:

1. I'd recommend to add this case test in the test `test_transfer/_async` before the success call.

    And here is why I asking you to do it: we running all e2e tests with all supported python version. Currently they are 3.10-3.14. You added 2 tests (sync and async versions). After all the workflow have to run 10 new instances of the chain to run those small tests. 

2. You have to create the PR to `staging`, not to `master` branch.

Pls apply suggested approach.

PR #3228, opened 2025-12-22
MeV: fix for weird edge case

basfroman:
case: `block X` contains an `EncryptedSubmitted` event and in the same `block X` there is an `AddStake` event that was MeV executed


bittensor/core/extrinsics/asyncex/mev_shield.py
thewhaleking, line 27: if you're now starting at the submission block, shouldn't you change timeout blocks to 4?

basfroman: nope

thewhaleking (approved):

PR #3236, opened 2025-12-30
Fix for async metagraph initialization

basfroman:


ibraheem-abe (dismissed):

bittensor/core/metagraph.py
```diff
@@ -1422,7 +1422,7 @@ async def _initialize_subtensor(
             # Lazy import due to circular import (subtensor -> metagraph, metagraph -> subtensor)
             from bittensor.core.async_subtensor import AsyncSubtensor
 
-            self.subtensor = AsyncSubtensor(network=self.chain_endpoint)
+            subtensor = AsyncSubtensor(network=self.chain_endpoint)
             await self.subtensor.initialize()
             self.subtensor = subtensor
```
thewhaleking, line 1427: Wouldn't it make more sense to do
```py
  if not subtensor:

      self.subtensor = AsyncSubtensor(network=self.chain_endpoint)
      await self.subtensor.initialize()
      subtensor = self.subtensor
  return subtensor
```

because otherwise it appears line 1426 could be calling `.initialize()` on a `None`.  This also ensures that `self.subtensor` is always an `AsyncSubtensor` object after calling `self. _initialize_subtensor`

thewhaleking (commented):
I could be wrong here.

thewhaleking (approved):

ibraheem-abe (approved):

PR #3237, opened 2025-12-31
Fix memory leak in get_async_subtensor on initialization failure

### LINKED ISSUE #3235: Memory leak if subtensor created/removed in a loop
### Describe the bug

      subtensor: bt.async_subtensor | None = None
      try:
          subtensor = await get_async_subtensor(self._subtensor_endpoint)
      except Exception as e:
          raise e
      finally:
          if subtensor is not None:
              logger.info("Closing subtensor")
              await subtensor.close()

If the code above is running in a loop then it causes memory leak

Version 10.0.1 is used.

### To Reproduce

1. Run the code above in a loop and track memory usage of the process

### Expected behavior

No memory leak

### Screenshots

_No response_

### Environment

Linux Ubuntu

### Additional context

_No response_

Dairus01:
# Memory Leak Fix in `get_async_subtensor`

Closes #3235

## Problem Description

A memory leak was identified in the `bittensor` library when using the `get_async_subtensor` factory function in a loop, specifically when the initialization step fails.

When `get_async_subtensor` is called, it performs two main steps:
1. Instantiates a new `AsyncSubtensor` object.
2. Awaits the `.initialize()` method on this object to establish a connection to the Substrate node.

If the `.initialize()` method raises an exception (e.g., due to a connection error, timeout, or invalid configuration), the exception propagates immediately to the caller.  Crucially, the partially initialized `AsyncSubtensor` instance—which may have already allocated resources such as an `AsyncSubstrateInterface` with pending tasks or open (but failing) connections—is lost.

Because the exception interrupts the flow before the instance is returned, the caller never receives the `subtensor` object. Consequently, the caller cannot call `.close()` on it. This leaves the underlying resources (like the `aiohttp` session or background tasks within `async_substrate_interface`) hanging, leading to a memory leak that becomes significant if the operation is retried in a loop.

## Thought Process

1.  **Analysis of the Issue:**
    -   The issue report highlighted a leak when `get_async_subtensor` is called repeatedly.
    -   I examined the code for `get_async_subtensor` in `bittensor/core/async_subtensor.py`.
    -   Code before fix: 
        ```python
        sub = AsyncSubtensor(network=network, ...)
        await sub.initialize()  # <--- If this fails, 'sub' is lost
        return sub
        ```
    -   If `sub.initialize()` fails, the `sub` variable goes out of scope, but if `initialize` started async tasks that reference `self`, the object might be kept alive by the event loop or other references, preventing garbage collection. Explicit cleanup via `.close()` is required for `AsyncSubtensor`.

2.  **Reproduction:**
    -   I created a reproduction script that mocked `AsyncSubstrateInterface` to simulate initialization failures.
    -   I observed that when `initialize` raises an exception, the `close` method of the underlying interface was *not* called.
    -   I confirmed that in a loop, this leads to an accumulation of unclosed objects.

3.  **Verification of Synchronous Implementation:**
    -   I checked `bittensor/core/subtensor.py`. The synchronous `Subtensor` class is typically instantiated directly (`Subtensor(...)`). Its initialization happens in `__init__`, and if that fails, the object is generally not created or fully formed. More importantly, there isn't a factory function performing a two-step "create then async init" process that hides the instance on failure. Thus, the issue appeared specific to the async factory pattern.

## Solution

The solution is to wrap the initialization step in a `try...except` block within the `get_async_subtensor` function. This ensures that if initialization fails, we explicitly close the `AsyncSubtensor` instance before allowing the exception to propagate. 

**Updated Code:**

```python
async def get_async_subtensor(
    network: Optional[str] = None,
    config: Optional["Config"] = None,
    mock: bool = False,
    log_verbose: bool = False,
) -> "AsyncSubtensor":
    # ... (instantiation)
    sub = AsyncSubtensor(
        network=network, config=config, mock=mock, log_verbose=log_verbose
    )
    
    # Wrap initialization to ensure cleanup on failure
    try:
        await sub.initialize()
        return sub
    except Exception as e: 
        # If initialization fails, close the instance to free resources
        await sub.close()
        raise e
```

By adding this error handling, we guarantee that any `AsyncSubtensor` created by this factory is either successfully initialized and returned, or properly closed and discarded. This prevents the resource leak even when connection errors occur repeatedly.

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=201190161

### DIFF
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -1,7 +1,10 @@
 from .core.settings import __version__, DEFAULTS, DEFAULT_NETWORK
 from .utils.btlogging import logging
 from .utils.async_substrate_interface_patch import apply_patch
-# Apply the memory leak patch for AsyncSubstrateInterface
+# Apply the memory leak patch for AsyncSubstrateInterface *before* importing anything
+# that may create AsyncSubstrateInterface instances. In particular, easy_imports
+# pulls in AsyncSubtensor, which uses AsyncSubstrateInterface, so it must be
+# imported only after apply_patch() has been called. Do not reorder these imports.
 apply_patch()
 
 from .utils.easy_imports import *
--- a/bittensor/__init__.py
+++ b/bittensor/__init__.py
@@ -1,3 +1,7 @@
 from .core.settings import __version__, DEFAULTS, DEFAULT_NETWORK
 from .utils.btlogging import logging
+from .utils.async_substrate_interface_patch import apply_patch
+# Apply the memory leak patch for AsyncSubstrateInterface
+apply_patch()
+
 from .utils.easy_imports import *
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -9481,5 +9481,9 @@ async def get_async_subtensor(
     sub = AsyncSubtensor(
         network=network, config=config, mock=mock, log_verbose=log_verbose
     )
-    await sub.initialize()
-    return sub
+    try:
+        await sub.initialize()
+        return sub
+    except Exception as e:
+        await sub.close()
+        raise e
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -9486,4 +9486,4 @@ async def get_async_subtensor(
         return sub
     except Exception as e:
         await sub.close()
-        raise e
+        raise
--- a/bittensor/utils/async_substrate_interface_patch.py
+++ b/bittensor/utils/async_substrate_interface_patch.py
@@ -1,0 +1,60 @@
+import weakref
+from async_substrate_interface.utils import cache
+
+
+class WeakMethodCallable:
+    """
+    A callable that holds a weak reference to a bound method.
+    Used to break reference cycles in CachedFetcher.
+    """
+
+    def __init__(self, bound_method):
+        self._weak_method = weakref.WeakMethod(bound_method)
+
+    async def __call__(self, *args, **kwargs):
+        method = self._weak_method()
+        if method is None:
+            raise ReferenceError("Method instance has been garbage collected")
+        return await method(*args, **kwargs)
+
+
+def _new_get(self, instance, owner):
+    """
+    Patched __get__ method for _CachedFetcherMethod to use WeakKeyDictionary
+    and WeakMethodCallable, preventing memory leaks.
+    """
+    if instance is None:
+        return self
+
+    # Migration/Safety: Ensure _instances is a WeakKeyDictionary.
+    # If it's the original dict (from before patching), replace it.
+    if not isinstance(self._instances, weakref.WeakKeyDictionary):
+        self._instances = weakref.WeakKeyDictionary()
+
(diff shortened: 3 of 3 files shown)


Dairus01:
@basfroman Happy New Year, 
Please, I would love to hear your thoughts on this PR

kudroma404:
@Dairus01 I am not sure that noticed any errors in subtensor initialization. But each iteration memory increases 50 Mb approximately.

kudroma404:
@Dairus01 we ran the code on Ubuntu in Runpod.

NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

kudroma404:
@Dairus01 @basfroman how to reproduce:

main.py
```
import bittensor as bt
import asyncio

async def main():
    while True:
        subtensor = await bt.get_async_subtensor("finney")
        await subtensor.close()
        await asyncio.sleep(10)
        print("Iteration")

if __name__ == "__main__":
    asyncio.run(main())
```

main.config.js
```
module.exports = {
  apps: [{
    name: 'bittensor_test',
    script: 'main.py',
    interpreter: '/usr/bin/python3.13',
  }]
};
```

You can see on the screenshot below that app size increases each 10-15 seconds:

{838BD804-8DF0-4B1E-8498-7406F61B1071}

Dairus01:
I would look into this and get back with a solution 
> @Dairus01 @basfroman how to reproduce:
> 
> main.py
> ```
> import bittensor as bt
> import asyncio
> 
> async def main():
>     while True:
>         subtensor = await bt.get_async_subtensor("finney")
>         await subtensor.close()
>         await asyncio.sleep(10)
>         print("Iteration")
> 
> if __name__ == "__main__":
>     asyncio.run(main())
> ```
> 
> main.config.js
> ```
> module.exports = {
>   apps: [{
>     name: 'bittensor_test',
>     script: 'main.py',
>     interpreter: '/usr/bin/python3.13',
>   }]
> };
> ```
> 
> You can see on the screenshot below that app size increases each 10-15 seconds:
> 
> {838BD804-8DF0-4B1E-8498-7406F61B1071}
>

basfroman:
Hey @Dairus01 , @kudroma404 , I am confident that the behavior you are describing is related to the implementation of `AsyncSubstrateInterface`, but not the SDK itself. I would recommend that you investigate the leak (if one exists) within the `async_substrate_interface` package.

basfroman:
Memory Leak Fix logic have to leave in ASI
See https://github.com/opentensor/async-substrate-interface/issues/255

PR #3270, opened 2026-02-19
Fix logging.info and state transitions in LoggingMachine

### LINKED ISSUE #1836: "logging.info" is still broken from version 6.10.0
### Describe the bug

 It seems the new update 6.12.0 is mandatory to use testnet now. However, logging is still broken starting from version 6.10.0. 

bt.logging.info doesn't show up. only warn and error show up 

https://discord.com/channels/1120750674595024897/1179557631061413968/1235755142377373756

### To Reproduce

bt.logging.info doesn't show up. only warn and error show up 

### Expected behavior

bt.logging.info doesn't show up. only warn and error show up 

### Screenshots

_No response_

### Environment

Unix
BT version >= 6.10.0
Python 3.10.13

### Additional context

_No response_

ionodeionode:
Fixes #1836. 

This PR addresses the issue where `logging.info` and other logging levels were not correctly updating the logger level during state transitions. 

### Changes:
- Explicitly set `self._logger.setLevel()` in `before_enable_info`, `before_enable_debug`, etc.
- Standardized transition methods (`disable_info`, `disable_debug`, etc.) to use `self.enable_default()` to ensure consistent state reset.
- Updated `before_enable_default` and `before_enable_info` to correctly toggle `self._stream_formatter.set_trace()`.

Verified with a test script simulating both CLI flag and manual `set_info` calls.


thewhaleking (commented):
I think this is fine, but I want @basfroman's opinion aswell.

basfroman (changes requested):
Thanks for looking into this! 
I reviewed the changes in detail and have some concerns:

**1. `disable_*` transitions replaced with plain methods**

Converting `disable_trace`, `disable_debug`, etc. from statemachine transitions to regular methods means `before_transition` / `after_transition` hooks no longer fire for these events. This skips `_listener.stop()` / `_listener.start()`, which every other transition relies on. It also silently removes the `before_disable_*` / `after_disable_*` callbacks.

**2. `set_trace(True)` in `before_enable_info`**

Info-level logs shouldn't use trace formatting (`name:filename:lineno`). This would change the output format for all Info-level messages, which seems unintentional.

**3. Redundant `setLevel` calls in `before_enable_trace`, `before_enable_debug`, `before_enable_warning`**

The `for logger in all_loggers()` loop already covers `self._logger` (there's no `_primary_loggers` filter in these methods), so the explicit `self._logger.setLevel(...)` before the loop has no effect.

**4. `set_trace(False)` in `before_enable_default` -- this one is valuable**

This is a good defensive addition. Currently we rely on `before_disable_*` callbacks to reset the trace formatter, but if `enable_default()` is called directly (not via `disable_*`), the formatter state could be stale. I'd suggest we keep this single change.

The original issue #1836 was resolved in 8.4 via #2366 - is there a specific scenario you're still seeing with the current code?

Could we scope this PR down to just the `set_trace(False)` addition in `before_enable_default` (`bittensor/utils/btlogging/loggingmachine.py` line `416`)?

ionodeionode:
Closing to manage open PR count. Will revisit when capacity allows.

basfroman:
> Closing to manage open PR count. Will revisit when capacity allows.

well, are you accidentally opened it back?

ionodeionode:
Thank you for the thorough review! You're absolutely right on all points:

1. **disable_* transitions** — I see now that converting them to plain methods breaks the before/after hook chain. That was an overreach.
2. **set_trace(True) in before_enable_info** — Agreed, this was unintentional and would affect output formatting incorrectly.
3. **Redundant setLevel calls** — Good catch, the loop already covers self._logger.
4. **set_trace(False) in before_enable_default** — Happy to hear this is the valuable part.

I'll scope this PR down to just the `set_trace(False)` addition in `before_enable_default` as you suggested. Will push the update shortly.

basfroman:
Hey @ionodeionode pls merge staging. All tests should pass now.

basfroman (approved):

thewhaleking (approved):

PR #3275, opened 2026-02-25
perf: parallelize sequential awaits with asyncio in async_subtensor

### LINKED ISSUE #3129: Async stuff optimization (for Cortex only)
- [x] improve `determine_block_hash` usage (we need to call it before `asyncio.gather` or before substrate calls) [PR# 3375 ](https://github.com/latent-to/bittensor/pull/3375)
- [x] apply `asyncio.gather` where this can improve performance

altis0725:
Fixes #3129

## Summary
Apply `asyncio.gather` to three methods to run sequential RPC calls concurrently:

- **`get_all_dynamic_info`**: `runtime_call` and `get_subnet_prices` in parallel
- **`blocks_since_last_update`**: `get_block_number` and `get_hyperparameter` in parallel, with pre-resolved `block_hash` for chain state consistency
- **`blocks_until_next_epoch`**: dynamic gather for `get_block_number` and `tempo` when both need fetching

Each parallelized pair saves ~1 RPC round-trip (50-200ms depending on node latency).

## Test plan
- [x] Updated `test_blocks_since_last_update_success` and `test_blocks_since_last_update_no_last_update` for new calling convention
- [ ] CI passes

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -1014,24 +1014,31 @@ class AsyncSubtensor(SubtensorMixin):
         if not block_hash and reuse_block:
             block_hash = self.substrate.last_block_hash
 
-        query = await self.substrate.runtime_call(
-            api="SubnetInfoRuntimeApi",
-            method="get_all_dynamic_info",
-            block_hash=block_hash,
+        async def _safe_get_prices():
+            try:
+                return await self.get_subnet_prices(block_hash=block_hash)
+            except (SubstrateRequestException, ValueError) as e:
+                logging.warning(
+                    f"Unable to fetch subnet prices for block {block}, block hash {block_hash}: {e}"
+                )
+                return None
+
+        query, subnet_prices = await asyncio.gather(
+            self.substrate.runtime_call(
+                api="SubnetInfoRuntimeApi",
+                method="get_all_dynamic_info",
+                block_hash=block_hash,
+            ),
+            _safe_get_prices(),
         )
-        subnet_prices = await self.get_subnet_prices(block_hash=block_hash)
 
         decoded = query.decode()
 
-        if not isinstance(subnet_prices, (SubstrateRequestException, ValueError)):
+        if subnet_prices is not None:
             for sn in decoded:
                 sn.update(
                     {"price": subnet_prices.get(sn["netuid"], Balance.from_tao(0))}
                 )
-        else:
-            logging.warning(
-                f"Unable to fetch subnet prices for block {block}, block hash {block_hash}: {subnet_prices}"
-            )
         return DynamicInfo.list_from_dicts(decoded)
 
     async def blocks_since_last_step(
@@ -1087,14 +1094,18 @@ class AsyncSubtensor(SubtensorMixin):
             The number of blocks since the last update, or None if the subnetwork or UID does not exist.
         """
         block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
-        block = block or await self.substrate.get_block_number(block_hash)
-        call = await self.get_hyperparameter(
+        if block_hash is None:
+            block_hash = await self.get_block_hash()
+        block_coro = self.substrate.get_block_number(block_hash) if block is None else None
+        call_coro = self.get_hyperparameter(
             param_name="LastUpdate",
             netuid=netuid,
-            block=block,
             block_hash=block_hash,
-            reuse_block=reuse_block,
         )
+        if block_coro is not None:
+            block, call = await asyncio.gather(block_coro, call_coro)
+        else:
+            call = await call_coro
         return None if call is None else (block - int(call[uid]))
 
     async def blocks_until_next_epoch(
@@ -1118,8 +1129,17 @@ class AsyncSubtensor(SubtensorMixin):
             The number of blocks until the next epoch of the subnet with provided netuid.
         """
         block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
-        block = block or await self.substrate.get_block_number(block_hash=block_hash)
-        tempo = tempo or await self.tempo(netuid=netuid, block_hash=block_hash)
+        tasks = {}
+        if block is None:
+            tasks["block"] = self.substrate.get_block_number(block_hash=block_hash)
+        if tempo is None:
+            tasks["tempo"] = self.tempo(netuid=netuid, block_hash=block_hash)
+        if tasks:
+            results = dict(
+                zip(tasks.keys(), await asyncio.gather(*tasks.values()))
+            )
+            block = results.get("block", block)
+            tempo = results.get("tempo", tempo)
 
         if not tempo:
             return None
--- a/tests/unit_tests/test_async_subtensor.py
+++ b/tests/unit_tests/test_async_subtensor.py
@@ -2462,7 +2462,11 @@ async def test_blocks_since_last_update_success(subtensor, mocker):
     last_update_block = 50
     current_block = 100
     fake_blocks_since_update = current_block - last_update_block
+    fake_block_hash = "fake_block_hash"
 
+    mocker.patch.object(
+        subtensor, "determine_block_hash", return_value=fake_block_hash
+    )
     mocker.patch.object(
         subtensor.substrate,
         "get_block_number",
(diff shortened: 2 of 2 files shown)


basfroman (changes requested):
Could you pls explain the real benefit of this pr?

altis0725:
Hi @basfroman, thanks for the feedback. I've rebased the branch to a single clean commit.

This PR implements the two items from #3129:
1. **`asyncio.gather`** applied to 3 methods — each saves ~1 RPC round-trip by running sequential substrate calls concurrently
2. **`determine_block_hash` improvement** — pre-resolve `block_hash` in `blocks_since_last_update` before parallel calls to ensure both reference the same chain state

The previous version had too many incremental commits that obscured the intent — apologies for the noise.

bittensor/core/async_subtensor.py
```diff
@@ -1015,12 +1014,18 @@ async def all_subnets(
...
+                api="SubnetInfoRuntimeApi",
+                method="get_all_dynamic_info",
+                block_hash=block_hash,
+            ),
+            self.get_subnet_prices(block_hash=block_hash),
+            return_exceptions=True,
         )
-        subnet_prices = await self.get_subnet_prices(block_hash=block_hash)
+
+        if isinstance(query, BaseException):
```
thewhaleking: In what case would this raise a `BaseException` and why would that be the only thing we'd want to raise?

If we just don't return exceptions [i.e. `return_exceptions=False` (the default)], we don't have to do this logic at all.

bittensor/core/async_subtensor.py
```diff
@@ -1119,8 +1128,21 @@ async def blocks_until_next_epoch(
...
+        if need_tempo:
+            coros.append(self.tempo(netuid=netuid, block_hash=block_hash))
+        if coros:
+            results = await asyncio.gather(*coros)
+            idx = 0
+            if need_block:
+                block = results[idx]
+                idx += 1
+            if need_tempo:
+                tempo = results[idx]
```
thewhaleking: This works, but this logic is fairly fragile. Why don't we either hold it in a dict or assign tasks and use the task name?

thewhaleking (changes requested):
Overall, I think this is good. Just needs some tweaks.

bittensor/core/async_subtensor.py
altis0725: Good point — removed return_exceptions=True entirely. The price fetch is now wrapped in a local try/except so it fails gracefully while the main query propagates exceptions naturally.

bittensor/core/async_subtensor.py
altis0725: Agreed, switched to a dict-based approach. Keys name the tasks explicitly ("block", "tempo") and results are unpacked via results.get().

bittensor/core/async_subtensor.py
```diff
@@ -1015,24 +1014,31 @@ async def all_subnets(
         if not block_hash and reuse_block:
             block_hash = self.substrate.last_block_hash
```
basfroman, line 1017: I recently requested a new implementation to retrieve subnet prices in a single call. Nucleus plans to introduce a new RPC call that will return already pre-calculated prices, replacing the current complex `query_map` + SDK-side calculations in `get_subnet_prices`. After that, applying `asyncio.gather` to `all_subnets` becomes trivial - just two clean RPC calls in parallel, [same approach as `get_subnets_info`](https://github.com/opentensor/bittensor/blob/v10.1.0/bittensor/core/async_subtensor.py#L1348-L1361).

I want to avoid double work and wait for the new `get_subnet_prices` RPC, then apply `asyncio.gather` in one clean change.

bittensor/core/async_subtensor.py
```diff
@@ -1088,14 +1094,18 @@ async def blocks_since_last_update(
...
             netuid=netuid,
-            block=block,
             block_hash=block_hash,
-            reuse_block=reuse_block,
         )
+        if block_coro is not None:
+            block, call = await asyncio.gather(block_coro, call_coro)
+        else:
+            call = await call_coro
         return None if call is None else (block - int(call[uid]))
```
basfroman, line 1109: The `blocks_since_last_update` change should be dropped from this PR.

The current implementation is straightforward - three sequential steps, easy to read and reason about:

```python
block = block or await self.substrate.get_block_number(block_hash)
call = await self.get_hyperparameter(
    param_name="LastUpdate",
    netuid=netuid,
    block=block,
    block_hash=block_hash,
    reuse_block=reuse_block,
)
return None if call is None else (block - int(call[uid]))
```

The proposed `asyncio.gather` here parallelizes `get_block_number` (a single lightweight RPC) with `get_hyperparameter` (which internally chains `determine_block_hash` -> `subnet_exists` -> `substrate.query`). The actual savings is one network round-trip - `get_block_number` runs alongside the first call inside `get_hyperparameter`. That's ~20-50ms at best on a remote node, negligible on a local one.

The trade-off is not worth it:
- Code grows from 4 lines to 9 with conditional branching (`block_coro`/`call_coro`)
- `reuse_block` is dropped from the `get_hyperparameter` call
- An extra `get_block_hash()` call is introduced when `block_hash` is `None`
- Readability decreases significantly for a marginal gain

bittensor/core/async_subtensor.py
```diff
@@ -1119,8 +1129,17 @@ async def blocks_until_next_epoch(
...
+        if block is None:
+            tasks["block"] = self.substrate.get_block_number(block_hash=block_hash)
+        if tempo is None:
+            tasks["tempo"] = self.tempo(netuid=netuid, block_hash=block_hash)
+        if tasks:
+            results = dict(
+                zip(tasks.keys(), await asyncio.gather(*tasks.values()))
+            )
+            block = results.get("block", block)
+            tempo = results.get("tempo", tempo)
```
basfroman, line 1142: Applies to `blocks_until_next_epoch` - the dict-based gather pattern (`dict(zip(tasks.keys(), ...))`) for at most two values is **overengineered** for the same negligible round-trip saving.

basfroman (changes requested):
Thank you for the PR. I've reviewed each change in detail and left comments. Pls read all of my comments I provided to the code changes.

Conclusion: Pls close this PR or wait until new RPC with fetching all subnets prices wull be produced and then update `all_subnets` method with simple `asyncio.gather` usage. Pls let me know which way you prefer.

altis0725:
Hi @basfroman, thank you for the detailed review — I appreciate the thorough explanations.

I agree that `blocks_since_last_update` and `blocks_until_next_epoch` changes aren't worth the added complexity for marginal gains. I'll drop those.

Regarding `all_subnets`: I understand you've requested a new RPC from Nucleus for pre-calculated subnet prices. I checked subtensor and didn't find a related issue or PR yet — do you have a rough timeline for when that might land?

- **If it's coming soon (weeks)**: Happy to close this PR and resubmit a clean `asyncio.gather` once the new RPC is available.
- **If it's further out (months)**: Would it make sense to keep the current `all_subnets` gather as a stopgap?

Either way works for me — just want to avoid the PR sitting idle indefinitely. Let me know your preference.

bittensor/core/async_subtensor.py
thewhaleking: I disagree. I think there's a slightly cleaner way of writing this, but this is the general flow that needs to happen for this.

thewhaleking:
> * **If it's coming soon (weeks)**: Happy to close this PR and resubmit a clean `asyncio.gather` once the new RPC is available.
> * **If it's further out (months)**: Would it make sense to keep the current `all_subnets` gather as a stopgap?
> 
> Either way works for me — just want to avoid the PR sitting idle indefinitely. Let me know your preference.

Can you just write normal comments? I don't want to feel like I'm talking to ChatGPT when I read PR comments/questions.

bittensor/core/async_subtensor.py
thewhaleking: I would opt for something like this:

```py
async def noop(x: Any) -> Any:
    return x

block_ = self.substrate.get_block_number(block_hash=block_hash) if block is None else noop(block)
tempo_ = self.tempo(netuid=netuid, block_hash=block_hash) if tempo is None else noop(tempo)
block, tempo = await asyncio.gather(block_, tempo_)
```

I too am annoyed that Python does not have asyncio lambdas.

altis0725:
Closing as suggested — will revisit once the new RPC for fetching subnet prices lands. Thanks for the reviews.

PR #3293, opened 2026-03-21
fix: next nonce should be next nonce

Arthurdw:
This is causing some issues, afaik this should just be the current nonce + 1 right?


Arthurdw:
The e2e's fail, because they expect n+1 atm, but it should be n+2 right? So are the e2e's wrong? (asking before I change anything)

Arthurdw:
Closed as per https://discord.com/channels/1120750674595024897/1242999357436071956/1485613120834633839

PR #3298, opened 2026-03-25
Removes munch

thewhaleking:
Removes munch, a lib we barely used, and which hasn't been updated in years.

### DIFF
--- a/bittensor/core/settings.py
+++ b/bittensor/core/settings.py
@@ -3,8 +3,6 @@ import os
 import re
 from pathlib import Path
 
-from munch import munchify
-
 ROOT_TAO_STAKE_WEIGHT = 0.18
 
 READ_ONLY = os.getenv("READ_ONLY") == "1"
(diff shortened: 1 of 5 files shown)


bittensor/core/settings.py
```diff
@@ -114,50 +112,46 @@
...
+        external_ip = os.getenv("BT_AXON_EXTERNAL_IP") or None
+        max_workers = int(_BT_AXON_MAX_WORKERS) if _BT_AXON_MAX_WORKERS else 10
+
+    class logging:
+        debug = bool(os.getenv("BT_LOGGING_DEBUG")) or False
+        trace = bool(os.getenv("BT_LOGGING_TRACE")) or False
+        info = bool(os.getenv("BT_LOGGING_INFO")) or False
+        record_log = bool(os.getenv("BT_LOGGING_RECORD_LOG")) or False
+        logging_dir = (
+            None
```
basfroman, line 134: I understand that, in terms of the effort-to-result ratio, this is an excellent approach, but my eye is still twitching LOL

basfroman (approved):

bittensor/core/settings.py
thewhaleking: I mean, this is what it looked like after being munchified. I would not have chosen this structure myself lol

PR #3300, opened 2026-03-30
There is not Balances.transfer

thewhaleking:
Just changes an incorrect docstring reference


thewhaleking:
Test failures are because GitHub Actions is about as reliable as a blind, deaf dog.

basfroman (approved):

PR #3262, opened 2026-02-16
fix: correct exception handling in get_external_ip and add validation…

amussara:
… helper

- Fix bug where get_external_ip() caught ExternalIPNotFound in each fallback block, but the actual exceptions (ConnectionError, AssertionError, Timeout, etc.) were never caught, causing the function to crash instead of falling through to the next service.
- Replace bare assert with a dedicated _validate_ip_response() helper that strips whitespace and validates IP format.
- Add timeout parameters to requests.get() and urllib_request.urlopen() calls to prevent indefinite hangs.
- Add return type annotation to ip__str__().
- Add comprehensive test coverage for ip_version, ip_to_int validation, ip__str__ edge cases, _validate_ip_response, and get_external_ip fallthrough behavior.

Welcome!

Due to [GitHub limitations](https://github.com/orgs/community/discussions/4620),
please switch to **Preview** for links to render properly.

Please choose the right template for your pull request:

- 🐛 Are you fixing a bug? [Bug fix](?expand=1&template=bug_fix.md)
- 📈 Are you improving performance? [Performance improvement](?expand=1&template=performance_improvement.md)
- 💻 Are you changing functionality? [Feature change](?expand=1&template=feature_change.md)

PR #3311, opened 2026-04-11
fix: correct exception handling in get_external_ip fallback chain

### LINKED ISSUE #3309: `get_external_ip` fallback chain never triggers — all attempts catch `ExternalIPNotFound` but none of the called functions raise it
`get_external_ip()` has five fallback IP providers, each wrapped in `try/except ExternalIPNotFound`. none of the called functions — `requests.get`, `ip_to_int` (raises `netaddr.core.AddrFormatError`), `os.popen`/curl, or `urllib_request.urlopen` — ever raise `ExternalIPNotFound`. when any attempt fails the uncaught exception propagates and the function crashes instead of falling through to the next provider

**to reproduce**

1. disable or block outbound requests to `checkip.amazonaws.com` (or run with no network)
2. call `get_external_ip()`
3. observe `requests.exceptions.ConnectionError` propagating up instead of the AWS attempt being skipped and the curl fallback being tried

**expected behaviour**

each failed provider should fall through to the next; only after all five are exhausted should `ExternalIPNotFound` be raised. the `except` clauses need to catch the exceptions that actually occur: `requests.exceptions.RequestException`, `netaddr.core.AddrFormatError`, `OSError`, and `urllib.error.URLError`

**environment**

- file: `bittensor/utils/networking.py`, function `get_external_ip()`

Grizouforever:
## Summary

Fixes #3309

The fallback chain in `get_external_ip()` catches `ExternalIPNotFound`, but none of the code inside the try blocks actually raises that exception. Real exceptions (`requests.RequestException`, `AssertionError`, `OSError`, `json.JSONDecodeError`, etc.) propagate unhandled, preventing fallthrough to the next IP provider. This means if the first provider (AWS checkip) fails, the function crashes instead of trying the remaining 5 providers.

## Changes

**`bittensor/utils/networking.py`**:
- Replace all `except ExternalIPNotFound` with `except Exception` to properly catch real exceptions
- Replace deprecated `os.popen()` calls with `subprocess.run()` (Python 3.0+ recommended)
- Replace bare `assert isinstance(...)` validation with direct `ip_to_int()` calls that raise on invalid input regardless of `-O` optimization flag

## Context

PR #3262 previously attempted this fix but was closed due to being based on the wrong branch and author not responding to review. This PR is based on `master` (the current default branch) and incorporates the additional improvements suggested.

### DIFF
--- a/bittensor/utils/networking.py
+++ b/bittensor/utils/networking.py
@@ -1,6 +1,7 @@
 """Utils for handling local network with ip and ports."""
 
 import os
+import subprocess
 from typing import Optional
 from urllib import request as urllib_request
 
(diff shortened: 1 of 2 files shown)


thewhaleking:
Read CONTRIBUTING

bittensor/utils/networking.py
```diff
@@ -1,6 +1,7 @@
 """Utils for handling local network with ip and ports."""
 
 import os
```
basfroman, line 3: dead import

bittensor/utils/networking.py
```diff
@@ -1,6 +1,7 @@
...
+import subprocess
```
basfroman, line 4: why use `subprocess` when `requests` is available?

bittensor/utils/networking.py
```diff
@@ -66,55 +67,58 @@ def get_external_ip() -> str:
...
-    # --- Try ipconfig.
+    # --- Try ifconfig.me
     try:
-        process = os.popen("curl -s ifconfig.me")
-        external_ip = process.readline()
-        process.close()
-        assert isinstance(ip_to_int(external_ip), int)
+        result = subprocess.run(
+            ["curl", "-s", "ifconfig.me"], capture_output=True, text=True, timeout=10
+        )
```
basfroman, line 79: `subprocess` + `curl` -> `requests.get`

tests/unit_tests/utils/test_networking.py
```diff
@@ -134,31 +134,22 @@ def mock_call():
...
-    class FakeReadline:
-        def readline(self):
-            return 1
-
-    def mock_call():
-        return FakeReadline()
+    def mock_subprocess_run(*args, **kwargs):
+        raise OSError("subprocess.run mocked to fail")
 
     class FakeResponse:
```
basfroman, line 142: isn't `FakeResponse` dead now?

tests/unit_tests/utils/test_networking.py
basfroman, line 1: Two things missing on the test side:

First: test_get_external_ip_os_broken still patches os.popen, but this PR removes all os.popen usage from production code. That mock now does nothing; the test passes solely because requests.get is mocked to succeed on the first try. The test name and docstring say "when os.popen is broken," but it no longer tests that scenario. 
Pls update it to mock subprocess.run failing (or whichever replacement we land on) while keeping requests.get succeeding, so it actually validates "first method works, broken fallback methods don't interfere."

Second: there's no test for the cascade itself. For example: mock requests.get to raise, mock the first subprocess.run call to return a valid IP -> assert the function returns that IP. This is the core contract of get_external_ip() - "try multiple sources in order" - and it has zero coverage today.

basfroman (changes requested):
You really found a bug. I'll happily accept the PR, but please take a look at the comments and rebase your pr to `staging` branch.
Thanks for this contribution!

thewhaleking:
No response from user.

PR #3313, opened 2026-04-14
Removes flake8, uses ruff for linter check

### LINKED ISSUE #3312: Change flake8 to ruff

thewhaleking:
Uses ruff (which we already use) instead of flake8 for the linter

Also fixes items which fail the linter check


basfroman (commented):
- `Makefile` has to be updated
- `.flake8` file has to be deleted
- `.github/workflows/flake8-and-mypy.yml` has to be renamed
- it would be nice to extend `pyproject.toml` with `[tool.ruff.lint]` with rules
- `pruning_score` with `# noqa` could be removed (requires updating MockSubtensor)

basfroman (approved):

PR #3318, opened 2026-04-20
fix(networking): catch real exceptions in get_external_ip fallback chain

ArtificialXai:
## Summary

Closes #3309.

Each provider in `get_external_ip` is wrapped in `except ExternalIPNotFound`, but none of the underlying calls (`requests.get`, `ip_to_int`, `os.popen` + `readline`, `urllib.request.urlopen`, `json.loads`, the Wikipedia header lookup) ever raise `ExternalIPNotFound`. As a result the **first** failing provider crashes the entire function instead of falling through to the next one — exactly the symptom reported in the issue (a `requests.exceptions.ConnectionError` propagates out when AWS is unreachable, rather than the curl fallback being attempted).

## Fix

Replace each `except ExternalIPNotFound:` with a tuple of exceptions that the providers actually raise:

| Exception | Where it originates |
|---|---|
| `requests.exceptions.RequestException` | `requests.get(...)` (AWS, Wikipedia) |
| `netaddr.core.AddrFormatError` | `ip_to_int(...)` with malformed input |
| `OSError` | `os.popen(...)` + `readline()` (ifconfig.me, ipinfo.io, dnsomatic) |
| `urllib.error.URLError` | `urllib_request.urlopen(...)` (ident.me) |
| `ValueError` | `json.loads(...)` parse failures, `int(...)` parse failures |
| `KeyError` | Wikipedia response missing `X-Client-IP` header |
| `AssertionError` | `assert isinstance(ip_to_int(...), int)` guard |
| `ExternalIPNotFound` | Kept for backward compatibility with any wrapper that raises it |

The set is defined once at module scope as `_IP_LOOKUP_EXCEPTIONS` so the providers stay in sync.

## Tests

Three new regression tests in `tests/unit_tests/utils/test_networking.py`:

- `test_get_external_ip_aws_connection_error_falls_through_to_curl` — AWS `requests.get` raises `ConnectionError`, curl fallback returns a valid IP, function returns that IP (was: raised `ConnectionError`).
- `test_get_external_ip_malformed_ip_falls_through` — AWS returns a non-IP string (triggers `AddrFormatError` in `ip_to_int`), curl fallback succeeds.
- `test_get_external_ip_all_providers_exhausted_raises` — every provider raises its real exception; function must raise `ExternalIPNotFound` (not leak the last underlying exception).

All 22 tests in `tests/unit_tests/utils/test_networking.py` pass locally.

## Test plan

[...]

PR #3346, opened 2026-05-12
fix(axon): narrow preprocess exception handling and chain causes

RUNECTZ33:
## Summary

`AxonMiddleware.preprocess` previously wrapped any `Exception` from URL path parsing and `Synapse.from_headers` as `InvalidRequestNameError` / `SynapseParsingError`. This caused two problems:

1. **Unrelated infrastructure failures got mislabeled.** A `RuntimeError`, `MemoryError`, or a real bug in starlette/pydantic plugins would surface to the client as a misleading "improperly formatted request" — making the actual cause invisible.
2. **Original tracebacks were lost.** The bare `raise X(...)` (without `from e`) dropped the chained cause, so even when the wrapping was appropriate (e.g. pydantic validation), the underlying failure couldn't be inspected.

## Changes

- Narrow the path-parse catch to `(IndexError, AttributeError)` — the only failures `request.url.path.split(\"/\")[1]` can realistically produce.
- Narrow the synapse-parse catch to `(ValidationError, TypeError, ValueError)` — the realistic failure modes of pydantic model construction from header inputs. `parse_headers_to_inputs` already catches its own per-header decode errors internally, so what propagates is pydantic validation failures at the `cls(**input_dict)` step.
- `raise ... from e` on both paths so the original exception remains on `__cause__` for debugging.

## Tests

Two regression tests added to `TestAxonMiddleware`:
- `test_preprocess_wraps_validation_error_with_cause` — asserts `ValidationError` from `from_headers` becomes `SynapseParsingError` AND `__cause__` is preserved.
- `test_preprocess_does_not_swallow_unrelated_errors` — asserts an unrelated `RuntimeError` from `from_headers` propagates unchanged.

## Style match

Mirrors the exception-narrowing approach already accepted in #3318 (`fix(networking): catch real exceptions in get_external_ip fallback chain`).

### DIFF
--- a/bittensor/core/axon.py
+++ b/bittensor/core/axon.py
@@ -15,6 +15,7 @@ from inspect import signature, Signature, Parameter
 from typing import Any, Awaitable, Callable, Optional, Tuple
 
 from async_substrate_interface.utils import json
+from pydantic import ValidationError
 import uvicorn
 from bittensor_wallet import Wallet, Keypair
 from fastapi import APIRouter, Depends, FastAPI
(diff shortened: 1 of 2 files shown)


thewhaleking (changes requested):
overall looks fine, but sign your commits

RUNECTZ33:
Re-signed via SSH commit signing — thanks for the catch, @thewhaleking. Should show `verified` now. Ready for another look when you have a minute.

ibraheem-abe (approved):

basfroman (approved):

PR #3349, opened 2026-05-14
Mev Shield Nonce Increment Fix

thewhaleking:
Each sign_and_send_extrinsic call previously triggered create_signed_extrinsic's cache-incrementing fallback.

Compatibility with current asi and asi 2.0.4

### DIFF
--- a/bittensor/core/extrinsics/asyncex/coldkey_swap.py
+++ b/bittensor/core/extrinsics/asyncex/coldkey_swap.py
@@ -392,6 +392,11 @@ async def swap_coldkey_announced_extrinsic(
             )
 
         if response.success:
+            # The swap may reap the old coldkey's account on chain (resetting its
+            # nonce). Drop the in-process nonce cache entry so the next extrinsic
+            # signed by this ss58 re-fetches from the chain instead of incrementing
+            # a stale value.
+            subtensor.substrate._nonces.pop(wallet.coldkeypub.ss58_address, None)
             logging.debug("[green]Coldkey swap executed successfully.[/green]")
         else:
             logging.error(f"[red]{response.message}[/red]")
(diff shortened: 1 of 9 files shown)


ibraheem-abe (approved):

bittensor/core/extrinsics/asyncex/coldkey_swap.py
```diff
@@ -392,6 +392,11 @@ async def swap_coldkey_announced_extrinsic(
             )
 
         if response.success:
+            # The swap may reap the old coldkey's account on chain (resetting its
+            # nonce). Drop the in-process nonce cache entry so the next extrinsic
+            # signed by this ss58 re-fetches from the chain instead of incrementing
+            # a stale value.
+            subtensor.substrate._nonces.pop(wallet.coldkeypub.ss58_address, None)
```
basfroman, line 399: we will not be using calls to private methods. The ASI API need to be made public.
the case in another places

basfroman (changes requested):

bittensor/core/extrinsics/asyncex/coldkey_swap.py
thewhaleking: Well, then it's a catch22. So will we release 2.0.4 with a failing test?

basfroman: That’s a good question. We could proceed that way. However, I’d prefer not to introduce technical debt into the SDK. Once the ASI is released, please implement the change to make the private method public, and we will immediately apply those changes here. 
does if make sense for you?

thewhaleking: Yes, it's already done in https://github.com/latent-to/async-substrate-interface/pull/355

basfroman (approved):

PR #3365, opened 2026-06-01
Hoist determine_block_hash out of asyncio.gather

Yupsecous:
Resolve the block hash once before three asyncio.gather sites (get_all_subnets_info, AsyncMetagraph._process_root_weights and _apply_extra_info) and fan out on block_hash, removing one redundant concurrent get_block_hash RPC per site. Behavior unchanged; adds regression tests asserting get_block_hash is awaited once.

Refs #3129

### Requirements for Contributing a Performance Improvement

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only affect performance of existing functionality
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution.

### Branch Acknowledgement
[ ] I am acknowledging that I am opening this branch against `staging`

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -1356,18 +1356,15 @@ class AsyncSubtensor(SubtensorMixin):
             A list of SubnetInfo objects, each containing detailed information about a subnet.
 
         """
+        block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
         result, prices = await asyncio.gather(
             self.query_runtime_api(
                 runtime_api="SubnetInfoRuntimeApi",
                 method="get_subnets_info_v2",
                 params=[],
-                block=block,
                 block_hash=block_hash,
-                reuse_block=reuse_block,
             ),
-            self.get_subnet_prices(
-                block=block, block_hash=block_hash, reuse_block=reuse_block
-            ),
+            self.get_subnet_prices(block_hash=block_hash),
             return_exceptions=True,
         )
         if not result:
--- a/bittensor/core/metagraph.py
+++ b/bittensor/core/metagraph.py
@@ -1510,9 +1510,10 @@ class AsyncMetagraph(NumpyOrTorch):
                 self.root_weights = self._process_root_weights(raw_root_weights_data, "weights", subtensor)
         """
         data_array = []
+        block_hash = await subtensor.determine_block_hash(block)
         n_subnets_, subnets = await asyncio.gather(
-            subtensor.get_total_subnets(block=block),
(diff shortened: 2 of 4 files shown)


thewhaleking (approved):
Looks fine to me

thewhaleking:
Flaky test flaking. Should be fixed by https://github.com/latent-to/bittensor/pull/3367

basfroman (changes requested):
Hey, looked at this closer - `AsyncSubstrateInterface.get_block_hash` already goes through `CachedFetcher` which deduplicates concurrent calls via an inflight Future dict (`_inflight`). Meaning when two coroutines in a gather both call `get_block_hash(123)`, only one RPC actually fires, the second one awaits the first's Future.

So the scenario this PR prevents (duplicate `chain_getBlockHash` RPCs from a single gather) can't actually happen at the transport layer. The hoisted `determine_block_hash` resolves to the same single RPC either way.

The change isn't harmful and arguably reads better, but it's not a performance fix - it's cosmetic. Wanted to flag there's no measurable gain to claim here.

If we still want it as a readability refactor - totally fine, just worth reframing the commit message and removing the "Refs #3129" perf issue link (or closing that issue as already-resolved by the caching layer).

thewhaleking:
> Hey, looked at this closer - `AsyncSubstrateInterface.get_block_hash` already goes through `CachedFetcher` which deduplicates concurrent calls via an inflight Future dict (`_inflight`). Meaning when two coroutines in a gather both call `get_block_hash(123)`, only one RPC actually fires, the second one awaits the first's Future.
> 
> So the scenario this PR prevents (duplicate `chain_getBlockHash` RPCs from a single gather) can't actually happen at the transport layer. The hoisted `determine_block_hash` resolves to the same single RPC either way.
> 
> The change isn't harmful and arguably reads better, but it's not a performance fix - it's cosmetic. Wanted to flag there's no measurable gain to claim here.
> 
> If we still want it as a readability refactor - totally fine, just worth reframing the commit message and removing the "Refs #3129" perf issue link (or closing that issue as already-resolved by the caching layer).

I think it makes more sense to drop this PR, and just use ASI 2.1.0+, as it accomplishes the same goal while being more readable (e.g. keeping the current code in staging)

Yupsecous:
Thanks both — you're right. I'd missed that get_block_hash already collapses concurrent same-block calls through CachedFetcher's in-flight future map, so the duplicate never reaches the transport layer and this hoist is cosmetic rather than a perf win. Agreed that keeping the current code in staging (and relying on ASI 2.1.0+) is cleaner. Closing this out — thanks for the careful review. 🙏

PR #3368, opened 2026-06-02
Forward version_key from commit_weights

Yupsecous:
# Fix commit_weights dropping the caller's version_key

> Base branch: `staging`. No separate issue — the bug is described in full below
> (per maintainer guidance that a detailed PR description suffices).

### Bug

`commit_weights` (sync + async) ignores a caller-supplied `version_key`,
producing a commit hash that a matching `reveal_weights` cannot satisfy. No
separate issue was filed; the full reproduction, root cause, and fix are
described below.

### Description of the Change

`Subtensor.commit_weights` / `AsyncSubtensor.commit_weights` accept `version_key`
but never forwarded it to `commit_weights_extrinsic`, which uses it in
`generate_weight_hash` to build the commit hash. The extrinsic therefore fell
back to its default (`version_as_int`), so a custom `version_key` was silently
ignored. Because `reveal_weights` *does* forward `version_key`, the commit and
reveal hashes disagreed and the reveal was rejected on-chain.

The fix forwards `version_key=version_key` in both wrappers (one line each),
matching how `reveal_weights` and `set_weights` already forward it.

### Alternate Designs

None needed — the parameter simply has to be threaded through, exactly as the
sibling weights wrappers already do.

### Possible Drawbacks

Behavior-preserving for default callers: the wrapper default and the extrinsic
default are both `version_as_int` (10004000, runtime-verified), so the default
commit hash is unchanged. Only callers passing a custom `version_key` change
behavior — which is the fix. Scope is `commit_weights` only: `generate_weight_hash`
is used solely by `commit_weights_extrinsic`, and `reveal_weights`/`set_weights`
(both `commit_timelocked_weights_extrinsic` and `set_weights_extrinsic`) already
forward `version_key`.

### Verification Process

- Repro-first: `test_commit_weights_forwards_version_key` (sync + async) fails on
  `staging` with `KeyError: 'version_key'`; passes with the fix.
- Symptom-level `test_commit_weights_hash_matches_reveal` (**sync and async**,
  each parametrized default + custom) runs `commit_weights` through the real
  extrinsic (chain submission mocked), captures the actual committed hash, and
  asserts it equals the hash the chain recomputes at reveal for the same inputs
  and `version_key`. The **custom** case fails on `staging` (hashes differ) while
  the **default** case passes, confirming default callers are unaffected and the
  test is precise. The async variant exercises the real `asyncex/weights.py`
  hashing (extrinsic not mocked) and is verified to fail under a
  `version_key=version_as_int` hardcode mutation.
- Python 3.12, async-substrate-interface 2.1.0:
  `-k "commit_weights or reveal_weights or set_weights"` → **15 passed**; full
  `test_subtensor.py` + `test_async_subtensor.py` → **510 passed**;
  `ruff format --check`, `ruff check`, `mypy` clean.

### Release Notes

Fixed `commit_weights` ignoring a custom `version_key`, which caused the
subsequent `reveal_weights` to be rejected on-chain.

### Branch Acknowledgement

### DIFF
--- a/bittensor/core/async_subtensor.py
+++ b/bittensor/core/async_subtensor.py
@@ -7041,6 +7041,7 @@ class AsyncSubtensor(SubtensorMixin):
                     uids=uids,
                     weights=weights,
                     salt=salt,
+                    version_key=version_key,
                     mev_protection=mev_protection,
                     period=period,
                     raise_error=raise_error,
--- a/bittensor/core/subtensor.py
+++ b/bittensor/core/subtensor.py
@@ -5846,6 +5846,7 @@ class Subtensor(SubtensorMixin):
                     uids=uids,
                     weights=weights,
                     salt=salt,
+                    version_key=version_key,
                     mev_protection=mev_protection,
                     period=period,
                     raise_error=raise_error,
--- a/tests/unit_tests/test_async_subtensor.py
+++ b/tests/unit_tests/test_async_subtensor.py
@@ -4885,6 +4885,104 @@ async def test_get_crowdloans(mocker, subtensor):
     assert result == [mocked_decode_crowdloan_entry.return_value]
 
 
+@pytest.mark.asyncio
+async def test_commit_weights_forwards_version_key(subtensor, fake_wallet, mocker):
+    """Regression: async commit_weights must forward the caller's version_key to the
+    extrinsic. If dropped, the commit hash is built with the default version_key and
+    a reveal made with the intended version_key mismatches on-chain."""
+    # Preps
+    netuid = 1
+    uids = [1, 2, 3, 4]
+    weights = [10, 20, 30, 40]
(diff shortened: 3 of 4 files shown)


basfroman (approved):
Fix is correct and consistent with how `reveal_weights` and `set_weights` already forward `version_key`. No impact on default callers. Worth noting: no existing caller passes a custom `version_key` to `commit_weights` today, so this is a correctness fix for an unused code path.

basfroman (changes requested):
pls fix broken tests

Yupsecous:
Got it

basfroman:
> Got it

nvm. upon closer examination, I realized that this is not an issue with your PR
I will let you know once the fix is ​​deployed to staging, and then you can pull in the updates.

basfroman (approved):

PR #3383, opened 2026-06-11
Support `dynamic tempo`

basfroman:
Adds SDK support for the new stateful epoch model introduced in subtensor.

## AC for merge:
- [ ] `bittensor-drand v2.0.0` released
- [ ] `bittensor-drand` updated in SDK dependency in `pyproject.toml` file
- [ ] Subtensor [PR#2638](https://github.com/opentensor/subtensor/pull/2638) merged

### Storage getters
- `get_last_epoch_block`
- `get_pending_epoch_at`
- `get_subnet_epoch_index`
- `get_activity_cutoff_factor_milli`
- `get_owner_hyperparam_rate_limit`
- `get_epoch_schedule_state`.

### Extrinsics 
- `set_tempo`
- `set_activity_cutoff_factor`
- `root_set_activity_cutoff_factor`
- `trigger_epoch`.

## Details
- `blocks_until_next_epoch` and `get_next_epoch_start_block` now delegate to the `SubnetInfoRuntimeApi` runtime API instead of local modulo arithmetic. The `tempo` parameter is removed from `blocks_until_next_epoch`.
- `commit_timelocked_weights_extrinsic` migrated from `get_encrypted_commit` to `get_encrypted_commit_v2` which takes `EpochScheduleState` for epoch prediction.
- `SubnetHyperparameters` refactored to a hybrid container that supports v1/v2/v3 chain formats via `from_any`. Unknown fields spill into `hyperparameters` dict. Added `activity_cutoff_factor` field. `activity_cutoff` is now `Optional[int]`.

## New entries:
- `EpochScheduleState` dataclass in `types.py`. 
- `epoch_schedule` utility module with `blocks_until_next_auto_epoch` and `is_in_admin_freeze_window`.

## Breaking changes:
- Requires `bittensor-drand>=2.0.0`.
- `Subtensor.blocks_until_next_epoch` and `AsyncSubtensor.blocks_until_next_epoch`: `block` is now keyword-only. Callers passing it positionally must update to `blocks_until_next_epoch(netuid, block=N)`.

## Related links:
- Subtensor base PR: https://github.com/opentensor/subtensor/pull/2638
- Bittensor Drand v.2.0.0 PR: https://github.com/latent-to/bittensor-drand/pull/71
- [e2e tests](https://github.com/latent-to/bittensor/actions/runs/27373904328) based on subtensor PR docker image [E2E Subtensor Tests](https://github.com/latent-to/bittensor/actions/runs/27398990384)


bittensor/core/async_subtensor.py
```diff
@@ -2831,6 +2864,55 @@ async def get_ema_tao_inflow(
...
+        Parameters:
+            netuid: The unique identifier of the subnetwork.
+            block: The block number to query. Do not specify if using `block_hash` or `reuse_block`.
+            block_hash: The block hash at which to check. Do not set if using `block` or `reuse_block`.
+            reuse_block: Whether to reuse the last-used block hash.
+
+        Returns:
+            An `EpochScheduleState` populated from on-chain storage.
+        """
+        block_hash = await self.determine_block_hash(block, block_hash, reuse_block)
```
thewhaleking:
```suggestion
        if not (block_hash := await self.determine_block_hash(block, block_hash, reuse_block)):
        block_hash = await self.substrate.get_chain_head()
```
While the requests are coalesced within the later `gather`, if this happens let's say at second 11.5 of the block cycle and `block_number = ` takes 0.5 seconds to respond, it will be using a different block than ones from the gather.

There are other ways to do this, but that would be the simplest.

bittensor/core/subtensor.py
```diff
@@ -894,30 +902,25 @@ def blocks_since_last_update(
         return None if len(call) == 0 else (block - int(call[uid]))
 
     def blocks_until_next_epoch(
-        self, netuid: int, tempo: Optional[int] = None, block: Optional[int] = None
+        self, netuid: int, block: Optional[int] = None
```
thewhaleking: Could make the transition easier by requiring block as a kwarg only:

```suggestion
        self, netuid: int, *, block: Optional[int] = None
```

This should cause it to raise a TypeError rather than if someone were migrating their pre 10.5 code to this, would silently fail (well worse, would pass tempo as block).
Regardless, should list this under Breaking Changes.

bittensor/core/extrinsics/tempo_control.py
```diff
@@ -0,0 +1,283 @@
...
+    *,
+    mev_protection: bool = DEFAULT_MEV_PROTECTION,
+    period: Optional[int] = None,
+    raise_error: bool = False,
+    wait_for_inclusion: bool = True,
+    wait_for_finalization: bool = True,
+    wait_for_revealed_execution: bool = True,
+) -> ExtrinsicResponse:
+    """
+    Triggers an immediate epoch on a subnet. Owner (coldkey) only.
```
thewhaleking: Not sure I'd say "immediate" here because it's current block + admin freeze window (incl per-block epoch cap). Maybe check with Maciej or Trexman or Dera to update wording. Also the other places where this language occurs.

bittensor/core/types.py
```diff
@@ -673,3 +673,27 @@ class LockState(TypedDict):
...
+        tempo: The subnet's tempo (epoch period in blocks).
+        blocks_since_last_step: Blocks elapsed since the last epoch.
+        current_block: The reference block at which the snapshot was taken.
+    """
+
+    last_epoch_block: int
+    pending_epoch_at: int
+    subnet_epoch_index: int
+    tempo: Optional[int]
+    blocks_since_last_step: Optional[int]
```
thewhaleking: These are both `int`s chain side, not Optional. Not sure if this was just for mypy or what.

bittensor/core/chain_data/subnet_hyperparameters.py
```diff
@@ -1,120 +1,161 @@
...
+    def _typed_field_names() -> frozenset[str]:
+        return frozenset(
+            f.name for f in fields(SubnetHyperparameters) if f.name != "hyperparameters"
         )
+
+    @staticmethod
+    def _decode_value(value: Any) -> Any:
+        """Decode a single ``{<type_tag>: <payload>}`` hyperparameter value."""
+        if isinstance(value, dict) and set(value.keys()) == {"bits"}:
+            return fixed_to_float(value["bits"], frac_bits=32)
```
thewhaleking, line 73: Don't hardcode this 32 here, let the `_FIXED_POINT_TAG` handle that.

thewhaleking (dismissed):

ibraheem-abe (approved):

PR #3393, opened 2026-06-18
fix(axon): reject requests with missing signature in default_verify

Thykof:
### Bug

Issue: https://github.com/latent-to/bittensor/issues/3392

Summary: Axon.default_verify (bittensor/core/axon.py) skips signature verification entirely when the incoming request carries no signature, allowing any caller to impersonate any dendrite hotkey.

### Description of the Change

  default_verify reconstructs the signed message from the request headers and then runs three checks in order: it rejects a missing nonce ("Missing Nonce"), rejects a stale nonce, and finally verifies the signature.

  The signature check was written as a single combined condition:

```python
  if synapse.dendrite.signature and not keypair.verify(message, synapse.dendrite.signature):
      raise Exception("Signature mismatch ...")
```

  Because signature presence is and-ed with the verification, an empty or absent `synapse.dendrite.signature` short-circuits the expression to False, so `keypair.verify(...)` is never called. The request then falls through the rest of
  `default_verify` as authenticated — the nonce is even recorded. There is a "Missing Nonce" guard but no matching "Missing Signature" guard.

  This change adds the missing presence guard (mirroring the nonce guard) and then verifies unconditionally:

```python
  if not synapse.dendrite.signature:
      raise Exception("Missing signature")

  if not keypair.verify(message, synapse.dendrite.signature):
      raise Exception(f"Signature mismatch with {message} and {synapse.dendrite.signature}")
```

  A present-but-invalid signature was already rejected correctly (keypair.verify → False → raise). The only hole was the empty/absent case, which this closes.

  Impact: an attacker can claim any hotkey in bt_header_dendrite_hotkey — including a high-stake validator — and send no signature. The forged request passes default_verify and reaches the subnet's blacklist/forward. On the standard miner
  blacklist (gated on validator_permit + stake), this lets an attacker harvest miner outputs while masquerading as a legitimate validator, with no secret material required.

### Alternate Designs

  - Keep one combined condition (if not signature or not keypair.verify(...): raise). Functionally equivalent, but it collapses two distinct failure modes into one message and loses parity with the existing "Missing Nonce" guard. The explicit
  two-step form is clearer and gives an actionable error.
  - Let keypair.verify handle the empty input (drop the presence check, always call verify). This relies on keypair.verify("") returning False for every keypair backend — implementation-dependent and easy to regress. An explicit presence
  check is safer and self-documenting.

  The proposed version was chosen for explicitness, a clear distinct error message, and symmetry with the surrounding nonce checks.

### Possible Drawbacks

  - Requests that previously "succeeded" with no signature will now be rejected. That path was the vulnerability itself, so this is the intended behavior — no legitimately signed client is affected (the dendrite always attaches a signature).
  - No change to the public API, headers, or the valid-signature path; the only behavioral change is that the previously-bypassing unsigned case now fails closed.

### Verification Process

  - Unit: a synapse with dendrite.signature = "" (and None) now raises "Missing signature" from default_verify instead of returning successfully.
  - Unit (regression): a correctly signed synapse still verifies; a present-but-wrong signature still raises "Signature mismatch".
  - Integration: an unsigned request claiming a registered validator hotkey is rejected at the axon (HTTP 401) and never reaches forward.
  - Run: pytest tests/unit_tests/test_axon.py.

### Release Notes

Reject requests with missing signature

### Branch Acknowledgement

  - [x] I am acknowledging that I am opening this branch against staging.

### DIFF
--- a/bittensor/core/axon.py
+++ b/bittensor/core/axon.py
@@ -975,7 +975,7 @@ class Axon:
                     raise Exception("Nonce is too old, a newer one was last processed")
 
             if not synapse.dendrite.signature:
-                raise Exception("Missing signature")
+                raise Exception("Missing Signature")
 
             if not keypair.verify(message, synapse.dendrite.signature):
                 raise Exception(
--- a/bittensor/core/axon.py
+++ b/bittensor/core/axon.py
@@ -974,9 +974,10 @@ class Axon:
                 ):
                     raise Exception("Nonce is too old, a newer one was last processed")
 
-            if synapse.dendrite.signature and not keypair.verify(
-                message, synapse.dendrite.signature
-            ):
+            if not synapse.dendrite.signature:
+                raise Exception("Missing signature")
+
+            if not keypair.verify(message, synapse.dendrite.signature):
                 raise Exception(
                     f"Signature mismatch with {message} and {synapse.dendrite.signature}"
                 )
--- a/tests/unit_tests/test_axon.py
+++ b/tests/unit_tests/test_axon.py
@@ -21,7 +21,7 @@ from bittensor.core.axon import Axon, AxonMiddleware, FastAPIThreadedServer
 from bittensor.core.errors import RunException
 from bittensor.core.settings import version_as_int
 from bittensor.core.stream import StreamingSynapse
-from bittensor.core.synapse import Synapse
+from bittensor.core.synapse import Synapse, TerminalInfo
 from bittensor.core.threadpool import PriorityThreadPoolExecutor
 from bittensor.utils.axon_utils import (
     allowed_nonce_window_ns,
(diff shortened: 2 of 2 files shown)


basfroman (changes requested):
Nice fix, the change is correct and well-documented.

One suggestion: `test_default_verify_empty_signature_raises` and `test_default_verify_missing_signature_raises` test the same code path (`if not synapse.dendrite.signature`) with different falsy values. Consider merging them into a single parametrized test to avoid near-duplication:

```python
@pytest.mark.parametrize("empty_sig", ["", None])
async def test_default_verify_missing_signature_raises(empty_sig):
...
```
Otherwise LGTM. Please address this and re-request review.

Thykof:
Hi @basfroman, thank you for the review. I have pushed the requested changes.

basfroman (approved):

